PR #5, opened 2018-02-15
Configure the Gradle wrapper via the Wrapper task.

chrisr3:
Using the `Wrapper` task enables IntelliJ to work out which version of the Gradle APIs it is supposed to be building against.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,3 +82,8 @@ artifactory {
         }
     }
 }
+
+task wrapper(type: Wrapper) {
+    gradleVersion = "4.4.1"
+    distributionType = Wrapper.DistributionType.ALL
+}


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-bin.zip
```
Clintonio: Shouldn't this be "all" still?

chrisr3: This is what it generated by the `Wrapper` task. It doesn't seem to want/need the `-all` artifact.

chrisr3: Turns out that this is configurable.

PR #6, opened 2018-02-15
Remove Corda's public API definition.

chrisr3:
Corda's API definition belongs in the corda repository.


Clintonio:
Can we keep the shell script and readme? We'll be using those in the near future

chrisr3:
The API definition _at least_ should go, because it's the wrong API!

PR #7, opened 2018-02-16
[CORDA-926]: Node configuration parsing fails on unknown properties.

sollecitom:


sollecitom:
Needed by https://github.com/corda/corda/pull/2484

constants.properties
```diff
@@ -1,6 +1,6 @@
-gradlePluginsVersion=4.0.0
+gradlePluginsVersion=4.1.0
```
Clintonio: 4.0.1. It should match the intended version of Corda (eg; 4.0). It's not semver sadly.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -223,7 +223,13 @@ open class Node @Inject constructor(private val project: Project) : CordformNode
             config = config.withValue("notary", ConfigValueFactory.fromMap(notary))
         }
         if (extraConfig != null) {
-            config = config.withFallback(ConfigFactory.parseMap(extraConfig))
+            val extraProperties = ConfigFactory.parseMap(extraConfig)
```
Clintonio: The original point of these extraProperties was to allow users to add any extra config that we haven't exposed via the DSL. This would push this all under "custom" and break the feature.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -323,16 +322,73 @@ open class Node @Inject constructor(private val project: Project) : CordformNode
...
+                it.apply {
+                    from(webServerConfigFile)
+                    into(rootDir)
+                }
+            }
+        }
+    }
+
+    private fun Config.toNodeOnly(): Config {
+
```
Clintonio: Whitespace.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -323,16 +322,73 @@ open class Node @Inject constructor(private val project: Project) : CordformNode
...
+        config = copyTo("baseDirectory", config)
+        config = copyTo("keyStorePassword", config)
+        config = copyTo("trustStorePassword", config)
+        config = copyTo("exportJMXto", config)
+        config = copyTo("custom", config)
+        return config
+    }
+
+    private fun Config.copyTo(key: String, target: Config, targetKey: String = key): Config {
+
```
Clintonio: Whitespace.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -323,16 +322,73 @@ open class Node @Inject constructor(private val project: Project) : CordformNode
...
+        return if (hasPath(key)) {
+            target + (targetKey to getValue(key))
+        } else {
+            target
+        }
+    }
+
+    private operator fun Config.plus(property: Pair<String, Any>) = withValue(property.first, ConfigValueFactory.fromAnyRef(property.second))
```
Clintonio: Can this and copyTo be moved into a Utils.kt file. These are generic enough that they don't belong in the node.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
sollecitom: Removed all blank lines I introduced after function declaration. Not sure I like the convention against them, but fair enough.

sollecitom:
@Clintonio addressed your comments, please have another look.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
Clintonio: Yeah, I'm going with the Corda core code style to be consistent.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -323,14 +318,57 @@ open class Node @Inject constructor(private val project: Project) : CordformNode
...
+    private fun Config.toNodeOnly(): Config {
+        return if (hasPath("webAddress")) {
+            withoutPath("webAddress").withoutPath("useHTTPS")
+        } else {
+            this
+        }
+    }
+
+    private fun Config.toWebServerOnly(): Config {
+        var config = ConfigFactory.empty()
```
Clintonio: Can you rename this `newConfig` or something? It's a bit confusing to read otherwise. Might also be worth chaining these operations in some way.

PR #8, opened 2018-02-16
Enable hiding of annoations that have been annotated as @CordaInternal.

chrisr3:
Not every annotation needs to be documented in the public API definition file. Modify the API Scanner plugin so that it does not output annotations that have themselves been annotated as `@CordaInternal`.

And add tests. Lots of tests.


Clintonio:
double check the CI task runs your tests before merging.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -37,10 +37,12 @@
     private static final int FIELD_MASK = Modifier.fieldModifiers();
     private static final int VISIBILITY_MASK = Modifier.PUBLIC | Modifier.PROTECTED;
 
+    private static final String CORDA_INTERNAL = "net.corda.core.CordaInternal";
     private static final Set<String> ANNOTATION_BLACKLIST;
     static {
        Set<String> blacklist = new LinkedHashSet<>();
        blacklist.add("kotlin.jvm.JvmOverloads");
+       blacklist.add(CORDA_INTERNAL);
```
shamsasari: What happens if someone adds this annotation to an existing public API member? That's a compatibility break, right? I think we need to be including this annotation in the output.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -400,3 +420,17 @@ private static URL toURL(File file) throws MalformedURLException {
...
+    List<String> visible;
+    @SuppressWarnings("WeakerAccess") List<String> hidden;
+
+    Names(List<String> visible, List<String> hidden) {
+        this.visible = unmodifiable(visible);
+        this.hidden = unmodifiable(hidden);
+    }
+
+    private static <T> List<T> unmodifiable(List<T> list) {
+        return list.isEmpty() ? emptyList() : unmodifiableList(new ArrayList<>(list));
```
shamsasari: Abit excessive, no?

api-scanner/src/test/java/net/corda/plugins/AnnotatedClassesTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.rules.TemporaryFolder;
+
+import java.io.File;
+import java.io.IOException;
+
+import static org.junit.Assert.*;
+
+public class AnnotatedClassesTest {
+    @Rule
+    public TemporaryFolder testProjectDir = new TemporaryFolder();
```
shamsasari: `final`

api-scanner/src/test/java/net/corda/plugins/AnnotatedClassesTest.java
```diff
@@ -0,0 +1,55 @@
...
+        System.out.println(output);
+
+        BuildTask scanApi = result.task(":scanApi");
+        assertNotNull(scanApi);
+        assertEquals(TaskOutcome.SUCCESS, scanApi.getOutcome());
+
+        File api = new File(testProjectDir.getRoot(), "build/api/annotated-classes.txt");
+        assertTrue(api.isFile());
+        assertEquals(
+            "@net.corda.example.NotInherited @net.corda.example.IsInherited public class net.corda.example.HasInheritedAnnotation extends java.lang.Object\n" +
```
shamsasari: Asserting the equality of large texts like this makes the test brittle. Can we not at least split the output into a list of lines and assert that the list contains the bits we're looking for.

chrisr3: The output _is_ supposed to be reproducible and constant. And I have tried to minimise the number of classes in each test...

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
chrisr3: `emptyList()` is already unmodifiable, so wrapping it with `unmodifiableList()` might cause me to wake up screaming in the middle of the night.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
chrisr3: Hmmm, isn't this a new annotation that can only be applied to "property getters/setters" and new annotations? AIUI, the semantics here are that `@CordaInternal` things are not included in the public API, c.f. the `ScanApi.writeMethods()` function where `@CordaInternal` methods are _explicitly_ hidden (by design). If someone were to annotate an existing API method as `@CordaInternal` then that method would also completely disappear from the output and so the "API Stability" build would break. So what's the problem?

api-scanner/src/test/java/net/corda/plugins/AnnotatedClassesTest.java
```diff
@@ -0,0 +1,55 @@
...
+            .withPluginClasspath()
+            .build();
+        String output = result.getOutput();
+        System.out.println(output);
+
+        BuildTask scanApi = result.task(":scanApi");
+        assertNotNull(scanApi);
+        assertEquals(TaskOutcome.SUCCESS, scanApi.getOutcome());
+
+        File api = new File(testProjectDir.getRoot(), "build/api/annotated-classes.txt");
```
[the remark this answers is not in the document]
chrisr3: Hmm, I can _try_ but the Gradle API here is rather "`File` heavy".

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
shamsasari: I mean `unmodifiableList(list)` is sufficient

chrisr3: I'm not sure about that, because `unmodifiableList(list)` could still be modified if someone had a reference to the original `list`.

PR #11, opened 2018-02-20
[CORDA-941]: Add NetworkParameters contract implementation whitelist.

sollecitom:
Needed by https://github.com/corda/corda/pull/2580.

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
@@ -142,11 +142,15 @@ open class Baseform : DefaultTask() {
     protected fun bootstrapNetwork() {
         val networkBootstrapperClass = loadNetworkBootstrapperClass()
         val networkBootstrapper = networkBootstrapperClass.newInstance()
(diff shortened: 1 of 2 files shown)


sollecitom:
@Clintonio I need to understand how to get `allCordapps` now, because the plugins were not only extracted but also restructured.

mikehearn:
Do we have a circularity issue here now the plugins were split out? This PR needs an API change in the main repo, but the PR that introduces it to the main repo needs this PR?

sollecitom:
@mikehearn : yes and no. Yes, conceptually we do; and no, that code expects the `NetworkBootstrapper` JAR to have that method, so it's dynamic and it will blow up if it's not there, but we're safe in terms of compilation.

The issue here is that `NetworkBootstrapper` should be in its own module, rather than in `node-api`, and in its own repository, so that instead of having a soft circular dependency, we would have a clear triangle.

To be noted that the `NetworkBootstrapper` is a standalone Java application that it's not referenced by Corda.

mikehearn:
Either that or everything should be in one repository like before.... commits are transactions and transactional updates are good :) But I realise they just got split out and trying to keep a single repo seems to be a losing battle. So whatever. Just make sure it's not disruptive when this lands.

PR #15, opened 2018-02-27
Removed the forced jolokia dependency.

egeek:
Due to the way that the fat jar for the CorDapp is constructed after the dependency is forced, there are duplicated files in the JAR. Furthermore jolokia is a potential security risk as it increases the attack surfaces, and therefore should only be included by a project if they wish to use it.

Determine whether to use Quasar from the lib directory (as per older versions of Cordformation), or to pick up a specific version as a dependency based on whether or not a group for quasar has been
specified.

Log the dependencies that will be included in the final CorDapp JAR file to make it easier to see what the result of these plugins are.


mikehearn:
Thanks. That all looks reasonable but yes, I'd like to wait for Clinton. There's been quite a lot of change in these plugins recently and I need a catch-up with him on the current state anyway.

Most of this code is workarounds for lack of a module system. We're seeing more and more issues caused by Java 8's lack of a module system. The alternatives we have are:

* Roll our own (sigh)
* OSGi (too complicated)
* Jigsaw (java 9 breaks the world)

I'd much rather go for (3) so after we're over this hump with Finastra and R3C, I'd like us to start a "Java Next" team (j9 will be obsolete). The team would be tasked with upgrading us past Java 8 and utilising the new functionality. The downside is Java 9 breaks backwards compatibility all over the place, so that's why we probably need a small team to work on it, maybe one person at first to get us to the "it runs" stage and then one or two more people to start experimenting with a Jigsaw integration.

If it turns out to be too hard/too breaky, we can look at using custom ClassLoaders to hide implementation details of the node from loaded apps. But Jigsaw has the right sort of design and lots of work went into it, so I'd rather we go for gold first.

If we can get ourselves into module land, then the node will be able to understand the dependencies of an app JAR from the metadata within it and can automatically attach all the dependency JARs as well. So then we don't need fat-jars anymore.

The downside for large CorDapps is that to make a Java 9 module your dependencies also need to be modules, and so on down the graph recursively (which makes sense - otherwise you'd reach a point in the graph where you don't have enough metadata to continue). That might complicate things.

constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=4.0.3
+gradlePluginsVersion=4.0.4
```
Clintonio: You no longer need to bump the version number - that's done by a maintainer now.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -6,6 +6,7 @@ import java.util.*
 
 private val HEADLESS_FLAG = "--headless"
 private val CAPSULE_DEBUG_FLAG = "--capsule-debug"
+private val JOLOKIA_FLAG = "--jolokia"
```
Clintonio: Could you make this "--expose-jolokia" to make it more obvious to those who may want to understand what a particular flat does?

Clintonio:
Also if you want you can add yourself to the "CONTRIBUTORS.MD" file.

Clintonio:
@roastario has assured me that this can be closed.

PR #16, opened 2018-02-27
Fixes #12 - no longer pass --no-local-shell to headless webserver, causing it to fail.

### LINKED ISSUE #12: In headless mode NodeRunner passes --no-local-shell to WebServer causing it to fail
When you use `runnodes` this invokes [`NodeRunner`](cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt) and if the environment is headless or you specify `--headless` the it uses [`HeadlessJavaCommand`](https://github.com/corda/corda-gradle-plugins/blob/master/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt#L113) which causes the executable jars that are found to all be started with the additional argument `--no-local-shell`.

The [`Corda`](https://github.com/corda/corda/blob/master/node/src/main/kotlin/net/corda/node/Corda.kt) nodes know how to consume this argument as they use an [`ArgParser`](https://github.com/corda/corda/blob/master/node/src/main/kotlin/net/corda/node/ArgsParser.kt#L38) that expects this.

However the [`WebServer`](https://github.com/corda/corda/blob/master/webserver/src/main/kotlin/net/corda/webserver/WebServer.kt) processes choke on this argument as they use a different [`ArgParser`](https://github.com/corda/corda/blob/master/webserver/src/main/kotlin/net/corda/webserver/WebArgsParser.kt#L21) that doesn't know what to do with this argument.

So when you use `runnode` you see something like this:

```
$ ./runnodes
Starting nodes in /vagrant
Starting corda.jar in /vagrant/PartyB on debug port 5005
Starting corda-webserver.jar in /vagrant/PartyB on debug port 5006
Started 2 processes
Finished starting nodes
Listening for transport dt_socket at address: 5006
Unknown command line arguments: no-local-shell is not a recognized optionUnknown command line arguments: no-local-shell is not a recognized option
```

I.e. here you see it starting both a Corda node and a webserver process but the webserver quickly exits with the `Unknown command line arguments: no-local-shell` error.

This means the introductory [corda app](https://docs.corda.net/tutorial-cordapp.html) fails fairly mysteriously if you try running it on e.g. a headless VM (as I guess is typical with people using things like Docker or, in my case, Vagrant).

george-hawkins:
This minimal change causes `NodeRunner` to associate the argument `--no-local-shell` only with `NodeJarType` because also passing it to the web server caused it to fail as outlined in #12 

This change introduces the idea of a list of headless specific arguments (and for completeness a list of headed specific arguments). By default these lists are empty.

The `NodeJarType` adds `--no-local-shell` to its headless arguments while `WebJarType` does not, so we no longer have the issue with `--no-local-shell` being passed to the web server.

### DIFF
--- a/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
+++ b/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
@@ -55,17 +55,22 @@ private abstract class JarType(private val jarName: String) {
         val debugPort = debugPortAlloc.next()
         val monitoringPort = monitoringPortAlloc.next()
         println("Starting $jarName in $dir on debug port $debugPort")
-        val process = (if (headless) ::HeadlessJavaCommand else ::TerminalWindowJavaCommand)(jarName, dir, debugPort, monitoringPort, javaArgs, jvmArgs).start()
+        val appArgs = javaArgs + (if (headless) headlessArgs else headedArgs)
+        val process = (if (headless) ::HeadlessJavaCommand else ::TerminalWindowJavaCommand)(jarName, dir, debugPort, monitoringPort, appArgs, jvmArgs).start()
         if (os == OS.MACOS) Thread.sleep(1000)
         return process
     }
 
     internal abstract val configurationFileName: String
+    protected open val headlessArgs = listOf<String>()
+    protected open val headedArgs = listOf<String>()
+
 }
 
 private object NodeJarType : JarType("corda.jar") {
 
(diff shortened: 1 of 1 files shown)


andr3ej:
Hi @george-hawkins. Thanks for this PR but I have discovered this issue independently on Thursday and am halfway through fixing it, very similarly to how you're fixing it here. Please see https://r3-cev.atlassian.net/browse/CORDA-1131 to track progress, I've listed my pull requests in the comments section. Closing this PR as duplicate.

PR #20, opened 2018-03-01
API Scanner supports any @CordaInternal annotation, regardless of its package.

chrisr3:
Some modules may wish to define internal annotations, and yet not depend on `core`. Support this case by allowing _any_ package to define a `@CordaInternal` annotation. The API Scanner will handle these in the same way as `@net.corda.core.CoreInternal`.

### DIFF
--- a/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
+++ b/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
@@ -1,10 +1,7 @@
 package net.corda.plugins;
 
 import io.github.lukehutch.fastclasspathscanner.FastClasspathScanner;
-import io.github.lukehutch.fastclasspathscanner.scanner.ClassInfo;
-import io.github.lukehutch.fastclasspathscanner.scanner.FieldInfo;
-import io.github.lukehutch.fastclasspathscanner.scanner.MethodInfo;
-import io.github.lukehutch.fastclasspathscanner.scanner.ScanResult;
+import io.github.lukehutch.fastclasspathscanner.scanner.*;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.FileCollection;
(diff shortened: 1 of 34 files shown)


api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -200,9 +202,18 @@ void scan(PrintWriter writer, ClassLoader appLoader) {
             return spec;
         }
 
-        private void setInvisibleAnnotations(ScanResult result) {
-            Set<String> invisible = new LinkedHashSet<>(ANNOTATION_BLACKLIST);
-            invisible.addAll(result.getNamesOfAnnotationsWithMetaAnnotation(CORDA_INTERNAL));
+        private void loadAnnotationCaches(ScanResult result) {
+            Set<String> internal = result.getNamesOfAllAnnotationClasses().stream()
+                .filter((s) -> s.endsWith(INTERNAL_ANNOTATION_NAME))
```
shamsasari: The brackets are need around `s` I think

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -200,9 +202,18 @@ void scan(PrintWriter writer, ClassLoader appLoader) {
...
+                .collect(toSet());
```
shamsasari: `toSet` doesn't guarantee mutability. Use the suggested `toCollection` instead.

api-scanner/src/test/java/net/corda/plugins/InternalAnnotationTest.java
```diff
@@ -0,0 +1,50 @@
...
+    public void testInternalAnnotations() throws IOException {
+        BuildResult result = GradleRunner.create()
+            .withProjectDir(testProjectDir.getRoot())
+            .withArguments("scanApi", "--info")
+            .withPluginClasspath()
+            .build();
+        String output = result.getOutput();
+        System.out.println(output);
+
+        assertTrue(output.contains("net.corda.core.CordaInternal"));
```
shamsasari: Can you use AssertJ instead? You can do this in one assert statement and it will helpfully print out the list if it fails.

api-scanner/src/test/resources/kotlin-internal-annotation/kotlin/net/corda/kotlin/AnnotatedClass.kt
```diff
@@ -0,0 +1,4 @@
+package net.corda.kotlin
```
shamsasari: `net.corda.apiscanner.kotlin`

api-scanner/src/test/java/net/corda/plugins/InternalAnnotationTest.java
chrisr3: This is a (very) long `String` rather than a `List`. I am ensuring that the invisible annotation was at least mentioned in the `fast-classpath-scanner` output.

shamsasari: OK, but my comment about AssertJ still stands with regards to a String

api-scanner/src/test/resources/kotlin-internal-annotation/kotlin/net/corda/kotlin/AnnotatedClass.kt
chrisr3: These are test packages in distinct test Gradle projects. All of the rest are using `net.corda.example` packages except that IntelliJ is trying to conflate all of the classes into a single project. Perhaps `net.corda.example.kotlin` would be a better choice?

PR #21, opened 2018-03-01
CORDA-1131 Fix webserver startup when headless

andr3ej:


constants.properties
```diff
@@ -1,4 +1,4 @@
-gradlePluginsVersion=4.0.3
+gradlePluginsVersion=4.0.4
```
Clintonio: Bump to .5, someone has managed to get .4 in already.

PR #24, opened 2018-03-04
[CORDA-1162] Sort class/interface/method annotations into ascending order.

chrisr3:
The API Scanner currently writes `class`/`interface`/`method` annotations in the order they are returned from `fast-classpath-scanner:2.7`. While this ordering is stable, it is also an implementation detail of `fast-classpath-scanner` and so is likely to change in future versions. So we need to sort these annotations ourselves into ascending order.

**NOTE**: This change will require us to update `api-current.txt` in the Corda repository.

### DIFF
--- a/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
+++ b/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
@@ -389,7 +389,7 @@ public class ScanApi extends DefaultTask {
                 method.getTypeDescriptor(),
                 method.getAnnotationNames().stream()
                     .filter(this::isVisibleAnnotation)
-                    // TODO: Sort these annotations ourselves!
+                    .sorted()
                     .collect(toList())
             );
         }
--- a/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
+++ b/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
@@ -18,6 +18,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.*;
+import java.util.function.Function;
 import java.util.stream.StreamSupport;
 
 import static java.util.Collections.*;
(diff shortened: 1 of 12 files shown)


api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -348,14 +349,10 @@ private Names toNames(Collection<ClassInfo> classes) {
...
-            /*
-             * These annotations are returned in an order determined by
-             * fast-classpath-scanner. THIS IS BAD! The ordering here
-             * will almost certainly change/break in future versions.
-             * TODO: Order these annotations ourselves!
-             */
-            return new Names(partitioned.get(true), partitioned.get(false));
+                .collect(partitioningBy(this::isVisibleAnnotation, toCollection(LinkedList::new)));
+
+            Function<List<String>, List<String>> ordering = list -> { sort(list); return list; };
```
shamsasari: What you had before (method which sorted and then returned the list) was better

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -348,14 +349,10 @@ private Names toNames(Collection<ClassInfo> classes) {
...
                 .filter(ScanApi::isApplicationClass)
-                .collect(partitioningBy(this::isVisibleAnnotation));
-            /*
-             * These annotations are returned in an order determined by
-             * fast-classpath-scanner. THIS IS BAD! The ordering here
-             * will almost certainly change/break in future versions.
-             * TODO: Order these annotations ourselves!
-             */
-            return new Names(partitioned.get(true), partitioned.get(false));
+                .collect(partitioningBy(this::isVisibleAnnotation, toCollection(LinkedList::new)));
```
shamsasari: Why LinkedList? It doesn't look like you're removing elements from the head. In fact you're sorting and so ArrayList would be better anyway.

chrisr3: As a rule, I prefer `LinkedList` over `ArrayList` when I don't know how many elements I will be inserting, so that I don't need to resize the array. However, I suppose we can assume that the number of annotations in either list will be "small".

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
chrisr3: I thought a "local function" approach would be better precisely because `sort()` returns `void` in Java. Adding a private `sort()` method which returned the list started to have confusing implications for the sorts in the rest of the code. Semantically, wouldn't you expect the new function to be "generally applicable" and so return a sorted _copy_ of the list, rather than the sorted original list?

Basically I dodged the issue by creating a lambda for this very _specific_ use-case only.

shamsasari: The `Function` declaration just makes this a bit too verbose, that's all.

PR #26, opened 2018-03-05
Update to 4.0.6-SNAPSHOT

chrisr3:
4.0.5 has been released, so this is now 4.0.6-SNAPSHOT until we're ready to release 4.0.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,7 @@ To bootstrap this repository you can install the publishing plugins with;
 To install locally for testing a new build against Corda you can run the following from the project root;
 
     ./gradlew install
+
+## Release process
+
+The version number of the "bleeding edge" in `master` is always a `-SNAPSHOT` version. To create a new release, a _maintainer_ must create a new branch off the latest commit in this release, remove the `-SNAPSHOT` from the version number and then run the publish to artifactory task in TeamCity. The version number in `master` is than advanced to the next `-SNAPSHOT` number.
--- a/constants.properties
+++ b/constants.properties
@@ -1,4 +1,4 @@
-gradlePluginsVersion=4.0.5
+gradlePluginsVersion=4.0.6-SNAPSHOT
 kotlinVersion=1.2.20
 platformVersion=3
 guavaVersion=21.0


Clintonio:
In fact it seems like we'll need to tag branches, not master, otherwise the version number will be a snapshot one.

chrisr3:
I'm still thinking about how to make this work. Would the developer change `A.B.C-SNAPSHOT` to `A.B.(C+1)-SNAPSHOT` in master, and then create a release branch off `A.B.C-SNAPSHOT` for `A.B.C`?

Clintonio:
I think the Corda main repo makes sense;

1. Dev does a feature branch
2. Dev merges it once it passes CI and code review
3. Repeat 1-2 until we decide to release it (usually if it's required in Corda, which will be most PRs here)
4. A maintainer (ie; someone with write access to this repo and CI access) now branches off a release branch (eg; release-4.0.x)
5. They change the version number by removing -SNAPSHOT
6. They tag (eg; release/4.0.x)
7. Push the tag
8. Publish tag to Artifactory in TeamCity
9. They contact a _publisher_ (eg; me, Ben) to publish to Bintray and Maven Central (this will be changing soon, but worth writing down)
10. Update master to be 4.0.(x+1)-SNAPSHOT

This takes the burden of worrying about version numbers off of PRs and pushes it onto which ever maintainer wishes to release a new version for whatever reason.

This is obviously a lot more of a burden than the old "ask clint to do it" method but I'm okay with that. The number of releases per week has gotten significantly higher. 

Any suggestions welcome of course, this is getting written up soon and it seems I've written a small portion down here already.

Clintonio:
(I realise I wrote the same thing as you in long form)

andr3ej:
a gradle task could automate the git parts? maybe there is an existing gradle "release" plugin that does the above? otherwise writing our own shouldn't be too hard

Clintonio:
@andr3ej  I wouldn't be against automating this. First steps though are to automate the final deployment. 

I believe that there's a gradle git plugin that we used to extract the current hash, I wouldn't be surprised if it can also commit changes. If you have time to look at this I can review the design or implementation.

PR #32, opened 2018-03-13
CORDA-1155 Pass Jolokia through Slf4J

ags313:


ags313:
Changelog modified. Does this still make sense if we are trying to remove Jolokia?

PR #37, opened 2018-03-15
CORDA-1224: Update API Scanner not to output `@JvmStatic` annotations.

chrisr3:
The `@JvmStatic` annotation doesn't need to appear in the API output because it also causes the Kotlin compiler to generate `static final` modifiers. And it is only the `static final` modifiers that are important for the API.

This will allow us to improve the Java-friendliness of the API without the API Stability gate generating a lot of false-positives.


chrisr3:
This change is equivalent to the following `sed` script:
```bash
$ sed -e's/@kotlin.jvm.JvmStatic //g' api-current.txt
```

PR #38, opened 2018-03-15
Remove joloklia forced dep

roastario:
this is a PR to remove the explicit dependency on joloklia.


chrisr3:
Why do we have this new PR when #15 is still open?

changelog.md
```diff
@@ -5,7 +5,8 @@
 ### Version 4.0.9
 
 `api-scanner`: Remove the `@JvmStatic` annotation from the generated output.
-
+`cordformation`: Removed default dependency on jolokia, to provide HTTP JMX monitoring, explicitly depend on jolokia.
```
chrisr3: Oops, we have a formatting issue here. The markdown should be an itemised list, so this should be:
```markdown
* `api-scanner`: ...
* `cordformation`: ...
* `noderunner`: ...
```
Unfortunately I forgot to start some of the lines with `*`. Could you add them back in please?

Clintonio:
@roastario has assured me that this can be closed.

PR #45, opened 2018-04-04
CORDA-1301: Also exclude versions of dependencies that were created via jitpack.

chrisr3:
CorDapps are supposed to exclude `quasar-core` from their "fat jars". However, we have switched to a "jitpack" version of `quasar-core` which has `group=com.github.corda.quasar` instead of `group=co.paralleluniverse`. So extend the pattern-matching to handle jitpack artifacts in the `com.github.corda` namespace.

And while I'm here, fix some compiler warnings and upgrade the plugin to use `java-gradle-plugin`.

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
@@ -9,6 +9,7 @@ import java.io.File
  * The Cordapp plugin will turn a project into a cordapp project which builds cordapp JARs with the correct format
  * and with the information needed to run on Corda.
  */
+@Suppress("UNUSED")
 class CordappPlugin : Plugin<Project> {
     override fun apply(project: Project) {
         project.logger.info("Configuring ${project.name} as a cordapp")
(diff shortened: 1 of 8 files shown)


cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -57,11 +64,12 @@ class CordappPlugin : Plugin<Project> {
         val directDeps = runtimeConfiguration.allDependencies - excludeDeps
         // We want to filter out anything Corda related or provided by Corda, like kotlin-stdlib and quasar
         val filteredDeps = directDeps.filter { dep ->
-            excludes.none { exclude -> (exclude["group"] == dep.group) && (exclude["name"] == dep.name) }
+            excludes.none { exclude -> (exclude["name"] == dep.name) && (
+                    (exclude["group"] == dep.group) || dep.group?.startsWith("com.github.corda.") ?: false) }
```
Clintonio: Could you instead scan for the "com.github.corda" packages and add them to the "exclude deps" so we can keep the excluded deps in a single structure.

chrisr3: Actually, looking more closely at the code:
```kotlin
val excludeDeps = project.configuration("cordapp").allDependencies +
        project.configuration("cordaCompile").allDependencies +
        project.configuration("cordaRuntime").allDependencies
val directDeps = runtimeConfiguration.allDependencies - excludeDeps
// We want to filter out anything Corda related or provided by Corda, like kotlin-stdlib and quasar
val filteredDeps = directDeps.filter { dep ->
    excludes.none { exclude -> (exclude["group"] == dep.group) && (exclude["name"] == dep.name) }
}
```
You appear to be talking about `excludeDeps`, whereas the `excludes` (which is where the bug is) are actually applied to `filteredDeps`. This PR is about broadening the filter to include jitpack versions of the `excludes` artifacts.

PR #46, opened 2018-04-05
CORDA-1301: Test that quasar-core is added to runtime and cordaRuntime.

chrisr3:


vkolomeyko:
@Clintonio - we would like a second opinion from you.

PR #55, opened 2018-04-25
CORDA-1027 Fix api scanner tests on windows

anthonykeenan:
....but do they still work on the build agents?

### DIFF
--- a/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
+++ b/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
@@ -13,7 +13,6 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 
 import static net.corda.plugins.CopyUtils.*;
 import static org.assertj.core.api.Assertions.*;
--- a/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
+++ b/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
@@ -21,7 +21,7 @@ import static org.junit.Assert.*;
 
 public class AnnotatedClassTest {
     @Rule
-    public final TemporaryFolder testProjectDir = new TemporaryFolder(Paths.get("build").toFile());
+    public final TemporaryFolder testProjectDir = new TemporaryFolder();
 
     @Before
     public void setup() throws IOException {
--- a/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
+++ b/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
@@ -13,6 +13,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 
 import static net.corda.plugins.CopyUtils.*;
 import static org.assertj.core.api.Assertions.*;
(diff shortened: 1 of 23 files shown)


api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
```diff
@@ -13,14 +13,15 @@
...
+import java.nio.file.Paths;
 
 import static net.corda.plugins.CopyUtils.*;
 import static org.assertj.core.api.Assertions.*;
 import static org.junit.Assert.*;
 
 public class AnnotatedClassTest {
     @Rule
-    public final TemporaryFolder testProjectDir = new TemporaryFolder();
+    public final TemporaryFolder testProjectDir = new TemporaryFolder(Paths.get("build").toFile());
```
chrisr3: This _shouldn't_ be necessary, because the `build.gradle` file at the project root contains the following:
```gradle
allprojects {
    ...
    tasks.withType(Test) {
        // Prevent the project from creating temporary files outside of the build directory.
        systemProperty 'java.io.tmpdir', buildDir.absolutePath
        ...
    }
    ...
}
```
My understanding of JUnit's `TemporaryFolder()` rule is that it is equivalent to:
```java
File.createTempFile("junit", "", null)
```
which means that it will use `java.io.tmpdir` as the parent directory. So what error are you getting on Windows? Is the value of `buildDir.absolutePath` wrong somehow?

anthonykeenan: On windows, it is creating a temporary directory in my user account local data, nowhere near where the build is running, so all the relative directories are wrong.

chrisr3: The point is that this _cannot_ be a Windows problem because `java.io.tmpdir` works fine on Windows. It sounds more like this system property isn't being set - perhaps IntelliJ isn't parsing the `build.gradle` files correctly?

chrisr3: Having said all this, have you tested (say) the `KotlinLambdasTest`? Because if IntelliJ _is_ the problem becsuse it's only partially handling the `build.gradle` files, then I suspect it's not handling this either:
```gradle
processTestResources {
    filesMatching('**/kotlin-*/build.gradle') {
        expand(['kotlin_version': kotlin_version])
    }
}
```
which means that the Kotlin tests will fail. And if `java.io.tmpdir` isn't being set then `test.gradle.user/home` isn't going to be set either. And that means that each unit test using Kotlin will download the Kotlin artifacts from the Internet - _every time_!

Basically, these tests were written to run in Gradle. As indeed I thought the entire `corda-gradle-plugins` project is.

anthonykeenan: The tests fail in gradle as well as intellij for the same reasons on my machine. I'm just investigating the `java.io.tmpdir` variable now.

api-scanner/src/test/java/net/corda/plugins/KotlinVarargMethodTest.java
```diff
@@ -22,6 +22,7 @@
     @Before
     public void setup() throws IOException {
         File buildFile = testProjectDir.newFile("build.gradle");
+        System.out.println(System.getProperty("java.io.tmpdir"));
```
chrisr3: Oops, I think this can go... ;-)

cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
```diff
@@ -9,6 +9,7 @@ import org.junit.rules.TemporaryFolder
 import java.io.File
 import org.gradle.testkit.runner.GradleRunner
 import org.gradle.testkit.runner.TaskOutcome
+import java.nio.file.Paths
```
chrisr3: One more unused import.

PR #56, opened 2018-04-27
CORDA-1027 Anthony make api annotations more readable

anthonykeenan:
* Add deprecated annotations to blacklist
* Add a couple of tests for deprecated annotations
* Make annotations go on their own line so addition of a new annotation doesn't break api compatability
* Remove package names from annotations to make the file more human readable
* Moar tests

### DIFF
--- a/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
+++ b/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
@@ -45,6 +45,8 @@ public class ScanApi extends DefaultTask {
        blacklist.add("kotlin.jvm.JvmField");
        blacklist.add("kotlin.jvm.JvmOverloads");
        blacklist.add("kotlin.jvm.JvmStatic");
+       blacklist.add("kotlin.Deprecated");
+       blacklist.add("java.lang.Deprecated");
        blacklist.add(DEFAULT_INTERNAL_ANNOTATION);
        ANNOTATION_BLACKLIST = unmodifiableSet(blacklist);
     }
--- a/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
+++ b/api-scanner/src/main/java/net/corda/plugins/ScanApi.java
@@ -295,9 +295,9 @@ public class ScanApi extends DefaultTask {
                 /*
                  * Class declaration.
                  */
-                Names annotationNames = toNames(readClassAnnotationsFor(classInfo));
-                if (!annotationNames.visible.isEmpty()) {
-                    writer.append(asAnnotations(annotationNames.visible));
+                List<String> visibleAnnotations = toNames(readClassAnnotationsFor(classInfo)).visible.stream().map(ScanApi::removePackageName).collect(toList());
+                for(String annotation: visibleAnnotations) {
+                    writer.println("@" + annotation);
                 }
                 writer.append(Modifier.toString(modifiers & CLASS_MASK));
                 writer.append(" class ").print(classInfo);
@@ -313,9 +313,9 @@ public class ScanApi extends DefaultTask {
                 /*
                  * Interface declaration.
                  */
-                Names annotationNames = toNames(readInterfaceAnnotationsFor(classInfo));
-                if (!annotationNames.visible.isEmpty()) {
-                    writer.append(asAnnotations(annotationNames.visible));
+                List<String> visibleAnnotations = toNames(readClassAnnotationsFor(classInfo)).visible.stream().map(ScanApi::removePackageName).collect(toList());
+                for(String annotation: visibleAnnotations) {
+                    writer.println("@" + annotation);
                 }
                 writer.append(Modifier.toString(modifiers & INTERFACE_MASK));
                 writer.append(" interface ").print(classInfo);
@@ -334,18 +334,25 @@ public class ScanApi extends DefaultTask {
                         && isValid(method.getAccessFlags(), METHOD_MASK) // Excludes bridge and synthetic methods
                         && !hasInternalAnnotation(method.getAnnotationNames()) // Excludes methods annotated as @CordaInternal
                         && !isKotlinInternalScope(method)) {
-                    writer.append("  ").println(filterAnnotationsFor(method));
+                    for(String annotation: getAnnotationsFor(method)) {
+                        writer.append("  @").println(annotation);
+                    }
+                    writer.append("  ").println(removeAnnotationsFor(method));
                 }
             }
         }
 
-        private void writeFields(PrintWriter output, List<FieldInfo> fields) {
+        private void writeFields(PrintWriter writer, List<FieldInfo> fields) {
             sort(fields);
             for (FieldInfo field : fields) {
                 if (isVisible(field.getAccessFlags())
                         && isValid(field.getAccessFlags(), FIELD_MASK)
                         && !hasInternalAnnotation(field.getAnnotationNames())) {
-                    output.append("  ").println(filterAnnotationsFor(field));
+                    List<String> annotations = getAnnotationsFor(field);
+                    for(String annotation: annotations) {
+                        writer.append("  @").println(annotation);
+                    }
+                    writer.append("  ").println(removeAnnotationsFor(field));
                 }
             }
         }
@@ -395,31 +402,41 @@ public class ScanApi extends DefaultTask {
                 .collect(toList());
         }
 
-        private MethodInfo filterAnnotationsFor(MethodInfo method) {
+        private MethodInfo removeAnnotationsFor(MethodInfo method) {
             return new MethodInfo(
                 method.getClassName(),
                 method.getMethodName(),
                 method.getAccessFlags(),
                 method.getTypeDescriptor(),
-                method.getAnnotationNames().stream()
-                    .filter(this::isVisibleAnnotation)
-                    .sorted()
-                    .collect(toList())
+                emptyList()
             );
         }
 
-        private FieldInfo filterAnnotationsFor(FieldInfo field) {
+        private FieldInfo removeAnnotationsFor(FieldInfo field) {
             return new FieldInfo(
                 field.getClassName(),
                 field.getFieldName(),
                 field.getAccessFlags(),
                 field.getTypeDescriptor(),
                 field.getConstFinalValue(),
-                field.getAnnotationNames().stream()
-                    .filter(this::isVisibleAnnotation)
-                    .sorted()
-                    .collect(toList())
+                emptyList()
             );
+        }
+
+        private List<String> getAnnotationsFor(List<String> annotationNames) {
+            return annotationNames.stream()
+                .filter(this::isVisibleAnnotation)
+                .map(ScanApi::removePackageName)
+                .sorted()
+                .collect(toList());
+        }
+
+        private List<String> getAnnotationsFor(MethodInfo method) {
+            return getAnnotationsFor(method.getAnnotationNames());
+        }
+
+        private List<String> getAnnotationsFor(FieldInfo field) {
+            return getAnnotationsFor(field.getAnnotationNames());
         }
 
         private boolean isVisibleAnnotation(String annotationName) {
@@ -429,6 +446,13 @@ public class ScanApi extends DefaultTask {
         private boolean hasInternalAnnotation(Collection<String> annotationNames) {
             return annotationNames.stream().anyMatch(internalAnnotations::contains);
         }
+    }
+
+    private static String removePackageName(String className) {
+        if (!className.contains(".")) {
+            return className;
+        }
+        return className.substring(className.lastIndexOf(".") + 1);
     }
 
     private static <T extends Comparable<? super T>> List<T> ordering(List<T> list) {
@@ -452,8 +476,8 @@ public class ScanApi extends DefaultTask {
         return items.stream().map(ClassInfo::toString).collect(joining(", "));
     }
 
-    private static String asAnnotations(Collection<String> items) {
-        return items.stream().collect(joining(" @", "@", " "));
+    private static List<String> asAnnotations(Collection<String> items) {
+        return items.stream().map(ScanApi::removePackageName).collect(toList());
     }
 
     private static boolean isApplicationClass(String typeName) {
--- a/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
+++ b/api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
@@ -44,9 +44,18 @@ public class AnnotatedClassTest {
 
         Path api = pathOf(testProjectDir, "build", "api", "annotated-class.txt");
         assertThat(api).isRegularFile();
-        assertThat(Files.readAllLines(api)).containsOnlyOnce(
-            "@net.corda.annotation.AlsoInherited @net.corda.annotation.IsInherited @net.corda.annotation.NotInherited public class net.corda.example.HasInheritedAnnotation extends java.lang.Object",
-            "@net.corda.annotation.AlsoInherited @net.corda.annotation.IsInherited public class net.corda.example.InheritingAnnotations extends net.corda.example.HasInheritedAnnotation"
-        );
+        assertEquals(
+            "@AlsoInherited\n" +
+            "@IsInherited\n" +
+            "@NotInherited\n" +
+            "public class net.corda.example.HasInheritedAnnotation extends java.lang.Object\n" +
+            "  public <init>()\n" +
+            "##\n" +
+            "@AlsoInherited\n" +
+            "@IsInherited\n" +
+            "public class net.corda.example.InheritingAnnotations extends net.corda.example.HasInheritedAnnotation\n" +
+            "  public <init>()\n" +
+            "##", CopyUtils.toString(api));
+
     }
 }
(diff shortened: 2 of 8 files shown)


anthonykeenan:
After a bit of testing, I realised that if there were any annotations that appeared before `@DoNotImplement` it would cause issues if those annotations were changed, so I've made a modification to always put the `@DoNotImplement` annotation first if it exists, which sorts the issue.

api-scanner/src/test/java/net/corda/plugins/AnnotatedInterfaceTest.java
```diff
@@ -44,9 +44,13 @@ public void testAnnotatedInterface() throws IOException {
...
-            "@net.corda.annotation.AlsoInherited @net.corda.annotation.IsInherited @net.corda.annotation.NotInherited public interface net.corda.example.HasInheritedAnnotation",
-            "@net.corda.annotation.AlsoInherited @net.corda.annotation.IsInherited public interface net.corda.example.InheritingAnnotations extends net.corda.example.HasInheritedAnnotation"
-        );
+        assertEquals(
+            "@AlsoInherited\n" +
+            "@IsInherited\n" +
+            "@NotInherited\n" +
+            "public interface net.corda.example.HasInheritedAnnotation\n" +
+            "##\n" +
+            "public interface net.corda.example.InheritingAnnotations extends net.corda.example.HasInheritedAnnotation\n" +
```
chrisr3: The _whole point_ of this test is that interface `InheritingAnnotations` should inherit `@IsInherited` and `@AlsoInherited`. Did you notice that you'd changed the expectation here? See the above comment that interfaces should be using `readInterfaceAnnotationsFor()`.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -450,8 +483,8 @@ private static String stringOf(Collection<ClassInfo> items) {
         return items.stream().map(ClassInfo::toString).collect(joining(", "));
     }
 
-    private static String asAnnotations(Collection<String> items) {
-        return items.stream().collect(joining(" @", "@", " "));
```
chrisr3: I think I'd prefer
```java
private static String asAnnotations(Collection<String> items) {
    return items.stream().map(ScanApi::removePackageName)
        .collect(joining(System.lineSeparator() + '@', "@", System.lineSeparator()));
}
```

api-scanner/src/test/java/net/corda/plugins/AnnotatedClassTest.java
```diff
@@ -44,9 +44,18 @@ public void testAnnotatedClass() throws IOException {
 
         Path api = pathOf(testProjectDir, "build", "api", "annotated-class.txt");
         assertThat(api).isRegularFile();
-        assertThat(Files.readAllLines(api)).containsOnlyOnce(
-            "@net.corda.annotation.AlsoInherited @net.corda.annotation.IsInherited @net.corda.annotation.NotInherited public class net.corda.example.HasInheritedAnnotation extends java.lang.Object",
-            "@net.corda.annotation.AlsoInherited @net.corda.annotation.IsInherited public class net.corda.example.InheritingAnnotations extends net.corda.example.HasInheritedAnnotation"
-        );
+        assertEquals(
```
chrisr3: `assertj` has some rather nice matchers for `List` object so that we don't need to match the entire file:
```java
assertThat(Files.readAllLines(api, UTF_8))
    .containsSequence(
        "@AlsoInherited",
        "@IsInherited",
        "@NotInherited",
        "public class net.corda.example.HasInheritedAnnotation extends java.lang.Object")
    .containsSequence(
        "@AlsoInherited",
        "@IsInherited",
        "public class net.corda.example.InheritingAnnotations extends net.corda.example.HasInheritedAnnotation");
```

api-scanner/src/test/java/net/corda/plugins/AnnotatedFieldTest.java
```diff
@@ -44,9 +44,20 @@ public void testAnnotatedField() throws IOException {
 
         Path api = pathOf(testProjectDir, "build", "api", "annotated-field.txt");
         assertThat(api).isRegularFile();
-        assertThat(Files.readAllLines(api)).containsOnlyOnce(
-            "public class net.corda.example.HasAnnotatedField extends java.lang.Object",
-            "  @net.corda.example.A @net.corda.example.B @net.corda.example.C public static final String ANNOTATED_FIELD = \"<string-value>\""
-        );
+        assertEquals(
```
chrisr3: Again, no need for the entire file:
```java
assertThat(Files.readAllLines(api, UTF_8))
    .containsSequence(
        "public class net.corda.example.HasAnnotatedField extends java.lang.Object",
        "  public <init>()",
        "  @A",
        "  @B",
        "  @C",
        "  public static final String ANNOTATED_FIELD = \"<string-value>\"\n");
```

api-scanner/src/test/java/net/corda/plugins/AnnotatedMethodTest.java
```diff
@@ -44,9 +43,23 @@ public void testAnnotatedMethod() throws IOException {
 
         Path api = pathOf(testProjectDir, "build", "api", "annotated-method.txt");
         assertThat(api).isRegularFile();
-        assertThat(Files.readAllLines(api)).containsOnlyOnce(
-            "public class net.corda.example.HasAnnotatedMethod extends java.lang.Object",
-            "  @net.corda.example.A @net.corda.example.B @net.corda.example.C public void hasAnnotation()"
-        );
+        assertEquals(
```
chrisr3: Again, I think that this test is more readable if we use AssertJ's `containsSequence()` now. (And oops, it looks like I missed the `UTF_8` for `Files.readAllLines()`!)

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -332,18 +341,25 @@ private void writeMethods(PrintWriter writer, List<MethodInfo> methods) {
                         && isValid(method.getAccessFlags(), METHOD_MASK) // Excludes bridge and synthetic methods
                         && !hasInternalAnnotation(method.getAnnotationNames()) // Excludes methods annotated as @CordaInternal
                         && !isKotlinInternalScope(method)) {
-                    writer.append("  ").println(filterAnnotationsFor(method));
+                    for(String annotation: getAnnotationsFor(method)) {
```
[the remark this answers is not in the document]
chrisr3: I suppose I'm thinking of something like this:
```
println(writer, filterAnnotationsFor(method), "  ");
```
where the third parameter is the indentation string that we need to prefix each line with. I hope this would be prettier than a `toString()`-like function that is aware of both indentation-width and line separators :nauseated_face:.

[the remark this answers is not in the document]
chrisr3: Heh, or maybe even make a `PrintWriter` sub-class with the following extra methods:
```java
void println(Methodinfo method, String prefix);
void println(FieldInfo field, String prefix);
```
I think `ScanApi.java` is becoming rather "code-heavy", so maybe a custom `PrintWriter` class would help?

api-scanner/src/main/java/net/corda/plugins/ApiPrintWriter.java
```diff
@@ -0,0 +1,123 @@
...
+        super.append(asAnnotations(method.getAnnotationNames(), indentation));
+        super.append(indentation);
+        if (method.getModifiersStr() != null) {
+            super.append(method.getModifiersStr()).append(" ");
+        }
+        if (!method.isConstructor()) {
+            super.append(removeQualifierFromBaseTypes(method.getResultTypeStr())).append(" ");
+        }
+        super.append(method.getMethodName()).append("(");
+        List<TypeSignature> paramTypes = method.getTypeSignature().getParameterTypeSignatures();
```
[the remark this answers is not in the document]
anthonykeenan: Just doing a replace on the variable name won't work if you have:
``fun action(vararg arg: Array<String>)``
So I've kept the method of replacing the brackets the same but used the streaming method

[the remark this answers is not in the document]
anthonykeenan: I've also updated the vararg test to use that case

api-scanner/src/main/java/net/corda/plugins/ApiPrintWriter.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private static String stringOf(Collection<ClassInfo> items) {
+        return items.stream().map(ClassInfo::getClassName).collect(joining(", "));
+    }
+
+    private static String removeQualifierFromBaseTypes(String className) {
+        return className.replace("java.lang.", "");
+    }
+
+    public void println(ClassInfo classInfo, int filteredModifiers, List<String> filteredAnnotations) {
```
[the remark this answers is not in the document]
anthonykeenan: agreed

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -275,75 +281,42 @@ private void writeApis(PrintWriter writer, ScanResult result) {
...
 
-        private void writeClass(PrintWriter writer, ClassInfo classInfo, int modifiers) {
+        private void writeClass(ApiPrintWriter writer, ClassInfo classInfo) {
             if (classInfo.isAnnotation()) {
-                /*
-                 * Annotation declaration.
-                 */
-                writer.append(Modifier.toString(modifiers & INTERFACE_MASK));
-                writer.append(" @interface ").print(classInfo);
+                writer.println(classInfo, classInfo.getClassRef().getModifiers() & INTERFACE_MASK, new ArrayList<>());
```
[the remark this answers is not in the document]
anthonykeenan: ..because that's how I'd do it in c# ;-)

api-scanner/src/main/java/net/corda/plugins/ApiPrintWriter.java
```diff
@@ -0,0 +1,128 @@
...
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toCollection;
+
+public class ApiPrintWriter extends PrintWriter {
+    ApiPrintWriter(File file, String encoding) throws FileNotFoundException, UnsupportedEncodingException {
+        super(file, encoding);
+    }
+
+    public void println(ClassInfo classInfo, int modifierMask, List<String> filteredAnnotations) {
+        super.append(asAnnotations(filteredAnnotations, ""));
```
chrisr3: General comment: Why `super.append()` instead of just `append()`? You haven't overridden the `append` function in this class, but you'd presumably want to invoke your own own rather than the parent one, if you had.

api-scanner/src/main/java/net/corda/plugins/ApiPrintWriter.java
```diff
@@ -0,0 +1,128 @@
...
+            /*
+             * Interface declaration.
+             */
+            super.append(" interface ").print(classInfo.getClassName());
+            Set<ClassInfo> superinterfaces = classInfo.getDirectSuperinterfaces();
+            if (!superinterfaces.isEmpty()) {
+                super.append(" extends ").print(stringOf(superinterfaces));
+            }
+        }
+        super.println();
```
chrisr3: Similarly, why `super.println()` instead of just `println()`?

api-scanner/src/main/java/net/corda/plugins/ApiPrintWriter.java
```diff
@@ -0,0 +1,128 @@
...
+        if (method.isVarArgs() && !paramTypes.isEmpty()) {
+            String vararg = paramTypes.removeLast();
+            paramTypes.add(vararg.substring(0, vararg.length() - 2) + "...");
+        }
+        super.append(paramTypes.stream().collect(joining(", ")));
+        super.println(')');
+    }
+
+    public void println(FieldInfo field, String indentation) {
+        super.append(asAnnotations(field.getAnnotationNames(), indentation));
```
chrisr3: You could chain _all seven_ of these `append()` calls.

api-scanner/src/test/java/net/corda/plugins/KotlinAnnotationsTest.java
```diff
@@ -17,29 +20,57 @@
     @Rule
     public TestRule rules = RuleChain.outerRule(testProjectDir).around(testProject);
 
+    private String[] expectedClassWithDeprecatedFunctions = new String[] {
+        "public final class net.corda.example.HasDeprecatedFunctions extends java.lang.Object",
```
chrisr3: Constant arrays in Java:
```java
private static final String[] NAME = {
    "--- line ---",
    "--- line ---",
    ...
}
```

api-scanner/src/test/java/net/corda/plugins/KotlinAnnotationsTest.java
```diff
@@ -7,8 +7,11 @@
 import org.junit.rules.TestRule;
 
 import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
 
-import static org.junit.Assert.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
```
chrisr3: Unused imports...

api-scanner/src/test/java/net/corda/plugins/KotlinAutoGeneratedClassTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Rule
+    public TestRule rules = RuleChain.outerRule(testProjectDir).around(testProject);
+
+    @Test
+    public void testKotlinAutoGeneratedClassMethod() throws IOException {
+        System.out.println("Started output");
+        System.out.println(CopyUtils.toString(testProject.getApi()));
+        System.out.println("Finished output");
+        assertEquals("public final class net.corda.example.ContainsAutoGeneratedClass extends java.lang.Object\n" +
```
chrisr3: Did you mean to leave those `System.out.println()` lines in?

api-scanner/src/test/resources/kotlin-auto-generated-class/kotlin/net/corda/example/ContainsAutoGeneratedClass.kt
```diff
@@ -0,0 +1,19 @@
...
+        return object: PrivateInnerClass() {
+            fun run() = action()
+        }
+    }
+
+    private val timer = PrivateInnerClass()
+
+    fun methodContainingInlinedFunction() {
+        runSomething {
+            System.out.println()
```
chrisr3: As a Kotlin example, this should probably be `println()` rather than `System.out.println()`. But that's just me being _really_ picky :smiling_imp:.

api-scanner/src/test/resources/kotlin-vararg-method/kotlin/net/corda/example/KotlinVarargMethod.kt
```diff
@@ -1,5 +1,5 @@
 package net.corda.example
 
 interface KotlinVarargMethod {
-    fun action(vararg arg: Any?)
+    fun action(vararg arg: Array<String>)
```
chrisr3: I'm not sure this makes sense - your `action` function is asking for a variable number of `String` **arrays**! Did you mean:
```kotlin
fun action(vararg arg: String)
```
Which means that the API expectation would be:
```java
public abstract void action(String...)
```
I think our test cases should be realistic. If you _want_ a pathological case (which wouldn't be unreasonable) then that should be in addition to the realistic case.

anthonykeenan: No I did mean to do that - your suggestion in one of the earlier comments was to do a `.replace("[]", "...")` and I hypothesised that if someone had a vararg list of lists parameter you would end up with `String......` instead of `String[]...`. But I agree it is... out there. I'll add the more basic case back in.

api-scanner/src/test/java/net/corda/plugins/KotlinAutoGeneratedClassTest.java
anthonykeenan: Nope, I've deleted

api-scanner/src/test/resources/kotlin-auto-generated-class/kotlin/net/corda/example/ContainsAutoGeneratedClass.kt
anthonykeenan: Like my mum always says, if a job's worth doing, it's worth doing right.

anthonykeenan:
Back over to you

api-scanner/src/test/java/net/corda/plugins/KotlinVarargMethodTest.java
```diff
@@ -24,10 +19,22 @@
...
+        "##"
+    };
+
+    private static final String[] expectedInterfaceWithArrayVarargMethod = {
+        "public interface net.corda.example.KotlinVarargMethod",
+        "  public abstract void action(int...)",
+        "##"
+    };
+
     @Test
```
chrisr3: I think you have `expectedInterfaceWithVarargMethod` and `expectedInterfaceWithArrayVarargMethod` back-to-front :wink:

api-scanner/src/test/java/net/corda/plugins/KotlinVarargMethodTest.java
```diff
@@ -1,21 +1,16 @@
...
 import java.io.IOException;
-import java.nio.file.Path;
+import java.nio.file.Files;
 
-import static net.corda.plugins.CopyUtils.*;
-import static org.assertj.core.api.Assertions.*;
-import static org.gradle.testkit.runner.TaskOutcome.SUCCESS;
-import static org.junit.Assert.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
```
chrisr3: Unused import.

PR #58, opened 2018-04-30
CORDA-1410: Fix Jolokia runtime exception upon missing Jolokia jar file in driver…

josecoll:
…s directory.

Minor code refactoring/clean-up.
Updated old 3.0-SNAPSHOT references in sample tests and plugins.

### DIFF
--- a/cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
+++ b/cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
@@ -17,7 +17,7 @@ class CordformTest {
     private lateinit var buildFile: File
 
     private companion object {
-        const val cordaFinanceJarName = "corda-finance-3.1-corda"
+        const val cordaFinanceJarName = "corda-finance-3.0-SNAPSHOT"
         const val localCordappJarName = "locally-built-cordapp"
         const val notaryNodeName = "Notary Service"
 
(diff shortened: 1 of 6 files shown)


cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
```diff
@@ -17,7 +17,7 @@ class CordformTest {
     private lateinit var buildFile: File
 
     private companion object {
-        const val cordaFinanceJarName = "corda-finance-3.0-SNAPSHOT"
+        const val cordaFinanceJarName = "corda-finance-3.1-corda"
```
chrisr3: This isn't a great place for the version number - isn't this dependent on `$project.version`?

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordapp.gradle
```diff
@@ -1,7 +1,7 @@
 buildscript {
     ext {
         corda_group = 'net.corda'
-        corda_release_version = '3.0-SNAPSHOT' // TODO: Set to 3.0.0 when Corda 3 is released
+        corda_release_version = '3.1-corda'
```
chrisr3: Do these Gradle tests actually work yet? These `build.gradle` files are test resources, and so we can expand properties via a `processTestResources {}` block.

josecoll:
@chrisr3 CordForm tests do not seem to work before or after updating the corda version references. I will back out my changes related to this as they are not relevant to the code fix.

josecoll:
@Clintonio any idea why the CordFormTest do not currently work?
Also worth noting that there are numerous references to Jolokia 1.3.7 (the version was recently updated in Corda OS to 1.5.0).

PR #59, opened 2018-05-01
CORDA-1389: Delete the root nodes directory before initialising the nodes

shamsasari:
This is to enable multiple calls to deployNodes without doing a gradle clean.

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
@@ -119,6 +119,7 @@ open class Baseform : DefaultTask() {
                 "'directory' cannot be used when 'definitionClass' is specified. Use CordformDefinition.nodesDirectory instead."
             }
             directory = cd.nodesDirectory
+            deleteRootDir()
             val cordapps = cd.cordappDependencies
             cd.nodeConfigurers.forEach {
                 val node = node { }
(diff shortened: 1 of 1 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -119,6 +119,7 @@ open class Baseform : DefaultTask() {
                 "'directory' cannot be used when 'definitionClass' is specified. Use CordformDefinition.nodesDirectory instead."
             }
             directory = cd.nodesDirectory
+            project.delete(directory)
```
Clintonio: Please log this action. It will be very useful for debugging to know we're erasing a directory.

cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -143,11 +145,11 @@ open class Baseform : DefaultTask() {
         val networkBootstrapperClass = loadNetworkBootstrapperClass()
         val networkBootstrapper = networkBootstrapperClass.newInstance()
         val bootstrapMethod = networkBootstrapperClass.getMethod("bootstrap", Path::class.java, List::class.java).apply { isAccessible = true }
-        // Call NetworkBootstrapper.bootstrap
+        // Create a list of all cordapps used in this network and pass it to the bootstrapper.
```
Clintonio: This comment is too much of a what, not why. The below comment is useful because it provides a way to mentally shortcut trying to read reflection code but this comment isn't.

PR #68, opened 2018-05-11
CORDA-1461 move devMode to CorformNode to be accessible programatically

tudor-malene:
CORDA-1461 move devMode to CorformNode to be accessible programatically

### DIFF
--- a/cordform-common/src/main/java/net/corda/cordform/CordformNode.java
+++ b/cordform-common/src/main/java/net/corda/cordform/CordformNode.java
@@ -107,6 +107,14 @@ public class CordformNode implements NodeDefinition {
     }
 
     /**
+     *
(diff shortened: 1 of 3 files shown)


cordform-common/src/main/java/net/corda/cordform/CordformNode.java
```diff
@@ -106,6 +106,14 @@ public void p2pAddress(String p2pAddress) {
         setValue("p2pAddress", p2pAddress);
     }
 
+    /**
+     *
+     * @param devMode -
```
Clintonio: Javadoc is a bit sparse. If we're going to have it let's fill it in.

tudor-malene:
back to you @Clintonio

PR #70, opened 2018-05-17
Use default parent classpath when getting bootstrapper class

anthonykeenan:


chrisr3:
Please don't release just yet because I'm also fixing the `CordformDefinition`.

anthonykeenan:
No problems - I'll merge but leave the release for you

PR #71, opened 2018-05-17
Isolate CordformDefinition subclasses from the plugins classpath.

chrisr3:
Ensure that subclasses of `CordformDefinition` cannot access anything on the Gradle plugins classpath.


cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -172,4 +174,36 @@ open class Baseform : DefaultTask() {
...
+     * Note that this classloader's parent is the system classloader.
+     */
+    class CordformClassLoader(
+        urls: Array<URL>,
+        private val logger: Logger,
+        cordform: Class<*> = CordformDefinition::class.java // parameterised for testing
+)   : URLClassLoader(urls) {
+        private val pluginClassloader: ClassLoader = cordform.classLoader
+        private val prefixes: List<String>
+                      = listOf(cordform.`package`.name + '.', "com.typesafe.config.", "org.jetbrains.kotlin.")
```
Clintonio: Please also add a comment to the cordform project's build.gradle to tell people about this line needing to be updated if new packages can be added.

PR #72, opened 2018-05-17
Refactored distribution info in CorDappPlugin.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1501

### DIFF
--- a/cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
+++ b/cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
@@ -65,7 +65,7 @@ class CordformTest {
         createBuildFile(buildFileResourceName)
         return GradleRunner.create()
                 .withProjectDir(testProjectDir.root)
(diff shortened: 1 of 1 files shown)


cordformation/src/test/kotlin/net/corda/plugins/CordformTest.kt
```diff
@@ -65,7 +65,7 @@ class CordformTest {
         createBuildFile(buildFileResourceName)
         return GradleRunner.create()
                 .withProjectDir(testProjectDir.root)
-                .withArguments("deployNodes", "-s", "--info", "-g", testGradleUserHome)
+                .withArguments("deployNodes", "-s", "--getInfo", "-g", testGradleUserHome)
```
chrisr3: I think this must be a `find/replace` mistake :wink:

sollecitom: I should stop trusting intellij to rename things :) thanks @chrisr3 !

PR #74, opened 2018-05-31
ENT-1871 Rebrand R3Corda to Corda Enterprise 

josecoll:
Jar file has been renamed from
`corda-R3.<version>-<component>.jar`
to
`corda-enterprise.<version>-<component>.jar`  

Other minor fix to incorrect package exclusion path.


tlil:
@josecoll Should update the changelog :-)

PR #81, opened 2018-06-19
[CORDA-1649]: Add default `quasar_version` and `quasar_group` to quasar-utils plugin, so it's not require to specify them manually.

sollecitom:
https://r3-cev.atlassian.net/browse/CORDA-1649

### DIFF
--- a/quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
+++ b/quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
@@ -10,8 +10,8 @@ import org.gradle.api.tasks.JavaExec
  */
 class QuasarPlugin implements Plugin<Project> {
 
-    private defaultGroup = "co.paralleluniverse"
-    private defaultVersion = "0.7.10"
+    static defaultGroup = "co.paralleluniverse"
+    static defaultVersion = "0.7.10"
 
     @Override
     void apply(Project project) {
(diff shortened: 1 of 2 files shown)


quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
```diff
@@ -9,14 +9,21 @@ import org.gradle.api.tasks.JavaExec
...
     void apply(Project project) {
         Utils.createRuntimeConfiguration("cordaRuntime", project)
 
         project.configurations.create("quasar")
+
+        def quasarGroup = project.rootProject.hasProperty("quasar_group") ? project.rootProject.ext.quasar_group : defaultGroup
+        def quasarVersion = project.rootProject.hasProperty("quasar_version") ? project.rootProject.ext.quasar_version : defaultVersion
 //        To add a local .jar dependency:
 //        project.dependencies.add("quasar", project.files("${project.rootProject.projectDir}/lib/quasar.jar"))
-        project.dependencies.add("quasar", "${project.rootProject.ext.quasar_group}:quasar-core:${project.rootProject.ext.quasar_version}:jdk8@jar")
```
Clintonio: Could you allow this to be configurable still?

eg;

project.rootProject.ext.quasar_group || quasarGroup
project.rootProject.ext.quasar_version || quasarVersion

There's always the possibility that we'll release on a broken version and need to allow the users to upgrade without changing plugins version.

Ideally we'd provide a configurable extension but that's too much work for this fix.

Clintonio: I'm blind.

PR #83, opened 2018-06-19
CORDA-1647: Break circular dependency on publish-utils

Clintonio:
This PR essentially decomposes the logic of publish utils into the root build.gradle. This is the better, more gradle-y way of doing this. We have been mis-using gradle build files for some time ignoring the power of the ability to configure a subset of projects. This is all the publish utils plugins does.

This PR will lead way to remove publish utils from this repo in leiu of a single solution within the main build.gradle or buildSrc. 

Also;

* Added new gradle properties flags (-D) to configure the bintray details. 
* Added new gradle properties flags (-D) to configure the artifactory details.

Also:

* Removed automatic publishing from the gradle plugin plugin
* Upgraded to gradle 4.8 to fix an issue with the pom configuration (it was a hack before)


build.gradle
```diff
@@ -51,52 +46,102 @@ allprojects {
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    bintray {
+        user = System.getenv('CORDA_BINTRAY_USER') || System.getProperty('corda.bintray.user')
+        key = System.getenv('CORDA_BINTRAY_KEY') || System.getPropert('corda.bintray.key')
```
chrisr3: Ooopsee!! Typo! I think you meant
```groovy
System.getProperty('corda.bintray.key')
```

build.gradle
```diff
@@ -51,52 +46,102 @@ allprojects {
...
-    gpgPassphrase = System.getenv('CORDA_BINTRAY_GPG_PASSPHRASE')
-    publications = ['cordformation', 'quasar-utils', 'cordform-common', 'api-scanner', 'cordapp']
-    license {
-        name = 'Apache-2.0'
-        url = 'https://www.apache.org/licenses/LICENSE-2.0'
-        distribution = 'repo'
+def publishProjects = project.subprojects - project('api-scanner:annotations')
+
+configure(publishProjects) { subproject ->
+    apply plugin: 'java'
```
chrisr3: I'm slightly concerned that you're always applying the `java` plugin, even for plugins written in Kotlin or Groovy, but I suppose it makes no material difference to how everything is compiled.

build.gradle
Clintonio: I clearly rushed this before I left...

build.gradle
Clintonio: Yeah, we do it in the main corda repo too. Maybe we can fine grain it later but this was needed to get the sourceJar and docsJar tasks working.

build.gradle
```diff
@@ -26,14 +23,12 @@ buildscript {
     }
 
     dependencies {
-        classpath "net.corda.plugins:publish-utils:$gradle_plugins_version"
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jfrog.buildinfo:build-info-extractor-gradle:$artifactory_plugin_version"
```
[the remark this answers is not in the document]
Clintonio: Updated. Also removed constants.properties since it existed because of the hack this PR fixes. Yay.

build.gradle
```diff
@@ -51,52 +41,102 @@ allprojects {
...
+                gpg {
+                    sign = true
+                    passphrase = System.getenv('CORDA_BINTRAY_GPG_PASSPHRASE')
+                }
+            }
+        }
     }
+
+    task install(dependsOn: 'publishToMavenLocal')
 }
```
chrisr3: Move this task underneath the `sourceJar` and `javadocJar` tasks.

Clintonio:
I'll update TC when this PR is approved. That way there's nearly 0 time where the build logic differs from the code.

build.gradle
```diff
@@ -79,24 +124,19 @@ artifactory {
...
+            if (publishProjects.contains(project)) {
+                publications(project.name)
             }
         }
     }
 }
 
 wrapper {
-    gradleVersion = "4.4.1"
+    gradleVersion = "4.8"
```
chrisr3: _No, **not this**!_ We want [Gradle 4.8.1](https://github.com/gradle/gradle/releases/tag/v4.8.1)!!!!!!! :smiling_imp:

Clintonio: Hah, perfectly bad timing :D

chrisr3:
I'm not convinced that the wrapper JAR has changed between Gradle 4.8 and Gradle 4.8.1, but anyway...

PR #85, opened 2018-07-03
[Corda 1716] Dockerform task: Add sshdPort to docker-compose.yml.

florian-f:
https://r3-cev.atlassian.net/browse/CORDA-1716

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,8 @@
 
 * `cordformation`: The `sshd` port entry in `node` produces an entry in `node.conf` that matches the config API.
 
+* The Dockerform task exposes the sshdPort in docker-compose.yml
+
 ### Version 4.0.26
 
 ### Version 4.0.25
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -6,6 +6,8 @@
 
 * `cordformation`: The `sshd` port entry in `node` produces an entry in `node.conf` that matches the config API.
 
+* The Dockerform task exposes the sshdPort in docker-compose.yml
```
chrisr3: This changelog entry is corrupt; not least because we're currently on 4.0.30-SNAPSHOT but also because there are two lines saying:

```
### Version 4.0.26
```

cordform-common/src/main/java/net/corda/cordform/CordformNode.java
```diff
@@ -43,6 +43,13 @@ public String getName() {
 
     public int getRpcPort() { return rpcPort; }
 
+    private Integer sshdPort;
```
chrisr3: The "getter" is returning `Integer` but the setting is receiving `int`. Does Java/Kotlin still recognise `sshdPort` as a property, despite their *non-identical* types? (I think I might have been bitten by something like this in the past.)

Or put another way, why not pass `Integer` into the setter too? Might people want to _unset_ an optional field?

PR #88, opened 2018-07-12
Updated to provide Tmux support on Linux

kid101:

# PR Checklist:

- [ ] Have you run the unit, integration and smoke tests as described here? https://docs.corda.net/head/testing.html
- [ ] If you added/changed public APIs, did you write/update the JavaDocs?
- [ ] If the changes are of interest to application developers, have you added them to the changelog, and potentially release notes?
- [ ] If you are contributing for the first time, please read the agreement in CONTRIBUTING.md now and add to this Pull Request that you agree to it.

Thanks for your code, it's appreciated! :)

### DIFF
--- a/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
+++ b/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
@@ -104,8 +104,7 @@ private abstract class JavaCommand(
             try {
                 val jolokiaJvmArgs = "-javaagent:drivers/$jolokiaJar=port=$monitoringPort,logHandlerClass=net.corda.node.JolokiaSlf4jAdapter"
                 jvmArgs.add(jolokiaJvmArgs)
-            }
-            catch(e: RuntimeException) {
+            } catch (e: RuntimeException) {
                 println("${e.message}. Continuing without Jolokia instrumentation ...")
             }
         }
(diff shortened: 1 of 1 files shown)


kid101:
It's not perfect, so let me know what changes would be required, will do as per your feedback.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -124,52 +123,59 @@ private abstract class JavaCommand(
 }
 
 private class HeadlessJavaCommand(jarType: JarType, dir: File, debugPort: Int?, monitoringPort: Int?, args: List<String>, jvmArgs: List<String>)
-    : JavaCommand(jarType.jarName, dir, debugPort, monitoringPort, jarType.enableJolokia, dir.name, jarType.headlessArgs + args, jvmArgs) {
+    : JavaCommand(jarType.jarName, dir, debugPort, monitoringPort, jarType.enableJolokia, dir.name, if(isTmux()) args else jarType.headlessArgs + args, jvmArgs) {
```
Clintonio: This check can be replaced entirely by modifying line 31 from

val headless = GraphicsEnvironment.isHeadless() || args.contains(HEADLESS_FLAG)

to

     val headless = ((!isTmux() && GraphicsEnvironment.isHeadless()) || args.contains(HEADLESS_FLAG))

That way the any tmux terminal isn't automatically set to headless and will thus behave normally. This will also mean the headless flag still works as intended (by disabling the shell).

Similar the conditional on line 129 can be reverted.

kid101: yeah then I won't even need any of this  function `getCommandAsPerOS` actually just a ` val headless = ((!isTmux() && GraphicsEnvironment.isHeadless()) || args.contains(HEADLESS_FLAG))` will do the trick,

kid101: If I just undo everything and make ` val headless = ((!isTmux() && GraphicsEnvironment.isHeadless()) || args.contains(HEADLESS_FLAG))` it does solve the problem and everything works fine, I don't think there is a need to abstract and create this function as well `getCommandAsPerOS` what do you think?

Clintonio:
Thank you for your contribution @kid101. This will appear in Gradle Plugins V4.0.28 and beyond.

PR #91, opened 2018-07-17
Added gradle version check to the cordapp plugin

ivanr3:
This PR adds verfification of Gradle version that CorDapp is built with against the Gradle Wrapper config if one exists. This verification aims to save users from build-related issues if they use a wrong version of Gradle.

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
@@ -5,6 +5,9 @@ import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.Dependency
 import org.gradle.jvm.tasks.Jar
 import java.io.File
+import java.io.FileInputStream
+import java.util.*
+import java.util.regex.Pattern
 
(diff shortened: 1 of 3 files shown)


cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -34,6 +39,34 @@ class CordappPlugin : Plugin<Project> {
         configureCordappJar(project)
     }
 
+    /**
+     * Verifies that the version of Gradle the project is built with matches gradle-wrapper config if one exists
+     */
+    private fun verifyGradleVersion(project: Project) {
+        val gradleWrapperConfig = File(project.rootProject.projectDir!!.absolutePath + "/gradle/wrapper/gradle-wrapper.properties")
```
Clintonio: You can replace this with; 

    project.gradle.version

chrisr3:
You will need to rebase this PR against the `master` branch because I've just updated TC also to build the `jar-filter` plugin.

ivanr3:
@chrisr3 Rebased

PR #92, opened 2018-07-18
CORDA-1816: Migrate the JarFilter plugin from Corda

chrisr3:
There is no longer any reason for the `jar-filter` plugin to live in Corda's `buildSrc` directory because:
- It is now required by both Enterprise and Open Source
- Its hard dependency on Corda's Kotlin version has been removed by shading Kotlin's metadata support.
- Its extensive unit test coverage adding several minutes to everyone's build times.


build.gradle
```diff
@@ -41,7 +44,8 @@ allprojects {
     }
 }
 
-def publishProjects = project.subprojects - project('api-scanner:annotations')
+// Only the top-level projects are plugins.
```
Clintonio: Can you explain the line with a why.

PR #104, opened 2018-08-09
CORDA-1889: Update CordFormation to make deploying itself as a CorDapp optional.

chrisr3:
A CordFormation project does not necessarily build a CorDapp itself. Add a new
```
nodeDefaults {}
```
block to the task, which defines initial configuration that will be applied to every node. Also enhance the `cordapp` block to support
```
deploy = {true | false}
```
This allows the DSL to declare the following:
```
nodeDefaults {
    projectCordapp {
        deploy = false
    }
}
```

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
@@ -1,14 +1,27 @@
 package net.corda.plugins
 
 import org.gradle.api.Project
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Optional
 import java.io.File
 
-open class Cordapp private constructor(val coordinates: String?, val project: Project?) {
+open class Cordapp private constructor(
+    @get:[Optional Input] val coordinates: String?,
+    @get:[Optional Input] val project: Project?
+) {
     constructor(coordinates: String) : this(coordinates, null)
     constructor(cordappProject: Project) : this(null, cordappProject)
 
     // The configuration text that will be written
+    @Optional
+    @Input
     internal var config: String? = null
+
+    /**
+     * Determines whether or not CordFormation will deploy this CorDapp into Corda.
+     */
+    @Input
+    var deploy: Boolean = true
 
     /**
      * Set the configuration text that will be written to the cordapp's configuration file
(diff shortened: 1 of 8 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
```diff
@@ -1,15 +1,28 @@
 package net.corda.plugins
 
 import org.gradle.api.Project
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Optional
 import java.io.File
 
-open class Cordapp private constructor(val coordinates: String?, val project: Project?) {
+open class Cordapp private constructor(
+    @get:[Optional Input] val coordinates: String?,
```
Clintonio: Wow that's some weird looking syntax.

cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
```diff
@@ -1,15 +1,28 @@
...
     // The configuration text that will be written
+    @Optional
+    @Input
     internal var config: String? = null
 
+    /**
+     * Determines whether or not CordFormation with deploy this CorDapp into Corda.
+     */
+    @Input
+    var deployed: Boolean = true
```
Clintonio: In terms of naming this makes more sense from a DSL perspective, which are declarative, to be "deploy" or "enable" as opposed to "deployed" or "enabled".

cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
chrisr3: But necessary. Otherwise Kotlin will end up annotating the constructor parameters and not the properties! The syntax is telling Kotlin to annotate the property "getters".

cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
chrisr3: I had to changed from `enabled` because it was ambiguous with an existing `enabled` field - probably the task's.

cordformation/src/main/kotlin/net/corda/plugins/Cordapp.kt
Clintonio: Yeah, it makes sense. It's just strange looking. It reminds me of PHP somehow.

PR #110, opened 2018-08-31
Returning node runner to a basic script to remove unnecessary complexity

Clintonio:
This PR is to make the node runner maintainable again. It should never have been more than a branching set of if statements.


Clintonio:
Just realised I failed to add a test for this...

Clintonio:
I'm adding a ticket to restructure the project to add tests.

PR #119, opened 2018-10-02
CORDA-1915 CorDapp signed by default using development key

szymonsztuka:
https://r3-cev.atlassian.net/browse/CORDA-1915,  follow-up of the merged PR https://github.com/corda/corda-gradle-plugins/pull/108:

* Signing is enabled by default for `cordapp` and `cordformation` plugins.
* CorDapp JARs are signed by Corda development key (one as used in Corda node) out-of-box. You can opt-out. Default values can be contextually overwitted by system properties assigned to options in build file e.g.  `keystore System.getProperty('signingKeystore')`, if a system property is not set the option defaults to default value.
* If default Corda dev key is used `cordapp` prints info about this, is up to production user pipeline to ensure their production keyStore is used  or signing is disabled (by e.g. setting system properties).


cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -73,17 +71,41 @@ class CordappPlugin : Plugin<Project> {
         }
         jarTask.dependsOn(task)
 
-        val signTask = project.task("signCordappJar")
+        val optionalSignTask = createSigningTask(project, "singCordapp", false)
```
mikehearn: sing -> sign

You might want to get into the habit of running `git diff | grep sing` until this work is done :)

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -73,17 +71,41 @@ class CordappPlugin : Plugin<Project> {
...
+        jarTask.finalizedBy(optionalSignTask)
+
+        val signTask = project.tasks.findByName("sourceJar") //task in corda-publish-utils plugin
+        if (signTask != null ) {
+            val singTaskForPublishingTask = createSigningTask(project, "singCordappForPublishing", true)
```
mikehearn: sing -> sign

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -73,17 +71,41 @@ class CordappPlugin : Plugin<Project> {
...
+            val jarPublishingTask = signTask as Jar
+            jarPublishingTask.dependsOn(singTaskForPublishingTask)
```
mikehearn: sing -> sign

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -76,15 +79,53 @@ class CordappPlugin : Plugin<Project> {
...
+        val signTask = project.tasks.findByName("sourceJar") //task in corda-publish-utils plugin
+        if (signTask != null ) {
+            val signTaskForPublishingTask = createSigningTask(project, "signCordappForPublishing", true)
+            val jarPublishingTask = signTask as Jar
+            jarPublishingTask.dependsOn(signTaskForPublishingTask)
+        }
+    }
+
+    private fun createSigningTask(project: Project, taskName: String, warnOnDevSetup: Boolean) : Task {
+        val signTask = project.task(taskName)
```
Clintonio: If there's any sensible lines down which this function can be decomposed please do.

szymonsztuka:
@Clintonio slightly shorten code moved to ``private fun sign(project: Project)`` and then added as configureAction 'closure' when creating tasks

Clintonio:
Thanks, LGTM

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -76,15 +79,57 @@ class CordappPlugin : Plugin<Project> {
...
 
-        val signTask = project.task("signCordappJar")
-        signTask.doLast {
-            if (cordapp.signing.enabled) {
-                val options = cordapp.signing.options.toSignJarOptionsMap()
-                options["jar"] = project.tasks.getByName("jar").outputs.files.singleFile.toPath().toString()
+        val signTask = project.task("signCordapp") { it.doLast{ sign(project) } }
+        jarTask.finalizedBy(signTask)
+
+        val sourceJarTask = project.tasks.findByName("sourceJar") //task in corda-publish-utils plugin
```
mikehearn: Hmm, why is this specific to making source JARs? I am not sure where source JARs enter the picture.

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
```diff
@@ -5,28 +5,39 @@ import org.gradle.api.tasks.Input
...
-    @get:Input
-    var alias = "cordapp-signer"
+    companion object {
+        // Defaults to keyStore resource/certificates/cordadevcakeys.jks with Corda developer (non-production) certificates
+        const val DEFAULT_ALIAS = "cordaintermediateca"
+        const val DEFAULT_STOREPASS = "cordacadevpass"
+        const val DEFAULT_KEYSTORE = "cordadevcakeys.jks"
+        const val DEFAULT_STORETYPE = "JKS"
+        const val DEFAULT_KEYPASS = "cordacadevkeypass"
+        // jarSigner doesn't add "SIG-" prefix, and EC signature doesn't follow JAR File Specification
```
mikehearn: Can you elaborate on this? What happens if we use a non-standard signature file name? Do the standard Java APIs and tools still recognise it as a signed jar?

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
Clintonio: Good point. This should be checking for the maven-publish module and it's artifacts. The publish-utils plugin is really only used by R3 and Corda. Cordapp developers will have their own publishing pipeline. 

It'd make more sense to either a) not sign the sources JAR, if that's even an option, or b) search for all JAR tasks in this project and sign them all or c) sign all by default but allow configuration of which tasks get signed

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
szymonsztuka: Signature block file name in JAR file spec can be https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html
```
META-INF/*.DSA
META-INF/*.RSA
META-INF/SIG-*
```

and ` Digital Signatures` paragraph states: 
 ```
 The extension varies depending on the type of digital signature.
.RSA (PKCS7 signature, SHA-256 + RSA)
.DSA (PKCS7 signature, DSA)
Digital signature files for signature algorithms not listed 
above must reside in the META-INF directory and have the prefix "SIG-"
```

However JarSigner (https://docs.oracle.com/javase/8/docs/technotes/tools/windows/jarsigner.html) paragraph:  `The Signed JAR File` can produce (and it does):
```
A signature block file with a .DSA, .RSA, or .EC extension
```

So it produces invalid file name, and Corda internally checks for valid file names otherwise it throws exception (specifically it doesn't recognise that file).

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
szymonsztuka: The problem here is that for jar task run for development use the Corda dev key is ok (out-of-box) but once it's published at least we need to warn that dev key is in use - so I need a way to distinguish the context in which jar task is run. The jar task  itself doesn't know if it's running just as a part of dev build or publishing task (I could presumably walk the task graph and check if publishing is present).  
The simplification of this process would be as you said `c) sign all by default but allow configuration of which tasks get signed`  but if users forgets to add the specific configuration for own keyStore, then it will be silently published with our dev key

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
szymonsztuka: I check tested Java Security Manager (following this tutorial https://docs.oracle.com/javase/tutorial/security/toolsign/index.html)  using signature block file `META-INF/*.EC` and it is accepted. For `jarsigner` itself, if it also verifies as signed JARs with  `*.EC`. It's only the JAR file spec that states otherwise.  If we follow `jarsigner`way it's just matter to add `EC` to regex here https://github.com/corda/corda/blob/f81428eb53dcbc89462c052a94e893bb61e24749/core/src/main/kotlin/net/corda/core/internal/JarSignatureCollector.kt#L15

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
szymonsztuka: Discussed with Clinton, it will be simplified (sign by default but opt-out, info when signing with dev key)  and adding logic to publish-utils from cordapp plugin.

PR #122, opened 2018-10-12
CORDA-2101: Gradle plugins can now be published to the gradle plugins publishing portal

Clintonio:
Caveats

* JAR filter currently does not publish properly due to its shadow plugin, but the main goal of this work is to publish Cordapp and Cordform
* I had to bump the version number to .32 because I published .31 to the portal to test the process (it is identical to .30)
* There is no CI for this process yet and won't be for some time


chrisr3:
So in what sense does JarFilter _not_ publish properly?

jar-filter/build.gradle
```diff
@@ -22,6 +22,21 @@ gradlePlugin {
...
+    tags = ['corda']
+
+    plugins {
+        jarFilterPlugin {
+            displayName = 'Corda JAR Filter Plugin'
+        }
+    }
+}
+
+task('publishPlugin').dependsOn jar
```
chrisr3: This is incorrect. It should actually depend on the `shadowJar` task defined below.

chrisr3: I think you forgot to remove the dummy `publishPlugin` task declaration from `jar-filter` :stuck_out_tongue:.

jar-filter/build.gradle
```diff
@@ -74,8 +89,10 @@ task shadowJar(type: ShadowJar) {
 }
 jar.finalizedBy shadowJar
 
+configurations.archives.artifacts.clear()
 artifacts {
     shadow shadowJar
+    archives shadowJar, sourceJar, javadocJar
```
chrisr3: No need for `javadocJar` here; it's empty :grin:!

PR #130, opened 2018-10-22
CORDA-1915 create signJar task

szymonsztuka:
The task can be uses as to re-sign/sign arbitrary JAR file.
The task reuses Cordapp plugin signing configuration.

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
@@ -69,7 +69,7 @@ class CordappPlugin : Plugin<Project> {
                 attributes["Sealed"] = "true"
             }
         }.doLast {
-            sign(project, cordapp.signing)
+            sign(project, cordapp.signing, project.tasks.getByName("jar").outputs.files.singleFile)
         }
         task.doLast {
             jarTask.from(getDirectNonCordaDependencies(project).map {
@@ -85,9 +85,24 @@ class CordappPlugin : Plugin<Project> {
             }
         }
         jarTask.dependsOn(task)
+
+        val signJarTask = project.task("signJar")
+        signJarTask.outputs.files(signJarTask.inputs.files)
+        signJarTask.doLast {
+            if (it.inputs.files.isEmpty) {
+                throw InvalidUserDataException("No JAR file(s) defined to sign, ensure to configure inputs property for signJar task " +
+                        "by e.g. 'signJar.inputs.files(otherTask.outputs)' or do not invoke singJar task.")
+            }
+            val forceResignOptions = cordapp.signing
+            forceResignOptions.enabled(true)
+            forceResignOptions.options.force = "true" // Just in case - this allows to re-sign already signed JAR file with the same key
+            for (file: File in it.inputs.files) {
+                sign(project, forceResignOptions, file)
+            }
+        }
     }
 
-    private fun sign(project: Project, signing: Signing) {
+    private fun sign(project: Project, signing: Signing, file: File) {
         if (!signing.enabled) {
             project.logger.info("CorDapp JAR signing is disabled, the CorDapp's contracts will not use signature constraints.")
             return
@@ -99,7 +114,7 @@ class CordappPlugin : Plugin<Project> {
             options["keystore"] = keyStorePath.toString()
         }
 
-        val path = project.tasks.getByName("jar").outputs.files.singleFile.toPath()
+        val path = file.toPath()
         options["jar"] = path.toString()
 
         try {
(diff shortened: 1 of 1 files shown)


chrisr3:
Hurrrumph! I've looked at the `CordappExtension`, `Signing`, `SigningOptions` extensions in this plugin and am _sure_ that we can do better. For example, the `CordappExtension` could be rewritten as:
```kotlin
open class CordappExtension(project: Project) {

    /**
     * CorDapp distribution information.
     */
    val info: Info = project.extensions.create("info", Info::class.java)

    /**
     * Optional parameters for ANT signJar tasks to sign Cordapps
     */
    val signing: Signing = project.extensions.create("signing", Signing::class.java, project)
}
```
which instantly makes the `info` and `signing` fields non-null and read-only! The plugin would then create this extension using only:
```
cordapp = project.extensions.create("cordapp", CordappExtension::class.java, project)
configureCordappJar(project)
```
I suspect the original author wasn't aware that Gradle could pass constructor parameters :wink:.

The `Signing` extension could then use a similar approach with `SigningOptions`:
```kotlin
open class Signing(project: Project) {

    // etc etc...

    @get:Input
    val options: SigningOptions = project.extensions.create("options", SigningOptions::class.java)
}
```

cordapp/src/main/kotlin/net/corda/plugins/SignJar.kt
```diff
@@ -0,0 +1,66 @@
...
+        val file = if (!outputs.files.isEmpty) {
+            Files.copy(inputs.files.singleFile.toPath(), outputs.files.singleFile.toPath(), StandardCopyOption.REPLACE_EXISTING)
+            outputs.files.singleFile
+        } else {
+            inputs.files.singleFile
+        }
+        val signing = extension as Signing
+        signing.enabled(true)
+        sign(project, signing, file)
+    }
```
chrisr3: And then you should be able to simplify this, e.g. you will _know_ that it won't be invoked with empty inputs.
BTW, I don't think you should force `cordapp.signing.enabled = true` here; that's a global setting which the user is expected to have set correctly already. If you want to override it here then the task should have a private mechanism of its own.

chrisr3:
Going over this again, I think the `Signing` extension should declare `options` as `@get:Nested` rather than `@get:Input`:
```kotlin
/** JAR sign and control option. */
open class Signing(project: Project) {

    // etc etc

    @get:Nested
    val options: SigningOptions = project.extensions.create("options", SigningOptions::class.java)
}
```
This point is that `options` **has** `@Input` fields rather than being an `@Input` in its own right.

By the same logic, the `info` and `signing` fields on `CordappExtension` should also both be annotated as `@get:Nested`. And thefields on `Info` are all `@get:Input`.

chrisr3:
OK, I've refitted my Evil Gradle Helmet and come up with this:
```kotlin
open class CordappExtension @Inject constructor(objectFactory: ObjectFactory) {

    /**
     * CorDapp distribution information.
     */
    @get:Nested
    val info: Info = objectFactory.newInstance(Info::class.java)

    /**
     * Optional parameters for ANT signJar tasks to sign Cordapps
     */
    @get:Nested
    val signing: Signing = objectFactory.newInstance(Signing::class.java)

    fun info(action: Action<Info>) {
        action.execute(info)
    }

    fun signing(action: Action<Signing>) {
        action.execute(signing)
    }
}
```
which `CordappPlugin` would apply as:
```kotlin
cordapp = project.extensions.create("cordapp", CordappExtension::class.java, project.objects)
configureCordappJar(project)
```
So not too different :relieved:. And then we'd need something similar for `Signing`:
```kotlin
open class Signing @Inject constructor(objectFactory: ObjectFactory) {

    @get:Input
    var enabled: Boolean = true
        private set

    fun enabled(value: Boolean) {
        enabled = value
    }

    fun enabled(value: String?) {
        enabled = value?.toBoolean() ?: true
    }

    @get:Nested
    val options: SigningOptions = objectFactory.newInstance(SigningOptions::class.java)

    fun options(action: Action<SigningOptions>) {
        action.execute(options)
    }
}
```
The `SignJar` plugin would use the same approach too:
```kotlin
open class SignJar @Inject constructor(objectFactory: ObjectFactory) : DefaultTask() {
   // etc etc

    @Nested
    val signing: Signing = objectFactory.newInstance(Signing::class.java)

    fun signing(action: Action<Signing>) {
        action.execute(signing)
    }

    // etc etc
}
```

chrisr3:
I think we were abusing extensions before, when all we really wanted was to extend the DSL using nested objects.

cordapp/src/main/kotlin/net/corda/plugins/SignJar.kt
```diff
@@ -0,0 +1,94 @@
...
+                if (signing.options.hasDefaultOptions()) {
+                    Paths.get(options["keystore"]).toFile().delete()
+                }
+            }
+        }
+    }
+
+    private val signing: Signing = (project.extensions.findByName("cordapp") as CordappExtension).signing
+
+    private var _postfix = "-signed"
```
chrisr3: I think `postfix` can be rewritten as:
```kotlin
@get:Input
var postfix: String = "-signed"

fun postfix(value: String) {
    postfix = value
}
```

szymonsztuka: Indeed, even without method `postfix` will support both declaration with assignment `=` and without.

PR #131, opened 2018-10-25
Remove Groovy Closure from JarFilter DSL and simplify Kotlin/Gradle usage.

chrisr3:
There is no functional change here; I have only tidied up some documentation and Kotlin/Gradle usage.
- Remove the Groovy `Closure` from JarFilter's DSL and replace it with a Kotlin `Action`.
- Fix regular expression for jars.
- Use fewer lambdas.
- Update the README now that `JarFilter` has been migrated out of Corda's `buildSrc`.

### DIFF
--- a/jar-filter/src/test/resources/abstract-function/build.gradle
+++ b/jar-filter/src/test/resources/abstract-function/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'org.jetbrains.kotlin.jvm' version '$kotlin_version'
-    id 'net.corda.plugins.jar-filter'
+    id 'net.corda.plugins.jar-filter' apply false
 }
 apply from: 'repositories.gradle'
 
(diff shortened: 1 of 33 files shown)


jar-filter/src/test/resources/abstract-function/build.gradle
```diff
@@ -1,6 +1,6 @@
 plugins {
     id 'org.jetbrains.kotlin.jvm' version '$kotlin_version'
-    id 'net.corda.plugins.jar-filter'
+    id 'net.corda.plugins.jar-filter' apply false
```
szymonsztuka: Is not applied, because the plugin is needed to allow it's task to be imported only  e.g. `import net.corda.gradle.jarfilter.JarFilterTask` ?

chrisr3: The plugin only needs to be on the plugin classpath so that the `Task` classes are visible. The plugin itself is implemented as:
```kotlin
class JarFilterPlugin : Plugin<Project> {
    override fun apply(project: Project) {
        project.logger.info("Applying JarFilter plugin")
    }
}
```
because it has nothing else to do :grin:.

szymonsztuka: Indeed as explained in README.md.

PR #134, opened 2018-11-01
Few signing options defaults to system properties first, then to hardcoded defaults (improvements for CORDA-1915).

szymonsztuka:
Functional change: 
for few main `signing.options` the plugin tries to get a value from the relevant system property if the option is not set in the configuration entry, the system property name is 'signing.' + option name.
Refactorings:
String refactored to constants as some of them are used multiple times.
Removing redundant methods named as relevant property, for simple types both conventions works out-of-box (`key "value"` and `key="value"`).
Added methods with Boolean parameter for options which are true|false to enable (`key value` and not only `key "value"`).
Replace one 'instance' of Groovy `Closure` with Gradle `Action`.

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/Signing.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/Signing.kt
@@ -10,14 +10,14 @@ import javax.inject.Inject
 open class Signing @Inject constructor(objectFactory: ObjectFactory) {
 
     @get:Input
-    var enabled: Boolean = true
+    var enabled: Boolean = System.getProperty("signingEnabled", "true").toBoolean()
 
     fun enabled(value: Boolean) {
         enabled = value
     }
 
-    fun enabled(value: String?) {
-        enabled = value?.toBoolean() ?: true
+    fun enabled(value: String) {
+        enabled = value.toBoolean()
     }
 
     @get:Nested
--- a/cordapp/src/main/kotlin/net/corda/plugins/Signing.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/Signing.kt
@@ -1,16 +1,16 @@
 package net.corda.plugins
 
+import net.corda.plugins.SigningOptions.Companion.SYSTEM_PROPERTY_PREFIX
 import org.gradle.api.Action
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Nested
 import javax.inject.Inject
 
-/** JAR sign and control option. */
 open class Signing @Inject constructor(objectFactory: ObjectFactory) {
 
     @get:Input
-    var enabled: Boolean = System.getProperty("signingEnabled", "true").toBoolean()
+    var enabled: Boolean = System.getProperty(SYSTEM_PROPERTY_PREFIX + "enabled", "true").toBoolean()
 
     fun enabled(value: Boolean) {
         enabled = value
--- a/cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
@@ -2,11 +2,10 @@ package net.corda.plugins
 
 import org.gradle.api.tasks.Input
 
-/** JAR sign options. */
-//Option used for ANT task "signjar"
+/** Options for ANT task "signjar". */
 open class SigningOptions {
     companion object {
-        // Defaults to resource/certificates/cordadevcakeys.jks keyStore with Corda development key
+        // Defaults to resource/certificates/cordadevcakeys.jks keystore with Corda development key
         private const val DEFAULT_ALIAS = "cordaintermediateca"
         private const val DEFAULT_STOREPASS = "cordacadevpass"
         private const val DEFAULT_STORETYPE = "JKS"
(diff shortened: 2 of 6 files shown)


cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
```diff
@@ -2,207 +2,160 @@ package net.corda.plugins
...
+        }
     }
 
     @get:Input
-    var storepass = ""
-
-    fun storepass(value: String?) {
-        storepass = value ?: ""
-    }
+    var alias = System.getProperty(SYSTEM_PROPERTY_PREFIX + Key.ALIAS, DEFAULT_ALIAS)
```
Clintonio: The fact these will resolve to null isn't a problem?

szymonsztuka: It' ok, if not specified in plugin config it will try system property and then the hardcoded `DEFAULT_ALIAS`, or did you mean something else?

Clintonio: Aha, I missed the second parameter to getProperty.

cordapp/src/main/kotlin/net/corda/plugins/Signing.kt
```diff
@@ -1,23 +1,23 @@
...
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Nested
 import javax.inject.Inject
 
-/** JAR sign and control option. */
 open class Signing @Inject constructor(objectFactory: ObjectFactory) {
 
     @get:Input
-    var enabled: Boolean = true
+    var enabled: Boolean = System.getProperty(SYSTEM_PROPERTY_PREFIX + "enabled", "true").toBoolean()
```
chrisr3: There's a little-known Java API called [Boolean.getBoolean(String)](https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#getBoolean(java.lang.String)) ... :wink:

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
```diff
@@ -2,207 +2,160 @@ package net.corda.plugins
...
-    // instead of requiring 'alias System.getProperty('key','')'
-    @get:Input
-    var alias = ""
-
-    fun alias(value: String?) {
-        alias = value ?: ""
+    /** Option keys for ANT task. */
+    class Key {
+        companion object {
+            val JAR = "jar"
```
chrisr3: These can all be `const val`

cordformation/src/main/kotlin/net/corda/plugins/KeyGenAndSigning.kt
```diff
@@ -1,30 +1,49 @@
...
+    fun generateKeystore(value: Boolean) {
+        generateKeystore = value
+    }
+
+    fun generateKeystore(value: String) {
+        generateKeystore = value.toBoolean()
+    }
+
+    @get:Nested
+    var options: KeyGenAndSigningOptions = objectFactory.newInstance(KeyGenAndSigningOptions::class.java)
```
chrisr3: Isn't this a `val` rather than `var`?

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
```diff
@@ -2,207 +2,160 @@ package net.corda.plugins
...
-
     @get:Input
     var signedjar = ""
 
-    fun signedjar(value: String?) {
-        signedjar = value ?: ""
-    }
-
     @get:Input
     var verbose = ""
```
chrisr3: I don't think `verbose` is an `@Input`. Probably more likely it should be `@Console`.

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
```diff
@@ -2,207 +2,160 @@ package net.corda.plugins
...
     var lazy = ""
 
-    fun lazy(value: String?) {
-        lazy = value ?: ""
+    fun lazy(value: Boolean) {
+        lazy = value.toString()
     }
 
     @get:Input
     var maxmemory = ""
```
chrisr3: Would changing this setting affect the hash of the output artifact? If it wouldn't then the property is probably `@Internal`. (Not sure _whose_ memory is having its maximum set here, to be fair.)

cordformation/src/main/kotlin/net/corda/plugins/KeyGenAndSigningOptions.kt
```diff
@@ -2,28 +2,34 @@ package net.corda.plugins
...
 
-/** JAR sign and keystore generation options, with default values set for minimal required set. */
-//Option used for ANT tasks "genkey" and "signjar"
-class KeyGenAndSigningOptions : SigningOptions() {
+/** Options for ANT tasks "genkey" and "signjar". */
+open class KeyGenAndSigningOptions : SigningOptions() {
+    /** Additional option keys for ANT task. */
+    class Key {
+        companion object {
+            val KEYALG = "keyalg"
```
chrisr3: `const val` again

cordapp/src/main/kotlin/net/corda/plugins/Signing.kt
szymonsztuka: Indeed, however in this case I want to default to `true` if property is missing.

cordapp/src/main/kotlin/net/corda/plugins/SigningOptions.kt
szymonsztuka: yes, it's memory to run `jarsigner`

szymonsztuka:
@chrisr3 thanks, all fixed

cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -66,14 +67,13 @@ open class Baseform : DefaultTask() {
     var nodeDefaults: Closure<in Node>? = null
 
     /**
-     * Options for sign Cordapp JARs and generate keyStore for it,
-     * by default all Cordapps are signed with default keyStore options.
+     * Configuration for keystore generation and JAR signing.
      */
     @Input
-    var signing: KeyGenAndSigning = KeyGenAndSigning()
+    var signing: KeyGenAndSigning = project.objects.newInstance(KeyGenAndSigning::class.java)
```
chrisr3: Oops, one more `val` :stuck_out_tongue_winking_eye:

szymonsztuka: fixed

PR #146, opened 2018-12-04
CORDA-2148 New Cordapp plugin versioning naming conventions 

josecoll:
The following CorDapp jar manifest attributes:

```
["Name"]
["Implementation-Version"]
["Implementation-Vendor"]
```
are **enhanced** with distinct identifiers for Contract and WorkFlow jars:

```
["Cordapp-Contract-Name"]
["Cordapp-Contract-Version"]
["Cordapp-Contract-Vendor"]
["Cordapp-Workflow-Name"]
["Cordapp-Workflow-Version"]
["Cordapp-Workflow-Vendor"]
```

and supplemented with:

```
["Cordapp-Contract-Licence"]
["Cordapp-Workflow-Licence"]
```

An associated PR in Corda will align these new attributes for usage in Corda 4.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext {
-        gradle_plugins_version = '4.0.37-SNAPSHOT'
+        gradle_plugins_version = '4.0.37'
         bouncycastle_version = '1.57'
         typesafe_config_version = '1.3.1'
         jsr305_version = '3.0.2'
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext {
-        gradle_plugins_version = '4.0.37'
+        gradle_plugins_version = '4.0.37-SNAPSHOT'
         bouncycastle_version = '1.57'
         typesafe_config_version = '1.3.1'
         jsr305_version = '3.0.2'
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
@@ -68,6 +68,8 @@ class CordappPlugin : Plugin<Project> {
                 skip = true
             }
             // Deprecated support (Corda 3)
+            if (skip && cordapp.info.name != null)
+                project.logger.warn("Ignoring deprecated 'info' attributes. Using 'contract' and 'workflow' attributes.")
             if (!skip && cordapp.info.name != null) {
                 attributes["Name"] = cordapp.info.name ?: "${project.group}.${jarTask.baseName}"
                 attributes["Implementation-Version"] = cordapp.info.version ?: project.version
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
@@ -53,13 +53,13 @@ class CordappPlugin : Plugin<Project> {
             val attributes = jarTask.manifest.attributes
             if (cordapp.contract.name != null) {
                 attributes["Cordapp-Contract-Name"] = cordapp.contract.name ?: "${project.group}.${jarTask.baseName}"
-                attributes["Cordapp-Contract-Version"] = parseVersion(cordapp.contract.versionId)
+                attributes["Cordapp-Contract-Version"] = parseVersion(cordapp.contract.versionId.toString())
                 attributes["Cordapp-Contract-Vendor"] = cordapp.contract.vendor ?: UNKNOWN
                 attributes["Cordapp-Contract-Licence"] = cordapp.contract.licence ?: UNKNOWN
             }
             if (cordapp.workflow.name != null) {
                 attributes["Cordapp-Workflow-Name"] = cordapp.workflow.name ?: "${project.group}.${jarTask.baseName}"
-                attributes["Cordapp-Workflow-Version"] = parseVersion(cordapp.workflow.versionId)
+                attributes["Cordapp-Workflow-Version"] = parseVersion(cordapp.workflow.versionId.toString())
                 attributes["Cordapp-Workflow-Vendor"] = cordapp.workflow.vendor ?: UNKNOWN
                 attributes["Cordapp-Workflow-Licence"] = cordapp.workflow.licence ?: UNKNOWN
             }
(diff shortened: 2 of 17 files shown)


cordapp/src/main/kotlin/net/corda/plugins/Info.kt
```diff
@@ -4,11 +4,21 @@ import org.gradle.api.tasks.Input
 
 open class Info {
     @get:Input
-    var name: String? = null
+    var cordappContractName: String? = null
     @get:Input
-    var version: String? = null
+    var cordappContractVersion: String? = null
```
szymonsztuka: Maybe `Int` ?

build.gradle
```diff
@@ -1,6 +1,6 @@
 buildscript {
     ext {
-        gradle_plugins_version = '4.0.37-SNAPSHOT'
+        gradle_plugins_version = '4.0.37'
```
szymonsztuka: This should go as separate PR, after the release.

cordapp/src/main/kotlin/net/corda/plugins/Info.kt
josecoll: @szymonsztuka lets keep it as a String in the gradle plugins to cater for future evolvability.
But we will treat it as an integer for the purposes of Contract version downgrade checking.

chrisr3:
:man_dancing: LGTM :man_dancing:

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -130,6 +141,16 @@ class CordappPlugin : Plugin<Project> {
         return Pair(targetPlatformVersion, minimumPlatformVersion)
     }
 
+    private fun parseVersion(versionStr: String?, project: Project): Int {
+        if (versionStr == null) return 1
+        return try {
+            Integer.parseInt(versionStr)
+        } catch (e: NumberFormatException) {
+            project.logger.info("Invalid version identifier: $versionStr. Defaulting to 1")
```
shamsasari: It's most likely they meant something other than 1. Better to throw an exception explaining `versionStr` is not an _integer_ version.

chrisr3: i.e. Gradle's `InvalidUserDataException`.

cordapp/src/test/resources/net/corda/plugins/CorDappWithContractInfo.gradle
```diff
@@ -0,0 +1,25 @@
+plugins {
+    id 'java'
+    id 'net.corda.plugins.cordapp'
+}
+
+repositories {
```
chrisr3: Perhaps refactor some of the repeated elements from these test scripts into common files? E.g. you could create a `repositories.gradle` file in `test/resources` containing:
```
repositories {
    mavenCentral()
    maven { url 'https://blah.blah.blah' }
}
```
Then replaces each `repositories` block with
```
apply from: 'repositories.gradle`
```
Finally, you'd need just need to copy this `repositories.gradle` file into the `testProjectDir`. You could do this in the `CordappTest`'s `@Before` function.

The point is that we can then update these repositories in just one place rather than `N`.

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
josecoll: Addressed.

cordapp/src/test/resources/net/corda/plugins/CorDappWithContractInfo.gradle
josecoll: Done!

cordapp/src/main/kotlin/net/corda/plugins/Workflow.kt
```diff
@@ -0,0 +1,15 @@
+package net.corda.plugins
+
+import org.gradle.api.tasks.Input
+
+open class Workflow {
+    @get:Input
```
chrisr3: Are any of these parameters `@Optional`?

cordapp/src/test/kotlin/net/corda/plugins/Utilities.kt
```diff
@@ -0,0 +1,25 @@
+import org.junit.rules.TemporaryFolder
```
chrisr3: Wot no `package`?

cordapp/src/test/resources/net/corda/plugins/CorDappWithInfoAll.gradle
```diff
@@ -0,0 +1,34 @@
...
+apply from: 'repositories.gradle'
+
+jar {
+    baseName 'test-cordapp'
+}
+
+cordapp {
+    targetPlatformVersion target_version_arg.toInteger()
+    minimumPlatformVersion min_platform_version_arg.toInteger()
+    info {
```
shamsasari: Surely specifying `info` alongside `contract` or `workflow` should not be allowed. Also if you specify the top level `targetPlatformVersion` or `minimumPlatformVersion` with `info`. You have to have extra checks to make sure the values are consistent.

cordapp/src/test/kotlin/net/corda/plugins/Utilities.kt
josecoll: bizarre how that got through.

cordapp/src/main/kotlin/net/corda/plugins/Workflow.kt
josecoll: There are all optional in as far as the CorDapp developer can specify either a `contract`, `workflow` or `info` `cordapp` section in their respective build.gradle files.

chrisr3: Then _technically speaking_ they should also be annotated as `@Optional`. Although this probably doesn't matter unless a Gradle task uses this object as a `@Nested` field.

cordapp/src/test/resources/net/corda/plugins/CorDappWithInfoAll.gradle
josecoll: Fixed.
New top-level, `contract` and `workflow` attributes will always take precedence over older (deprecated `info` object items)

chrisr3:
:man_cartwheeling: LGTM :man_cartwheeling:

PR #150, opened 2018-12-28
CORDA-2349 - Include synthetic methods in the api-scanning

dimosr:
## Changes
Including the synthetic methods in those generated by the `api-scanner` plugin, which are subsequently checked as part of the API backwards-compatibility checks. This is because for some cases (i.e. `@ JvmOverloads `), invocation of a method is different depending on whether it's done from Java or Kotlin. Kotlin generates a separate synthetic method that is used, instead of the ones that are generated for Java. As a result, these synthetic methods also need to be included in the API checks, otherwise we might be breaking clients that are consuming the API in Kotlin (note: the change might be in theory backwards-compatible, i.e. adding an optional variable in the end, it's Kotlin's synthesis process that makes it non compatible).

## Testing
* Update the unit tests & run `./gradlew api-scanner:test`
* Switched the corda project to use the snapshot for the `api-scanner` plugin and run:
    * `./gradlew generateApi`
    * `.ci/check-api-changes.sh`
    * there are 171 additional methods (see attached file [api-changes-4.0_4.0-with-synthetic.txt](https://github.com/corda/corda-gradle-plugins/files/2714739/api-changes-4.0_4.0-with-synthetic.txt))
* Also, performed a run for the release 3.3 and the snapshot of version 4 of `corda` with the new version of `api-scanner`. The results of it are a bit more involved, so I added some more info in the JIRA ticket.

## Rollout plan
- Merge changes in `corda-gradle-plugins` in the current version `4.0.38-SNAPSHOT`
- Release `4.0.38` of `corda-gradle-plugins` and bump to `4.0.39-SNAPSHOT`
- Upgrade `corda` to use `4.0.38` of corda-gradle-plugins (at least for the `api-scanner` module) and update the `api-current.txt` file with the additional synthetic methods that are introduced.


dimosr:
Thanks for the catch, I opened a separate PR on the `corda` repo with the change to exclude `kotlin.jvm.internal`: https://github.com/corda/corda/pull/4478

There are only 2 additions now (from the `net.corda.node` package).
The updated results are: [api-changes-4.0_4.0-with-synthetic-v2.txt](https://github.com/corda/corda-gradle-plugins/files/2718174/api-changes-4.0_4.0-with-synthetic-v2.txt)

chrisr3:
These two remaining cases are presumably flagged because they expose `net.corda.node.`, but isn't this still a problem? I.e. this will still break the API Stability build with the new `api-scanner` plugin.

dimosr:
1. Yes, that's correct. Given that this is not the classical case of constructors with added default parameters (where there's the straightforward workaround), but a method that exposes internal classes, I am not sure what should be the approach to resolve it. I'll have a chat with @sollecitom and @shamsasari, who were looking on stuff around this space.

For reference, looked on the file and detected the exact components, which are the following:
```
public static final class net.corda.testing.node.InMemoryMessagingNetwork$MessageTransfer extends java.lang.Object
  public <init>(net.corda.testing.node.InMemoryMessagingNetwork$PeerHandle, net.corda.node.services.messaging.Message, net.corda.core.messaging.MessageRecipients, kotlin.jvm.internal.DefaultConstructorMarker)

public class net.corda.testing.node.MockServices extends java.lang.Object implements net.corda.core.node.ServiceHub
	public class net.corda.testing.node.MockServices extends java.lang.Object implements net.corda.core.node.ServiceHub
  public <init>(net.corda.node.cordapp.CordappLoader, net.corda.core.node.services.IdentityService, net.corda.core.node.NetworkParameters, net.corda.testing.core.TestIdentity, java.security.KeyPair[], kotlin.jvm.internal.DefaultConstructorMarker)
```

2. Yes, any of the methods that are added in the `.txt` file do not contain `@CordaInternal`. The filtering is done in `ScanApi#writeMethods()`, see: https://github.com/corda/corda-gradle-plugins/pull/150/files#diff-c44603596d65bcd19d3dc3d123df8d7fR310

api-scanner/src/test/java/net/corda/plugins/KotlinAnnotationsTest.java
```diff
@@ -43,6 +43,7 @@
 
     private static final String[] expectedClassWithJvmStaticFunctionCompanion = {
         "public static final class net.corda.example.HasJvmStaticFunction$Companion extends java.lang.Object",
+        "  public <init>(kotlin.jvm.internal.DefaultConstructorMarker)",
```
chrisr3: This isn't a valid test-case - Kotlin should also be adding an `int` for the bitmask.
```java
public <init>(int, kotlin.jvm.internal.DefaultConstructorMarker)",
```

chrisr3:
If we're going to use `@CordaInternal` to remove those two final constructors then we should probably test that API Scanner can ignore constructors that are annotated with `@CordaInternal` too :wink:.

It should be easy to create an `internal-constructor` test-case by examining `internal-method` and then adding `CONSTRUCTOR` to the `@Target` in the `common-internal` resource.

chrisr3:
*Sigh* - I have _no idea_ why Kotlin is generating these two public synthetic constructors for `MockServices` and `InMemoryMessagingNetwork$MessageTransfer`.

I also have this horrible suspicion that Kotlin has made them both `public` for a _reason_, and that we risk breaking our ABI again unless we understand what that reason is :man_facepalming:.

chrisr3:
I have just remembered that the API Scanner plugin implements an `excludeClasses` block, e.g. in the `core` module:
```gradle
scanApi {
    excludeClasses = [
        // Kotlin should probably have declared this class as "synthetic".
        "net.corda.core.Utils\$toFuture\$1\$subscription\$1"
    ]
}
```
Perhaps we could also implement an `excludeMethods = []` clause here for those two pesky synthetic constructors?

dimosr:
I've added am option to exclude methods for specific classes, so that we can exclude these methods.
I noticed they are both in `testing`, which implies that they will be used in context, where the client code will be re-compiled against the new version, so I believe the impact of whitelisting them should be minimal.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -434,6 +450,23 @@ private static boolean isValid(int modifiers, int mask) {
         return (modifiers & mask) == modifiers;
     }
 
+    private boolean isExcluded(ApiPrintWriter writer, MethodInfo methodWithoutAnnotations) {
+        final StringWriter stringWriter = new StringWriter();
```
chrisr3: You have formatted the method signature nicely for use in `build.gradle`, but would this be more stable if we used the method's descriptor?
```java
String className = methodWithoutAnnotations.getClassName();
String methodDescriptor = methodWithoutAnnotations.getTypeDescriptorStr();
return this.excludeMethods.containsKey(className) &&
                this.excludeMethods.get(className).contains(methodDescriptor);
```
The method descriptor is the stable part of the ABI that is written into the bytecode, whereas our formatting of it may change.

api-scanner/src/test/java/net/corda/plugins/KotlinAnnotationsTest.java
```diff
@@ -95,4 +97,10 @@ public void testJvmDefaultAnnotation() throws IOException {
                  "##"
             );
     }
+
+    @Test
+    public void testFilteredMethodsAreExcluded() throws IOException {
+        assertThat(testProject.getApiLines())
+                .doesNotContain("  public <init>(int, String, int, kotlin.jvm.internal.DefaultConstructorMarker)");
```
chrisr3: You can use something like `doesNotContainSequence("...")` to make this less brittle. I am concerned that we are sensitive to things like indentation and whitespace here.

api-scanner/src/test/resources/kotlin-annotations/kotlin/net/corda/example/ClassWithFilteredMethods.kt
```diff
@@ -0,0 +1,5 @@
+@file:Suppress("UNUSED")
+package net.corda.example
+
+
+class ClassWithFilteredMethods @JvmOverloads constructor(val number: Int = 1, val text: String = "yummy")
```
chrisr3: Technically speaking, this isn't a case where we would like to exclude the constructor from the API Scanner.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -68,13 +71,15 @@
...
     private final Set<String> excludeClasses;
+    private final Map<String, Collection<String>> excludeMethods;
     private final File outputDir;
     private boolean verbose;
 
     public ScanApi() {
         sources = getProject().files();
         classpath = getProject().files();
         excludeClasses = new LinkedHashSet<>();
+        excludeMethods = new HashMap<>();
```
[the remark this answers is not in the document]
chrisr3: I agree that this shouldn't _really_ make a difference here, but "consisency _is_ good" :grinning:

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -434,6 +450,23 @@ private static boolean isValid(int modifiers, int mask) {
...
+        final ApiPrintWriter apiPrintWriter = ApiPrintWriter.withStringWriter(stringWriter);
+        apiPrintWriter.println(methodWithoutAnnotations, "");
+        final String methodSignature = removeTrailingSpaceAndNewLines(stringWriter.toString());
+        final String className = methodWithoutAnnotations.getClassName();
+
+        return this.excludeMethods.containsKey(className) &&
+                this.excludeMethods.get(className).contains(methodSignature);
+    }
+
+    private String removeTrailingSpaceAndNewLines(final String value) {
```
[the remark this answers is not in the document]
chrisr3: _Dislike regular expressions!?!?!??!_ :hear_no_evil:

They're also slow to compile, and so by invoking `replace(target, value)` twice you pay that penalty twice too :grinning:.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
```diff
@@ -108,6 +113,16 @@ void setExcludeClasses(Collection<String> excludeClasses) {
         this.excludeClasses.addAll(excludeClasses);
     }
 
+    @Input
+    public Map<String, Collection<String>> getExcludeMethods() {
+        return unmodifiableMap(excludeMethods);
+    }
+
+    void setExcludeMethods(Map<String, Collection<String>> excludeMethods) {
```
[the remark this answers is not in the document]
chrisr3: It's worth pointing out here that this `Collection<String>` is being passed to us from Groovy / Gradle, so it's safer to convert it to what the API Scanner needs it to be anyway - i.e. `Set<String` in this case.

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
chrisr3: For reference,
```java
<init>(int, String, int, kotlin.jvm.internal.DefaultConstructorMarker)
```
has a descriptor of
```java
(ILjava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```

api-scanner/src/test/resources/kotlin-exclude-method/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+}
+
+jar {
+    baseName = "kotlin-exclude-method"
+}
+
+scanApi {
+    verbose = true
+    excludeMethods = [
+            "net.corda.example.ClassWithExtraConstructorGenerated": ["(Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/internal/DefaultConstructorMarker;)V"]
```
dimosr: ugh

api-scanner/src/main/java/net/corda/plugins/ScanApi.java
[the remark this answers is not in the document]
dimosr: http://a.yu8.us/xkcd-208-regular_expressions.png

api-scanner/src/test/java/net/corda/plugins/KotlinExcludeMethodTest.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final TemporaryFolder testProjectDir = new TemporaryFolder();
+    private static final GradleProject testProject = new GradleProject(testProjectDir, "kotlin-exclude-method");
+
+    @ClassRule
+    public static final TestRule rules = RuleChain.outerRule(testProjectDir).around(testProject);
+
+    @Test
+    public void testFilteredMethodsAreExcluded() throws IOException {
+        assertThat(testProject.getApiText())
+                .doesNotContain("<init>(String, String, kotlin.jvm.internal.DefaultConstructorMarker)");
```
chrisr3: This test would also pass if the API text were completely empty, though :cry:. So perhaps also test that
```java
.contains("<init>(kotlin.jvm.internal.DefaultConstructorMarker)");
```
You can chain these together so that `getApiText()` is still only called once.

api-scanner/src/test/resources/kotlin-exclude-method/kotlin/net/corda/example/ClassWithExtraConstructorGenerated.kt
```diff
@@ -0,0 +1,10 @@
+@file:Suppress("UNUSED")
+package net.corda.example
+
+class ClassWithExtraConstructorGenerated private constructor(val val1: String, val val2: String) {
```
chrisr3: Declaring `val1` and `val2` as `private` would remove them from the API output and so reduce the size of the test case.

api-scanner/src/test/java/net/corda/plugins/KotlinAnnotationsTest.java
chrisr3: *Sigh*, _of course_ "sequence" doesn't mean "substring" in this context :man_facepalming:!

api-scanner/src/test/java/net/corda/plugins/KotlinExcludeMethodTest.java
dimosr: good idea, I'll add the class name.

PR #153, opened 2019-01-07
CORDA-2383 - Relax CorDapp info requirements in Gradle plugin

josecoll:
Relax cordapp metadata checking (warn on missing metadata rather than failing-fast).

This allows multi-module structured CorDapps to only define metadata for flows and/or contracts modules.


cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -52,34 +52,40 @@ class CordappPlugin : Plugin<Project> {
...
+            else {
+                // Corda 4 attributes support
+                if (cordapp.contract.name != null) {
+                    attributes["Cordapp-Contract-Name"] = cordapp.contract.name ?: "${project.group}.${jarTask.baseName}"
+                    attributes["Cordapp-Contract-Version"] = parseVersion(cordapp.contract.versionId.toString())
+                    attributes["Cordapp-Contract-Vendor"] = cordapp.contract.vendor ?: UNKNOWN
+                    attributes["Cordapp-Contract-Licence"] = cordapp.contract.licence ?: UNKNOWN
+                    skip = true
+                }
+                if (cordapp.workflow.name != null) {
```
chrisr3: And also:
```kotlin
if (!cordapp.workflow.isEmpty()) {
   ...
}
```

cordapp/src/main/kotlin/net/corda/plugins/Contract.kt
```diff
@@ -12,4 +12,5 @@ open class Contract {
     var vendor: String? = null
     @get:Input
     var licence: String? = null
+    internal fun isEmpty(): Boolean = (name == null && versionId == null && vendor == null && licence == null)
```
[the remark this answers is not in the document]
josecoll: Addressed

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -52,34 +52,42 @@ class CordappPlugin : Plugin<Project> {
...
-            val (targetPlatformVersion, minimumPlatformVersion) = checkPlatformVersionInfo()
-            attributes["Target-Platform-Version"] = targetPlatformVersion
-            attributes["Min-Platform-Version"] = minimumPlatformVersion
-            if (cordapp.sealing.enabled) {
-                attributes["Sealed"] = "true"
+            else {
+                // Corda 4 attributes support
+                if (!cordapp.contract.isEmpty()) {
+                    attributes["Cordapp-Contract-Name"] = cordapp.contract.name ?: "${project.group}.${jarTask.baseName}"
+                    attributes["Cordapp-Contract-Version"] = parseVersion(cordapp.contract.versionId.toString())
```
chrisr3:
```kotlin
cordapp.contract.versionId.toString()
```
This must be invoking `fun Any?.toString(): String`, which means that it will return either the string representation of `versionId` or a string value of `"null"`. Is this what we want?

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -52,34 +52,42 @@ class CordappPlugin : Plugin<Project> {
...
+                    attributes["Cordapp-Contract-Vendor"] = cordapp.contract.vendor ?: UNKNOWN
+                    attributes["Cordapp-Contract-Licence"] = cordapp.contract.licence ?: UNKNOWN
+                    skip = true
+                }
+                if (!cordapp.workflow.isEmpty()) {
+                    attributes["Cordapp-Workflow-Name"] = cordapp.workflow.name ?: "${project.group}.${jarTask.baseName}"
+                    attributes["Cordapp-Workflow-Version"] = parseVersion(cordapp.workflow.versionId.toString())
```
chrisr3:
```kotlin
cordapp.workflow.versionId.toString()
```
Same comment as above.

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
josecoll: Addressed

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -156,8 +156,8 @@ class CordappPlugin : Plugin<Project> {
         return Pair(targetPlatformVersion, minimumPlatformVersion)
     }
 
-    private fun parseVersion(versionStr: String?): Int {
-        if (versionStr == null)
+    private fun parseVersion(versionStr: Any?): Int {
+        if (versionStr == null || versionStr !is String)
```
chrisr3: _**AAAGH!** I'm so confused!_ :angry: 

This `versionId` is _documented_ as being `Any` type because
> relaxed type so users can specify Integer or String identifiers

which implies that we would also like to allow something like `versionId = '1-Potato'`.

So I'd have thought that we'd want something more like:
```kotlin
attributes["Cordapp-Contract-Version"] = cordapp.contract.versionId.let { v -> v as? CharSequence ?: parseVersion(v) }
```
and then
```kotlin
private fun parseVersion(versionStr: Any?): Int {
    if (versionStr == null)
        throw InvalidUserDataException("Target versionId not specified. Please specify a whole number starting from 1.")
    return try {
        val version = Integer.parseInt(versionStr.toString())
        ...
```
Although these error messages aren't consistent with that philosophy either :confused:.

And *YES*, we do need `as? CharSequence` instead of `as? String` because someone might pass Gradle a Groovy `GString` here by using double-quotes :man_facepalming:. E.g.
```gradle
cordapp {
    contract {
        versionId "$cordappContractVersion_info_arg"
```

Some extra test cases would be nice...

chrisr3: If `versionId` were a `String` then Groovy would convert the `GString` to `String` automatically. But we're using `Any` so that someone can also use an `Int`.

josecoll: yes - it was meant to cater for Int or String specifications (and nothing else).
And these changes are not even related to this PR so I'm going to rollback the last commit and leave it there.

chrisr3: OK, that new JIRA will be with you momentarily... :imp:

chrisr3:
Approved, subject to the "knock-on" PR.

PR #155, opened 2019-01-10
exclude all funky dependencies from the produced pom

roastario:


build.gradle
```diff
@@ -1,6 +1,6 @@
 buildscript {
     ext {
-        gradle_plugins_version = '4.0.38-SNAPSHOT'
+        gradle_plugins_version = '4.0.40-SNAPSHOT'
```
chrisr3: I think this change must have been accidental. We only change the version number at release time.

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -100,33 +106,79 @@ class CordappPlugin : Plugin<Project> {
         jarTask.dependsOn(task)
     }
 
+    private fun isCordappProject(project: Project): Boolean{
```
chrisr3: The `isCordappProject()` function bothers me, because it looks like something that shouldn't need to exist, i.e. it's asking "Have I applied me to myself?".

By applying the `cordapp` plugin, we should be _telling_ Gradle that _this_ project's POM _only_ needs to be handled differently. Remember that Corda contains multiple `cordapp` modules - `:finance:contracts` and `:finance:workflows` for instance - which means that the `apply` logic will be executed multiple times too.

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -100,33 +106,79 @@ class CordappPlugin : Plugin<Project> {
...
+        return try{
+            project.plugins.getPlugin(net.corda.plugins.CordappPlugin::class.java)
+            true
+        }catch (e: UnknownPluginException){
+            false
+        }
+    }
+
+    private fun configurePomCreation(project: Project) {
+        project.gradle.taskGraph.beforeTask { task ->
```
chrisr3: Scary!!! I am sure you _don't_ want to be iterating over Gradle's _entire task graph_ every time you apply the `cordapp` plugin inside a module.

cordapp/src/main/kotlin/net/corda/plugins/StrippedMavenPom.kt
```diff
@@ -0,0 +1,152 @@
...
+import org.gradle.api.Action
+import org.gradle.api.Project
+import org.gradle.api.XmlProvider
+import org.gradle.api.provider.Property
+import org.gradle.api.publish.maven.*
+import org.gradle.api.publish.maven.internal.dependencies.MavenDependencyInternal
+import org.gradle.api.publish.maven.internal.publication.MavenPomDistributionManagementInternal
+import org.gradle.api.publish.maven.internal.publication.MavenPomInternal
+import org.gradle.api.publish.maven.internal.publisher.MavenProjectIdentity
+import java.util.function.Function
```
chrisr3: Unused `import` statement.

cordapp/src/main/kotlin/net/corda/plugins/StrippedMavenPom.kt
```diff
@@ -0,0 +1,152 @@
+package net.corda.plugins;
+
+import org.gradle.api.Action
+import org.gradle.api.Project
+import org.gradle.api.XmlProvider
+import org.gradle.api.provider.Property
+import org.gradle.api.publish.maven.*
+import org.gradle.api.publish.maven.internal.dependencies.MavenDependencyInternal
```
chrisr3: Internal Gradle interfaces and APIs? Does this impose any version compatibility constraints on our uses of Gradle between Open Source, Enterprise and our Gradle Plugins projects?

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
roastario: because of the way that we apply the maven-publish plugin indirectly via the publish-utils project, the generateMavenPom task is not available when the cordapp plugin is applied, but rather it is only available after gradle has configured itself.

I have not been able to figure out a way to latch onto the generateMavenPom task without using this approach.

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
roastario: It is applied multiple times, and you are right - this is worrying - but I literally cannot find a way of hooking into the pom task in a better way. 

And yes, this is to guard against it being applied to generateMavenPom tasks outside of the current project.

build.gradle
roastario: done done.

cordapp/src/main/kotlin/net/corda/plugins/StrippedMavenPom.kt
roastario: In Maven's infinite wisdom - you *have* to use the Internal version. Because if you do not use the internal version of things, you get a nasty class cast error. 

I mean, look at this:
https://github.com/gradle/gradle/blob/42a750cf566da31cda1bf18437abaf3c40fcc659/subprojects/maven/src/main/java/org/gradle/api/publish/maven/tasks/GenerateMavenPom.java#L119

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
roastario: I'll investigate along the lines of what the Shadow plugin does:
https://github.com/johnrengelman/shadow/blob/bdf2b01df548e961d2ed202ad4214305f65d1c7e/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy#L66

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
chrisr3: This is the problem that all Gradle Plugin Writers are perpetually bashing their heads against - you cannot control the order in which Gradle executes `apply` blocks, and configuration information from users isn't available until all plugins have been applied and evaluation is complete.

chrisr3:
Have you seen the [release notes for Gradle 5.1](https://docs.gradle.org/5.1.1/release-notes.html#inherited-configuration-wide-dependency-excludes-are-now-published)?
> Previously, only exclude rules directly declared on published configurations (e.g. apiElements and runtimeElements for the java component defined by the Java Library Plugin) were published in the Ivy descriptor and POM when using the Ivy Publish Plugin or Maven Publish Plugins, respectively. Now, inherited exclude rules defined on extended configurations (e.g. api for Java libraries) are also taken into account.

Does this describe what we want to achieve here? Because if so, we might consider upgrading Gradle instead.

josecoll:
Discussed with @chrisr3 and perhaps we should defer this story until post Corda 4 (and consider using new Gradle 5 capabilities). 
We can live with the explicit transitive dependency exclusions for now right?

chrisr3:
Tried to compile this against Gradle 5.1.1 :cry::
```
e: /home/chris/workspace/corda-gradle-plugins/cordapp/src/main/kotlin/net/corda/plugins/StrippedMavenPom.kt: (13, 1): Class 'StrippedMavenPom' is not abstract and does not implement abstract member public abstract fun getImportDependencyManagement(): (Mutable)Set<MavenDependency!>! defined in org.gradle.api.publish.maven.internal.publication.MavenPomInternal
```

roastario:
We can, but any ENT downstream consumers of corda-finance-contracts, will also have to do this. 

It's an edge case where OS would be fine without the exclusion, but then if they switched to ENT all their tests will start failing for odd reasons (os core would be on the testRuntime classpath)

chrisr3:
@roastario Developers _are_ supposed to be able to modify POMs via the
```gradle
publishing {
    publications {
        "name of publication" {
            pom {
                // Do stuff in 'ere!
            }
        }
    }
}
```
extension block. So would it be possible to (say) add a `doFirst` handler into the relevant publish task(s) that would tweak the POM XML?

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
chrisr3: Why doesn't this work?
```kotlin
project.tasks.forEach { task ->
    if (task.name.startsWith("generatePomFileFor")) {
```

chrisr3: Perhaps even:
```kotlin
project.tasks.filter { it.name.startsWith("generatePomFile") }.forEach { task ->
```

chrisr3: Or possibly:
```kotlin
project.configure(project.tasks) { task ->
    if (task.name.startsWith("generatePomFile")) {
        task.doFirst { aboutToExecute ->
```

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -108,33 +111,63 @@ class CordappPlugin : Plugin<Project> {
...
+        //TODO if we intend cordapp jars to define transitive dependencies instead of just being fat
+        //we need to use the final artifact name, not the project name
+        //for example, project(":core") needs to be translated into net.corda:corda-core
+
+        return project.configuration("cordapp").allDependencies +
+                project.configuration("cordaCompile").allDependencies +
+                project.configuration("cordaRuntime").allDependencies
+    }
+
+    private fun hardCodedExcludes(): Set<Pair<String, String>>{
```
Clintonio: You could make this into a value.

PR #158, opened 2019-01-11
Update changelog and publish 4.0.38

josecoll:
Includes following changes:

- cordformation: deployNodes does not attempt to sign any JARs unless explicitly configured to do so using signing.enabled = true
- cordapp: enforce CorDapp workflow and contract version identifier metadata to be an integer.
- cordapp: relax cordapp metadata checking (warn on missing metadata rather than failing-fast).
This allows multi-module structured CorDapps to only define metadata for flows and/or contracts modules.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
     ext {
-        gradle_plugins_version = '4.0.38-SNAPSHOT'
+        gradle_plugins_version = '4.0.38'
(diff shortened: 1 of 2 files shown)


josecoll:
Awaiting @roastario PR https://github.com/corda/corda-gradle-plugins/pull/155

Clintonio:
(And update the changelog with the new version)

PR #160, opened 2019-01-12
Update cordapp plugin to modify GenerateMavenPom tasks once the task-graph is ready.

chrisr3:
We need to be _sure_ that Gradle has finished creating the entire project's dynamic tasks before we can reasonably expect to find any `GenerateMavenPom` tasks. Gradle provides the
```kotlin
withType(Class<S> type, Action<? super S> configureAction)
```
API for _precisely_ this purpose. (And I really should have remembered this before now  :man_facepalming:).

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
@@ -120,8 +120,7 @@ class CordappPlugin : Plugin<Project> {
                 project.logger.info("Modifying task: ${task.name} in project ${project.path} to exclude all dependencies from pom")
                 val pom = task.pom
                 if (pom is MavenPomInternal) {
-                    val filteredPom = FilteredPom(pom)
-                    task.pom = filteredPom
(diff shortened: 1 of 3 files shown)


cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -118,15 +115,13 @@ class CordappPlugin : Plugin<Project> {
...
-                        val filteredPom = FilteredPom(pom)
-                        aboutToExecute.pom = filteredPom
-                    }
+        project.tasks.withType(GenerateMavenPom::class.java) { task ->
+            task.doFirst { _ ->
+                project.logger.info("Modifying task: ${task.name} in project ${project.path} to exclude all dependencies from pom")
+                val pom = task.pom
+                if (pom is MavenPomInternal) {
+                    val filteredPom = FilteredPom(pom)
+                    task.pom = filteredPom
```
Clintonio: My only suggestion would be to collapse this into `task.pom = FilteredPom(pom)`

PR #165, opened 2019-01-20
CORDA-2474: Remove Kotlin library from published POMs for Kotlin plugins.

chrisr3:
Gradle plugins written in Kotlin will use Gradle's own provided Kotlin libraries and so should not be published with dependencies on different ones.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,11 +39,11 @@ allprojects {
     version gradle_plugins_version
     group 'net.corda.plugins'
 
-    tasks.withType(JavaCompile).all {
+    tasks.withType(JavaCompile) {
         options.encoding = 'UTF-8'
     }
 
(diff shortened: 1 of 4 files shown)


Clintonio:
I'd suggest moving kotlin dependency declarations to the root project and making a new configuration called "kotlinProjects" that group together common aspects of our kotlin plugins, especially since all plugins going forwards should be in Kotlin.

Clintonio:
(See the marketplace build.gradle for something similar, we should adopt similar standards on all projects)

chrisr3:
I did consider something like that, although was concerned that IntelliJ might not handle it correctly. Although I'm also hoping that Gradle will implement [#7721](https://github.com/gradle/gradle/issues/7721) at some point, which should simplify matters.

And to be fair, the reason I implemented the API Scanner plugin in Java in the first place was so that I didn't need to worry about the Kotlin runtime library.

PR #173, opened 2019-02-27
CORDA-2676: Unregister all SecurityProvider instances installed by Network Bootstrapper.

chrisr3:
Remove any `SecurityProvider` instances installed by the Network Bootstrapper because the JVM cannot discard their classes until all references to them are destroyed. We also want to ensure that the Network Bootstrapper links consistently to everything inside the single `ClassLoader` provided by `cordformation`, and not to any `ClassLoader` from a previous `deployNodes` invocation.

Also revert the `deployNodes` task back to using the runtime classpath because none of the extra jars on the test runtime classpath is needed by the Bootstrapper. Any extra SLF4J backend can be added to the Gradle's `runtimeOnly` configuration.

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
@@ -234,31 +234,37 @@ open class Baseform : DefaultTask() {
             } catch (e: InvocationTargetException) {
                 throw e.cause!!.let { InvalidUserCodeException(it.message ?: "", it) }
             } finally {
-                /*
-                 * Ensure we remove any [SecurityProvider] instances that the
-                 * Network Bootstrapper may have registered. Otherwise the JVM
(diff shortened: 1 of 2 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -230,6 +233,30 @@ open class Baseform : DefaultTask() {
                 bootstrapMethod.invoke(networkBootstrapper, rootDir, allCordapps)
             } catch (e: InvocationTargetException) {
                 throw e.cause!!.let { InvalidUserCodeException(it.message ?: "", it) }
+            } finally {
+                /*
```
Clintonio: Style: Split into a function called something like "cleanupBootstrapperClass(class)"

chrisr3: :+1:

PR #182, opened 2019-04-17
add dependency information about what classes are important for contracts

roastario:

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/ByteArrayStuff.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/ByteArrayStuff.kt
@@ -1,0 +1,284 @@
+package net.corda.plugins
+
+import java.io.ByteArrayInputStream
+import java.io.OutputStream
+import java.nio.ByteBuffer
+import java.security.MessageDigest
+import java.util.*
+import javax.xml.bind.DatatypeConverter
+
+open class OpaqueBytes(bytes: ByteArray) : ByteSequence(bytes, 0, bytes.size) {
+    companion object {
+        /**
+         * Create [OpaqueBytes] from a sequence of [Byte] values.
+         */
+        @JvmStatic
+        fun of(vararg b: Byte) = OpaqueBytes(byteArrayOf(*b))
+    }
+
+    init {
+        require(bytes.isNotEmpty()) { "Byte Array must not be empty" }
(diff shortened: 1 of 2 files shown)


cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -20,6 +30,12 @@ class CordappPlugin : Plugin<Project> {
     private companion object {
         private const val UNKNOWN = "Unknown"
         private const val MIN_GRADLE_VERSION = "4.0"
+
+        val contractCordappConfigurationName = "contractCordapp"
```
chrisr3: These strings should all be `const`.

cordapp/src/main/kotlin/net/corda/plugins/CordappPlugin.kt
```diff
@@ -79,6 +99,26 @@ class CordappPlugin : Plugin<Project> {
         jarTask.dependsOn(task)
     }
 
+    private fun addContractDependenciesListToJar(project: Project, directNonCordaDependencies: MutableSet<File>) {
+        val hashes = calculateContractDependencyHashes(project)
+        val cordInfFolder = File(project.buildDir, "CORDA-DEPS")
+        cordInfFolder.mkdir()
+        val backingFile = File(cordInfFolder, "deps.zip")
```
chrisr3: Doesn't Gradle contain functionality to do this kind of thing already? _It makes JARs..._

cordapp/src/main/kotlin/net/corda/plugins/ByteArrayStuff.kt
```diff
@@ -0,0 +1,284 @@
...
+     *
+     * Cloning like this may become a performance issue, depending on whether or not the JIT
+     * compiler is ever able to optimise away the clone. In which case we may need to revisit
+     * this later.
+     */
+    final override val bytes: ByteArray = bytes
+        get() = field.clone()
+}
+
+sealed class ByteSequence(private val _bytes: ByteArray, val offset: Int, val size: Int) : Comparable<ByteSequence> {
```
chrisr3: Why not use `ByteBuffer`? (This code has been copied and pasted from Corda, hasn't it... :wink:)

PR #210, opened 2019-05-07
CORDA-2924: JDK11 support: upgrade quasar to 0.8.0

josecoll:
Upgrade "quasar-utils" plugin to use quasar 0.8.0

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ## Version 5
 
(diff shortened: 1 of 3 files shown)


chrisr3:
This has been postponed until Corda 6, but instrumented Quasar code needs these transitive dependencies at runtime anyway.

Issue #216, opened 2019-05-09
Exclusions in jvmArgs are missing


Kokogino:
The jvm arg of the quasar plugin is set via `jvmArgs "-javaagent:${project.configurations.quasar.singleFile}"` which results in `-javaagent:lib/quasar.jar`

An actual node has the jvm arg like:
```
-javaagent:lib/quasar.jar=x(antlr**;bftsmart**;co.paralleluniverse**;com.codahale**;com.esotericsoftware**;com.fasterxml**;com.google**;com.ibm**;com.intellij**;com.jcabi**;com.nhaarman**;com.opengamma**;com.typesafe**;com.zaxxer**;de.javakaffee**;groovy**;groovyjarjarantlr**;groovyjarjarasm**;io.atomix**;io.github**;io.netty**;jdk**;junit**;kotlin**;net.bytebuddy**;net.i2p**;org.apache**;org.assertj**;org.bouncycastle**;org.codehaus**;org.crsh**;org.dom4j**;org.fusesource**;org.h2**;org.hamcrest**;org.hibernate**;org.jboss**;org.jcp**;org.joda**;org.junit**;org.mockito**;org.objectweb**;org.objenesis**;org.slf4j**;org.w3c**;org.xml**;org.yaml**;reflectasm**;rx**;org.jolokia**)
```

All the exclusions in the quasar plugin are missing.
Visible on https://github.com/corda/corda/pull/4808/files

chrisr3:
These exclusions on the Node look suspicious to me. Why would a Production Node need to exclude JUnit, Hamcrest and Mockito, for example? But adding the ability to exclude packages for Node Driver tests would definitely be a Good Thing.

chrisr3:
Added to releases 4.0.45 and 5.0.1.

PR #217, opened 2019-05-09
Added exclusions in jvmArgs.

Kokogino:
This is the pull-request for the [Issue-216](https://github.com/corda/corda-gradle-plugins/issues/216)

As described in the issue, the missing exclusions were added to the jvmArgs.

I agree to the agreement in CONTRIBUTING.md.

### DIFF
--- a/quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
+++ b/quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
@@ -36,11 +36,11 @@ class QuasarPlugin implements Plugin<Project> {
         project.dependencies.add("compileOnly", quasar)
 
         project.tasks.withType(Test) {
-            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}"
+            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}=x(antlr**;bftsmart**;co.paralleluniverse**;com.codahale**;com.esotericsoftware**;com.fasterxml**;com.google**;com.ibm**;com.intellij**;com.jcabi**;com.nhaarman**;com.opengamma**;com.typesafe**;com.zaxxer**;de.javakaffee**;groovy**;groovyjarjarantlr**;groovyjarjarasm**;io.atomix**;io.github**;io.netty**;jdk**;junit**;kotlin**;net.bytebuddy**;net.i2p**;org.apache**;org.assertj**;org.bouncycastle**;org.codehaus**;org.crsh**;org.dom4j**;org.fusesource**;org.h2**;org.hamcrest**;org.hibernate**;org.jboss**;org.jcp**;org.joda**;org.junit**;org.mockito**;org.objectweb**;org.objenesis**;org.slf4j**;org.w3c**;org.xml**;org.yaml**;reflectasm**;rx**;org.jolokia**)"
             jvmArgs "-Dco.paralleluniverse.fibers.verifyInstrumentation"
         }
         project.tasks.withType(JavaExec) {
-            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}"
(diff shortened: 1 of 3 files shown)


quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
```diff
@@ -36,11 +36,11 @@ class QuasarPlugin implements Plugin<Project> {
         project.dependencies.add("compileOnly", quasar)
 
         project.tasks.withType(Test) {
-            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}"
+            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}=x(antlr**;bftsmart**;co.paralleluniverse**;com.codahale**;com.esotericsoftware**;com.fasterxml**;com.google**;com.ibm**;com.intellij**;com.jcabi**;com.nhaarman**;com.opengamma**;com.typesafe**;com.zaxxer**;de.javakaffee**;groovy**;groovyjarjarantlr**;groovyjarjarasm**;io.atomix**;io.github**;io.netty**;jdk**;junit**;kotlin**;net.bytebuddy**;net.i2p**;org.apache**;org.assertj**;org.bouncycastle**;org.codehaus**;org.crsh**;org.dom4j**;org.fusesource**;org.h2**;org.hamcrest**;org.hibernate**;org.jboss**;org.jcp**;org.joda**;org.junit**;org.mockito**;org.objectweb**;org.objenesis**;org.slf4j**;org.w3c**;org.xml**;org.yaml**;reflectasm**;rx**;org.jolokia**)"
```
chrisr3: These exclusions are just an optimisation for running the tests, but the problem is that this kind of information should be configurable instead of hard-coded into the plugin. So (for example), the `quasar-utils` plugin would register a Gradle extension with a `List<String>` property and then a user could populate that property with package-exclusion patterns. The plugin would then "map" those patterns into a correctly-formatted `jvmArgs` value for all `Test` and `JavaExec` tasks.

The issue with hard-coded configuration data is that someone will _always_ want to modify it, and that then requires changing the code.

quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
```diff
@@ -36,11 +36,11 @@ class QuasarPlugin implements Plugin<Project> {
...
             jvmArgs "-Dco.paralleluniverse.fibers.verifyInstrumentation"
         }
         project.tasks.withType(JavaExec) {
-            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}"
+            jvmArgs "-javaagent:${project.configurations.quasar.singleFile}=x(antlr**;bftsmart**;co.paralleluniverse**;com.codahale**;com.esotericsoftware**;com.fasterxml**;com.google**;com.ibm**;com.intellij**;com.jcabi**;com.nhaarman**;com.opengamma**;com.typesafe**;com.zaxxer**;de.javakaffee**;groovy**;groovyjarjarantlr**;groovyjarjarasm**;io.atomix**;io.github**;io.netty**;jdk**;junit**;kotlin**;net.bytebuddy**;net.i2p**;org.apache**;org.assertj**;org.bouncycastle**;org.codehaus**;org.crsh**;org.dom4j**;org.fusesource**;org.h2**;org.hamcrest**;org.hibernate**;org.jboss**;org.jcp**;org.joda**;org.junit**;org.mockito**;org.objectweb**;org.objenesis**;org.slf4j**;org.w3c**;org.xml**;org.yaml**;reflectasm**;rx**;org.jolokia**)"
```
chrisr3: You shouldn't have needed to duplicate this information.

quasar-utils/src/main/groovy/net/corda/plugins/QuasarPlugin.groovy
Kokogino: Unfortunately I have no experience with gradle/groovy. After looking into it for a while, this is what I came up with (the new commit). I hope this meets your expectations.

chrisr3: Actually this is good, although not _quite_ what I meant :wink:. I think we should also handle the case where `quasar_extensions` is empty, in which case there wouldn't be anything to append to the `-javaagent` parameter.

Kokogino: Well I'm glad it's good :)
I added a check for an empty list of exclusions.

chrisr3:
I still think that a new `quasar` extension would be better than another property, but extensions in Gradle are tricky and so I'll merge this and see how realistic creating an extension is after that.

PR #234, opened 2019-06-06
migrate dockerForm to use official images

roastario:


cordapp/src/main/kotlin/net/corda/plugins/Utils.kt
```diff
@@ -14,7 +14,13 @@ import kotlin.math.max
  * accessor to the "ext" extension (See: ExtraPropertiesExtension)
  */
 @Suppress("UNCHECKED_CAST")
-fun <T : Any> Project.ext(name: String): T = (extensions.findByName("ext") as ExtraPropertiesExtension).get(name) as T
+fun <T : Any> Project.ext(name: String): T? {
```
chrisr3: _Wibble?!_ The `quasar-utils` plugin reads the `quasar_group`, `quasar_version` etc properties using just
```groovy
def quasarGroup = rootProject.hasProperty('quasar_group') ? rootProject.property('quasar_group') : defaultGroup
```
Do you _need_ this extension function?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -33,8 +33,16 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
 
     private val directoryPath: Path = project.projectDir.toPath().resolve(directory)
 
-    @get:InputFile
-    val dockerComposePath: Path = directoryPath.resolve("docker-compose.yml")
+
+    val dockerComposePath: Path
+        @get:InputFile
```
chrisr3: Either
```kotlin
@get:InputFile
val dockerComposerPath: Path
```
or
```kotlin
val dockerComposerPath: Path
   @InputFile
   get()
```

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -43,18 +51,35 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
     fun build() {
         project.logger.lifecycle("Running Cordform task")
         initializeConfiguration()
-        nodes.forEach(Node::installDockerConfig)
+        nodes.forEach { it -> it.installDockerConfig(22022) }
```
chrisr3: Magic Number Bad. Create `const val` inside the companion object.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -109,9 +109,10 @@ open class Node @Inject constructor(private val project: Project) {
      *
      * @return This node's P2P address.
      */
-    val p2pAddress: String
+    @get:Optional
+    val p2pAddress: String?
         @Input
-        get() = config.getString("p2pAddress")
```
chrisr3: Pick _one_ convention for applying these annotations, and then stick with it?
```kotlin
val p2pAddress: String?
    @Optional
    @Input
    get() = ...
```
You could even write _this_, if you were feeling _particularly_ perverse...
```kotlin
@get:[Optional Input]
val p2pAddress: String?
    get() = ...
```

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -63,7 +63,7 @@ open class Node @Inject constructor(private val project: Project) {
     internal var rpcPort = 10003
         @Input get
         private set
-    private var config = ConfigFactory.empty()
+    var config = ConfigFactory.empty()
```
chrisr3: Perhaps make this property `internal` rather than leave it as `public`?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -43,18 +51,35 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
     fun build() {
         project.logger.lifecycle("Running Cordform task")
```
chrisr3: No, it's the "Dockerform" task.

roastario:
Hokay, I've finally settled down on a single way of declaring an Input. 

Will take a look at adding testing for docker tomorrow.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -370,14 +376,8 @@ open class Node @Inject constructor(private val project: Project) {
     }
 
     internal fun buildDocker() {
-        project.copy {
-            it.apply {
-                from(Cordformation.getPluginFile(project, "Dockerfile"))
```
szymonsztuka: Should `/cordformation/src/main/resources/net/corda/plugins/Dockerfile` be removed as well?

roastario: yup, removed!

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -18,51 +18,80 @@ import javax.inject.Inject
...
+                            "$nodeBuildDir/node.conf:/etc/corda/node.conf",
+                            "$nodeBuildDir/certificates:/opt/corda/certificates",
+                            "$nodeBuildDir/logs:/opt/corda/logs",
+                            "$nodeBuildDir/persistence:/opt/corda/persistence",
+                            "$nodeBuildDir/cordapps:/opt/corda/cordapps",
+                            "$nodeBuildDir/network-parameters:/opt/corda/network-parameters",
+                            "$nodeBuildDir/additional-node-infos:/opt/corda/additional-node-infos",
+                            "$nodeBuildDir/drivers:/opt/corda/drivers"
+
+                    ),
```
VCAMP: @roastario does this need to also expose the `p2pAddress` port?

Also, if this doesn't bind the ports to the corresponding host ports, will the generated `node.conf` files be valid? (as `docker compose` chooses an ephemeral port)

roastario: as the nodes will be within the same docker network, you do not need to expose the p2p ports. 

As for the rpc ports, I don't think we need to bind them to a specific port as this would mean that if two networks were started they would start clashing.

VCAMP: Thanks for clarifying. Tested it manually, looks like it's all working.

cordapp/src/main/kotlin/net/corda/plugins/Utils.kt
```diff
@@ -13,8 +12,19 @@ import kotlin.math.max
  * Mimics the "project.ext" functionality in groovy which provides a direct
  * accessor to the "ext" extension (See: ExtraPropertiesExtension)
  */
-@Suppress("UNCHECKED_CAST")
-fun <T : Any> Project.ext(name: String): T = (extensions.findByName("ext") as ExtraPropertiesExtension).get(name) as T
+
```
chrisr3: I don't understand this function. Are our current semantics that `corda_release_version` _must_ be defined in `rootProject`? Does this changes those semantics? It looks like it's searching in every project between here and the root, which raises the possibility that someone will specify different values en route :worried:.

When I mentioned the `quasar-utils` plugin, I was only asking whether `rootProject.hasProperty(...)` and `rootProject.property(...)` would suffice ... :grin:

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -18,51 +18,80 @@ import javax.inject.Inject
  */
 @Suppress("unused")
 open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(objects) {
+
     private companion object {
-        const val nodeJarName = "corda.jar"
-        private val defaultDirectory: Path = Paths.get("build", "docker")
+        private const val DEFAULT_SSH_PORT = 22022;
```
chrisr3: _**AAGH, MY EYES!!**_ _A semi-colon!_ :see_no_evil:

cordapp/src/main/kotlin/net/corda/plugins/Utils.kt
```diff
@@ -40,14 +42,15 @@ class Utils {
...
 
         @JvmStatic
-        fun compareVersions(v1 : String, v2 : String) : Int {
-            fun parseVersionString(v : String) = v.split(".").flatMap { it.split("-") }.map {
+        fun compareVersions(v1: String, v2: String): Int {
+            fun parseVersionString(v: String) = v.split(".").flatMap { it.split("-") }.map {
                 try {
                     Integer.valueOf(it)!!
-                } catch (e : NumberFormatException) {
+                } catch (e: NumberFormatException) {
```
chrisr3: _**DEATH TO THE EVIL WHITESPACE!**_ :clap:

cordapp/src/main/kotlin/net/corda/plugins/Utils.kt
```diff
@@ -13,8 +12,11 @@ import kotlin.math.max
  * Mimics the "project.ext" functionality in groovy which provides a direct
  * accessor to the "ext" extension (See: ExtraPropertiesExtension)
  */
-@Suppress("UNCHECKED_CAST")
-fun <T : Any> Project.ext(name: String): T = (extensions.findByName("ext") as ExtraPropertiesExtension).get(name) as T
+
+fun <T : Any> Project.findRootProperty(name: String): T? {
+    return rootProject.property(name) as T?
```
chrisr3: Good, but this will throw an exception if the property cannot be found (see [here](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#property-java.lang.String-)). If you want it to return `null` for absent properties - and I think you _do_ - then you'll need to use [findProperty](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#findProperty-java.lang.String-) instead.

PR #248, opened 2019-09-16
CORDA-3024 Rename corda-webserver

tudor-malene:
see https://r3-cev.atlassian.net/browse/CORDA-3024

This change is needed to enable the renaming in the main repo

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -445,7 +445,7 @@ open class Node @Inject constructor(private val project: Project) {
             try {
                 Cordformation.verifyAndGetRuntimeJar(project, "corda-testserver")
             } catch (e: IllegalStateException) {
-                project.logger.lifecycle("Detecting older version of of corda. Falling back to the old webserver.")
+                project.logger.lifecycle("Detecting older version of corda. Falling back to the old webserver.")
                 Cordformation.verifyAndGetRuntimeJar(project, "corda-webserver")
             }
         } else {
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -25,7 +25,7 @@ open class Node @Inject constructor(private val project: Project) {
     internal data class ResolvedCordapp(val jarFile: Path, val config: String?)
 
     private companion object {
-        const val webJarName = "corda-webserver.jar"
+        const val webJarName = "corda-testserver.jar"
         const val configFileProperty = "configFile"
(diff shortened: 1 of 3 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -442,7 +442,7 @@ open class Node @Inject constructor(private val project: Project) {
         // If no webserver JAR is provided, the default development webserver is used.
         val webJar = if (webserverJar == null) {
             project.logger.lifecycle("Using default development webserver.")
-            Cordformation.verifyAndGetRuntimeJar(project, "corda-webserver")
+            Cordformation.verifyAndGetRuntimeJar(project, "corda-testserver")
```
chrisr3: Is this the name of the Gradle module in the Corda repository? Because if so then this change will break the `cordformatition` plugin for other versions of Corda!

The plugins are supposed to be backwards-compatible when posssible.

tudor-malene: Good point.
I was afraid of that. 
I'll do that

tudor-malene:
Back to you @chrisr3

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -7,7 +7,7 @@ import java.util.*
 private const val HEADLESS_FLAG = "--headless"
 private const val CAPSULE_DEBUG_FLAG = "--capsule-debug"
 private const val CORDA_JAR_NAME = "corda.jar"
-private const val CORDA_WEBSERVER_JAR_NAME = "corda-webserver.jar"
+private const val CORDA_WEBSERVER_JAR_NAME = "corda-testserver"
```
chrisr3: Why doesn't this filename end in `.jar`?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -442,7 +442,12 @@ open class Node @Inject constructor(private val project: Project) {
         // If no webserver JAR is provided, the default development webserver is used.
         val webJar = if (webserverJar == null) {
             project.logger.lifecycle("Using default development webserver.")
-            Cordformation.verifyAndGetRuntimeJar(project, "corda-webserver")
+            try {
+                Cordformation.verifyAndGetRuntimeJar(project, "corda-testserver")
+            } catch (e: IllegalStateException) {
+                project.logger.lifecycle("Detecting older version of of corda. Falling back to the old webserver.")
```
chrisr3: Eeek! It says "of of Corda"!!!

PR #249, opened 2019-09-17
CORDA-3228: Update Quasar utils plugin to support excluding classloaders.

chrisr3:
This adds an `excludeClassLoaders = []` option to the `quasar` extension. Currently, only Quasar 0.7.12_r3 and above (excluding 0.8.0) supports this.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,7 +4,10 @@
 
 ### Version 5.0.3
 
-* `webserver`: The `corda-webserver` component has been renamed to `corda-testserver.`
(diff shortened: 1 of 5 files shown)


tudor-malene:
Code looks good.
Maybe describe a bit how `excludeClassLoaders ` actually works.

chrisr3:
> Maybe describe a bit how `excludeClassLoaders ` actually works.

Hmm, I'm not sure I should be quoting the Gradle API docs here... :thinking:

PR #251, opened 2019-09-19
CORDA-3224 Java 11 support tweaks.

josecoll:
* `cordformation`: remove hard dependency on java 1.8 for macos in runnodes (and allow for usage of JAVA_HOME if set)

* `cordformation`: add support for classifier (eg. jdk11) in detection of corda runtime artifacts (eg. corda.jar, test-server.jar).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ import static org.gradle.api.JavaVersion.*
 
 buildscript {
     ext {
-        gradle_plugins_version = '5.0.4-SNAPSHOT'
+        gradle_plugins_version = '5.0.4'
         typesafe_config_version = '1.3.1'
         kotlin_version = '1.3.21'
         snake_yaml_version = '1.19'
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@ import static org.gradle.api.JavaVersion.*
 
 buildscript {
     ext {
-        gradle_plugins_version = '5.0.4'
+        gradle_plugins_version = '5.0.4-SNAPSHOT'
         typesafe_config_version = '1.3.1'
         kotlin_version = '1.3.21'
         snake_yaml_version = '1.19'
--- a/changelog.md
+++ b/changelog.md
@@ -2,8 +2,6 @@
 
 ## Version 5
 
-### Version 5.0.5
-
 ### Version 5.0.4
 
 * `cordformation`: remove hard dependency on java 1.8 for macos in runnodes (and allow for usage of JAVA_HOME if set)
--- a/changelog.md
+++ b/changelog.md
@@ -8,7 +8,7 @@
 
 * `cordformation`: remove hard dependency on java 1.8 for macos in runnodes (and allow for usage of JAVA_HOME if set)
 
-* `cordformation`: add support for classifier (eg. jdk11) in detection of corda.jar file.  
+* `cordformation`: add support for classifier (eg. jdk11) in detection of corda runtime artifacts (eg. corda.jar, test-server.jar).  
 
 ### Version 5.0.3
 
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,13 @@
 
 ## Version 5
 
+### Version 5.0.5
+
 ### Version 5.0.4
+
+* `cordformation`: remove hard dependency on java 1.8 for macos in runnodes (and allow for usage of JAVA_HOME if set)
+
+* `cordformation`: add support for classifier (eg. jdk11) in detection of corda.jar file.  
 
 ### Version 5.0.3
 
--- a/cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
@@ -37,16 +37,13 @@ class Cordformation : Plugin<Project> {
          * @param jarName The name of the JAR you wish to access.
          * @return A file handle to the file in the JAR.
          */
-        private const val VERSION_ID_PATTERN = "\\d\\.\\d(-SNAPSHOT|-RC\\d{2}|.\\d{8})?"
-        private val CORDA_JAR_REGEX = ".*corda-$VERSION_ID_PATTERN(-.*)?\\.jar\$".toRegex(RegexOption.IGNORE_CASE)
-        private val CORDA_ENT_JAR_REGEX = ".*corda-enterprise-$VERSION_ID_PATTERN(-.*)?\\.jar\$".toRegex(RegexOption.IGNORE_CASE)
-
         fun verifyAndGetRuntimeJar(project: Project, jarName: String): File {
             val releaseVersion = project.findRootProperty<String>("corda_release_version")
                     ?: throw IllegalStateException("Could not find a valid declaration of \"corda_release_version\"")
+            // need to cater for optional classifier (eg. corda-4.3-jdk11.jar)
+            val pattern = "\\Q$jarName\\E(-enterprise)?-\\Q$releaseVersion\\E(-.+)?\\.jar\$".toRegex()
             val maybeJar = project.configuration("runtime").filter {
-                it.toString().contains(CORDA_ENT_JAR_REGEX) ||
-                it.toString().contains(CORDA_JAR_REGEX)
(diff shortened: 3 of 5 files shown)


build.gradle
```diff
@@ -2,7 +2,7 @@ import static org.gradle.api.JavaVersion.*
 
 buildscript {
     ext {
-        gradle_plugins_version = '5.0.4-SNAPSHOT'
```
chrisr3: Releasing a new version of the plugins is done _separately_ from adding a new feature.

changelog.md
```diff
@@ -2,8 +2,14 @@
 
 ## Version 5
 
+### Version 5.0.5
```
chrisr3: Not yet...

cordformation/src/main/resources/net/corda/plugins/runnodes
```diff
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -eo pipefail
```
chrisr3: What does this do?

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -41,7 +41,8 @@ class Cordformation : Plugin<Project> {
             val releaseVersion = project.findRootProperty<String>("corda_release_version")
                     ?: throw IllegalStateException("Could not find a valid declaration of \"corda_release_version\"")
             val maybeJar = project.configuration("runtime").filter {
-                "$jarName-$releaseVersion.jar" in it.toString() || "$jarName-enterprise-$releaseVersion.jar" in it.toString()
+                it.toString().contains("$jarName-enterprise-$releaseVersion(-*)?.jar".toRegex()) ||
```
chrisr3: Compiling a `Regexp` is expensive. Can you factor these out please to ensure that they're only compiled once?

I suspect you want `(-.*)?\\.jar\$` at the endof the "enterprise" one, though.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -41,7 +41,8 @@ class Cordformation : Plugin<Project> {
...
+                it.toString().contains("$jarName-$releaseVersion(-.*)?.jar".toRegex())
```
chrisr3: This one should probebly end with `\\.jar\$` because a literal `.` needs to be escaped in a regexp.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
josecoll: Given the frequency of use of the plugins I hardly think this is a performance issue.

josecoll: The regex's have been tested to work.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
josecoll: The regex's have been tested to work.

cordformation/src/main/resources/net/corda/plugins/runnodes
josecoll: Causes the shell to interpret variables differently.

build.gradle
josecoll: OK - will revert this and create a separate PR to bump the version number (what a waste of time).

changelog.md
josecoll: ok ok ...

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
chrisr3: I suspect the `-enterprise` one _hasn't_ been tested because it says `(-*)`, which means "0 or more occurrences of `-`.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
chrisr3: A `.` matches _any_ single character (including a literal `.`) whereas `\\.` would match _only_ a literal `.`. So we have different definitions of "work" here.

cordformation/src/main/resources/net/corda/plugins/runnodes
chrisr3: This one?
> Treat unset variables and parameters other than the special parameters "@" and "*" as an error when  performing parameter  expansion.   If  expansion is attempted on an unset variable or parameter, the shell prints  an  error message,  and, if not interactive, exits with a non-zero status.

Hmm, why remove this?

josecoll: because the expansion of JAVA_HOME fails when it is set.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -37,11 +37,16 @@ class Cordformation : Plugin<Project> {
          * @param jarName The name of the JAR you wish to access.
          * @return A file handle to the file in the JAR.
          */
+        private const val VERSION_ID_PATTERN = "\\d\\.\\d(-SNAPSHOT|-RC\\d{2}|.\\d{8})?"
+        private val CORDA_JAR_REGEX = ".*corda-$VERSION_ID_PATTERN(-.*)?\\.jar\$".toRegex(RegexOption.IGNORE_CASE)
+        private val CORDA_ENT_JAR_REGEX = ".*corda-enterprise-$VERSION_ID_PATTERN(-.*)?\\.jar\$".toRegex(RegexOption.IGNORE_CASE)
+
         fun verifyAndGetRuntimeJar(project: Project, jarName: String): File {
```
chrisr3: I just meant defining the regexp pattern outside the _filter_, not outside the function :wink:. The original approach of
```kotlin
$jarName(-enterprise)?-$releaseVersion\\.jar
```
was fine. We only needed to extend it to allow for an arbitrary classifier.
```kotlin
    @Test
    fun regexp() {
        val jarName = "corda"
        val releaseVersion = "1.2.3.4"
        val pattern = "\\Q$jarName\\E(-enterprise)?-\\Q$releaseVersion\\E(-.+)?\\.jar\$".toRegex()
        println(pattern.toString())

        assertTrue(pattern.matches("corda-1.2.3.4.jar"))
        assertTrue(pattern.matches("corda-enterprise-1.2.3.4.jar"))
        assertTrue(pattern.matches("corda-1.2.3.4-pants.jar"))
        assertTrue(pattern.matches("corda-enterprise-1.2.3.4-wibble.jar"))
    }
```

chrisr3: And keep the `IGNORE_CASE` option too, seeing as both MacOS and Windows have case-insensitive filesystems:
```kotlin
    @Test
    fun regexp() {
        val jarName = "corda"
        val releaseVersion = "1.2.3.4"
        val pattern = "\\Q$jarName\\E(-enterprise)?-\\Q$releaseVersion\\E(-.+)?\\.jar\$".toRegex(RegexOption.IGNORE_CASE)

        assertTrue("/absolute/path/corda-1.2.3.4.jar".contains(pattern))
        assertTrue("other/path/corda-enterprise-1.2.3.4.jar".contains(pattern))
        assertTrue("my/path/corda-1.2.3.4-pants.jar".contains(pattern))
        assertTrue("big/path/corda-enterprise-1.2.3.4-wibble.jar".contains(pattern))
        assertFalse("corda-1.2.4-pants.jar".contains(pattern))
        assertFalse("corda-enterprise-1.2.4-wibble.jar".contains(pattern))
    }
```

build.gradle
chrisr3: :+1:

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
chrisr3: Did you want to keep the `RegexOption.IGNORE_CASE`?

PR #258, opened 2019-10-09
make 5005 & 7005 port configuare to avoid port confict if running two…

yywill:

The noderunner command `runnodes` or `runnodes.bat` is great for development, because it allow developers to run mutiple nodes in a single command (say, one batch), Instead of using `java -Dcapsule.jvm.args="-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005" -jar corda.jar`

Howerver, it could be protential problem if there are mutiple `batches` running at the same time. The problem comes here:  [sourcegraph: NodeRunner.kt ](https://sourcegraph.com/github.com/corda/corda-gradle-plugins/-/blob/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt#L25-28)

To solve this, I introduce two configuatable system enviroment variables to support that: 
```zsh
# to overide default starting port: 5005
--debug.port=5105
```

```zsh
# to override jmx starting port 7005
--monitoring.port=7105
```

This would allow developer change the switching starting port 5005++ to avoid `address already be used`

The config could be null, because I applied default value 5005 and 7005 to let legacy code works. 

# PR Checklist:

- [X] Have you run the unit, integration and smoke tests as described here? https://docs.corda.net/head/testing.html
- [X] If you added/changed public APIs, did you write/update the JavaDocs?  
- [X] If the changes are of interest to application developers, have you added them to the changelog, and potentially release notes?
- [x] If you are contributing for the first time, please read the agreement in CONTRIBUTING.md now and add to this Pull Request that you agree to it.

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or

(b) The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or

(c) The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.

(d) I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project or the open source license(s) involved.

### DIFF
--- a/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
+++ b/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
@@ -6,8 +6,8 @@ import java.util.*
 
 private const val HEADLESS_FLAG = "--headless"
 private const val CAPSULE_DEBUG_FLAG = "--capsule-debug"
-private const val BASE_DEBUG_PORT_FLAG = "-Dnet.corda.noderunner.debug.port="
-private const val BASE_MONITORING_PORT_FLAG = "-Dnet.corda.noderunner.monitoring.port="
+private const val BASE_DEBUG_PORT_FLAG = "--debug.port="
+private const val BASE_MONITORING_PORT_FLAG = "--monitoring.port="
 private const val CORDA_JAR_NAME = "corda.jar"
 private const val CORDA_WEBSERVER_JAR_NAME = "corda-testserver.jar"
 private const val OLD_CORDA_WEBSERVER_JAR_NAME = "corda-webserver.jar"
@@ -42,19 +42,19 @@ fun main(args: Array<String>) {
     val headless = ((!isTmux() && GraphicsEnvironment.isHeadless()) || args.contains(HEADLESS_FLAG))
     val capsuleDebugMode = args.contains(CAPSULE_DEBUG_FLAG)
     try{
-        base_debug_port = Integer.valueOf(args.first { it -> it.contains(BASE_DEBUG_PORT_FLAG)}.removePrefix(BASE_DEBUG_PORT_FLAG))
+        base_debug_port = Integer.valueOf(args.first { it -> it.startsWith(BASE_DEBUG_PORT_FLAG)}.removePrefix(BASE_DEBUG_PORT_FLAG))
         println("base_debug_port set to: $base_debug_port")
-    }catch(e: Throwable){
+    }catch(e: NumberFormatException){
         println("base_debug_port set to default： $base_debug_port")
     }
     try{
-        base_monitoring_port = Integer.valueOf(args.first { it -> it.contains(BASE_MONITORING_PORT_FLAG)}.removePrefix(BASE_MONITORING_PORT_FLAG))
+        base_monitoring_port = Integer.valueOf(args.first { it -> it.startsWith(BASE_MONITORING_PORT_FLAG)}.removePrefix(BASE_MONITORING_PORT_FLAG))
         println("base_monitoring_port set to: $base_monitoring_port")
-    }catch(e: Throwable){
+    }catch(e: NumberFormatException){
         println("base_monitoring_port set to default: $base_monitoring_port")
     }
     val workingDir = File(System.getProperty("user.dir"))
-    val javaArgs = args.filter { it != HEADLESS_FLAG && it != CAPSULE_DEBUG_FLAG && !it.contains(BASE_DEBUG_PORT_FLAG) && !it.contains(BASE_MONITORING_PORT_FLAG) }
+    val javaArgs = args.filter { it != HEADLESS_FLAG && it != CAPSULE_DEBUG_FLAG && !it.startsWith(BASE_DEBUG_PORT_FLAG) && !it.startsWith(BASE_MONITORING_PORT_FLAG) }
     val jvmArgs = if (capsuleDebugMode) listOf("-Dcapsule.log=verbose") else emptyList()
     println("Starting nodes in $workingDir")
     workingDir.listFiles { file -> file.isDirectory }.forEach { dir ->
--- a/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
+++ b/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
@@ -6,12 +6,17 @@ import java.util.*
 
 private const val HEADLESS_FLAG = "--headless"
 private const val CAPSULE_DEBUG_FLAG = "--capsule-debug"
+private const val BASE_DEBUG_PORT_FLAG = "-Dnet.corda.noderunner.debug.port="
+private const val BASE_MONITORING_PORT_FLAG = "-Dnet.corda.noderunner.monitoring.port="
 private const val CORDA_JAR_NAME = "corda.jar"
 private const val CORDA_WEBSERVER_JAR_NAME = "corda-testserver.jar"
 private const val OLD_CORDA_WEBSERVER_JAR_NAME = "corda-webserver.jar"
 private const val CORDA_CONFIG_NAME = "node.conf"
 private const val CORDA_WEBSERVER_CONFIG_NAME = "web-server.conf"
 private val CORDA_HEADLESS_ARGS = listOf("--no-local-shell")
+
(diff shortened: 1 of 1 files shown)


yywill:
My project requires two corda networks, So, the port 500x and 700x conficts, because the ports of the two both begins with 5005 and 7005. It would be better, if we can config the starting port to avoid any conflicts.


I aslo make the default value unchanged to keep the port consistent, so any lagacy projects won't break.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -20,15 +20,33 @@ private val os by lazy {
     else OS.LINUX
 }
 
+private val debugBasePort by lazy {
+    val debugport = System.getProperty("debug.base.port", "generic").toLowerCase(Locale.ENGLISH)
```
chrisr3: There are a number of issues here:
- The property name `debug.base.port`does not identify this property as belonging to `runnodes`. Something like `net.corda.noderunner.debug.port` would be better.
- This is an integer property, and so a default value of "generic" makes no sense. It would be better to use:
```kotlin
private val debugBasePort by lazy {
    Integer.getInteger("net.corda.noderunner.debug.port", 5005)
}
```
which handles any possible `NumberFormatException` automatically.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -20,15 +20,33 @@ private val os by lazy {
...
+
+private val monitoringPort by lazy {
+    val monitoringPort = System.getProperty("debug.monitoring.port", "generic").toLowerCase(Locale.ENGLISH)
+    try{
+        Integer.valueOf(monitoringPort ?:"7005")
+    }catch (e: NumberFormatException){
+        7005
+    }
+}
+
```
chrisr3: And similarly here:
```kotlin
private val monitoringBasePort bylazy {
    Integer.getInteger("net.corda.noderunner.monitoring.port", 7005)
}
```

yywill:
> > My project requires two corda networks, So, the port 500x and 700x conficts, because the ports of the two both begins with 5005 and 7005. It would be better, if we can config the starting port to avoid any conflicts.
> > I aslo make the default value unchanged to keep the port consistent, so any lagacy projects won't break.
> 
> This is not needed. DEBUG and JMX ports can be changed when running the node. See https://docs.corda.net/head/node-commandline.html#enabling-remote-debugging

I update my PR description.  please have a look.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
yywill: sovled it by using `system.getEnv`

yywill:
> > I update my PR description. please have a look.
> 
> I am familiar with the concept of setting Java system properties via Java's `-D` command line parameter. But I am still curious about _where_ you intend to add these. Can you walk me through the steps required to launch a node with non-default debug and monitoring ports please?

I bash or bat, `$@` or  %* would passing rest of parameters of comand line to process. 

I guess it would work, but I was wrong about that.

I change `system.getProperty`  to `system.getEnv`, and that would solve bugs. And, I tested it working on my laptop. `-D` may have ordering issues as well, and should abondon such way to pass variables in our cases.

chrisr3:
> I change `system.getProperty` to `system.getEnv`, and that would solve bugs. And, I tested it working on my laptop. `-D` may have ordering issues as well, and should abondon such way to pass variables in our cases.

Actually I disagree. These default port values are still logically better implemented as command line parameters for `runnodes` and `runnodes.bat` rather than as environment variables. The problem is that these system properties would then need to be set as JVM options rather than as arguments for `runnodes.jar`:
```bash
java [jvm-options] -jar runnodes.jar "$@"
```
Both `bash` and Windows `CMD` implement a `shift` function that allow you to renumber positional parameters, which would imply that `-Dnet.corda.noderunner.debug.port=xxx` etc would need to be set before all other command line parameters :thinking:.

yywill:
Hi @chrisr3 

I disagree with `shift`. I tried the `-Dxxx`, and it passes the parameters to the noderunner shell! And, the shell crashed because of `unregonizeed parameters`. I am superrised that it is throwing Errors rather than ignoring unregnized things. 

I guess it would be some confilcts with system shell and java shell emulator. And, a system envirnment would be the only solution for that.

Your solution of change constructor of `NodeRunner`  would be better than all of above.  Let me try to do it.

yywill:
hi @chrisr3 

I did what you said. Looks fine.

Please have a look at my test proofs.

proof-3
proof-1
proof-2

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -6,13 +6,18 @@ import java.util.*
 
 private const val HEADLESS_FLAG = "--headless"
 private const val CAPSULE_DEBUG_FLAG = "--capsule-debug"
+private const val BASE_DEBUG_PORT_FLAG = "-Dnet.corda.noderunner.debug.port="
+private const val BASE_MONITORING_PORT_FLAG = "-Dnet.corda.noderunner.monitoring.port="
 private const val CORDA_JAR_NAME = "corda.jar"
```
chrisr3: These are no longer Java system properties and so we can use flags values like
```
--debug-port
--monitoring-port
```
instead.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -23,21 +28,33 @@ private val os by lazy {
...
 }
 
 fun main(args: Array<String>) {
     val startedProcesses = mutableListOf<Process>()
     val headless = ((!isTmux() && GraphicsEnvironment.isHeadless()) || args.contains(HEADLESS_FLAG))
     val capsuleDebugMode = args.contains(CAPSULE_DEBUG_FLAG)
+    try{
+        base_debug_port = Integer.valueOf(args.first { it -> it.contains(BASE_DEBUG_PORT_FLAG)}.removePrefix(BASE_DEBUG_PORT_FLAG))
+        println("base_debug_port set to: $base_debug_port")
+    }catch(e: Throwable){
```
chrisr3: This should only be catching `NumberFormatException` because anything else is likely to be too serious for us to continue.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -23,21 +28,33 @@ private val os by lazy {
...
+        println("base_debug_port set to default： $base_debug_port")
+    }
+    try{
+        base_monitoring_port = Integer.valueOf(args.first { it -> it.contains(BASE_MONITORING_PORT_FLAG)}.removePrefix(BASE_MONITORING_PORT_FLAG))
+        println("base_monitoring_port set to: $base_monitoring_port")
+    }catch(e: Throwable){
```
chrisr3: Again, this should only be catching `NumberFormatException`.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -23,21 +28,33 @@ private val os by lazy {
...
+        println("base_monitoring_port set to default: $base_monitoring_port")
+    }
     val workingDir = File(System.getProperty("user.dir"))
-    val javaArgs = args.filter { it != HEADLESS_FLAG && it != CAPSULE_DEBUG_FLAG }
+    val javaArgs = args.filter { it != HEADLESS_FLAG && it != CAPSULE_DEBUG_FLAG && !it.contains(BASE_DEBUG_PORT_FLAG) && !it.contains(BASE_MONITORING_PORT_FLAG) }
```
chrisr3: Perhaps `startsWith()` is better than `contains()` here because we're detecting prefixes.

PR #263, opened 2019-10-28
CORDA-3377: Add `excludePackages` to the `scanApi` task.

chrisr3:
Add `excludePackages` to the API Scanner plugin's DSL. This is so that it can ignore all of the classes in the DJVM's `sandbox.**` packages.

I've also simplified some of the code to make it easier to upgrade to ClassGraph.

### DIFF
--- a/api-scanner/README.md
+++ b/api-scanner/README.md
@@ -34,12 +34,21 @@ scanApi {
     // Enable / disable the task within this module.
     enabled = {true|false}
 
+    // The `archiveClassifier` value for the Jar artifact that will be scanned.
+    // This is empty by default, to select the module's primary Jar artifact.
+    targetClassifier = '<classifier>'
+
     // Names of classes that should be excluded from the output.
     excludeClasses = [
         ...
     ]
 
-    //Methods that should be excluded from the output.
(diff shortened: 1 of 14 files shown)


dimosr:
I was not referring to the specific `internal` filtering that's already in place, perhaps the example was not helpful. What I was trying to understand is if a user specifies `excludePackages` equal to `foo`, does this also apply as a class filter (e.g. for classes that contain `foo`) implicitly due to the way this version of classgraph works.

If that's the case, it's probably worth making this clear in the `README` file, so that the user is as explicit as possible at each expression to prevent this from happening.

chrisr3:
> What I was trying to understand is if a user specifies `excludePackages` equal to `foo`, does this also apply as a class filter (e.g. for classes that contain `foo`) implicitly due to the way this version of classgraph works.

When you exclude package `foo`, you are excluding all classes both in `foo` and in all sub-packages of `foo`. In Ant terms, this would be `foo.**`, I think.

dimosr:
Yeah, was mostly worrying about classes outside this package (and sub-packages) that could be inadvertently excluded. Altered slightly your tests locally and performed a sanity check confirming that's not the case.

Issue #245, opened 2019-07-11
Add A Group For The Corda Tasks


opticyclic:
All the Corda tasks appear in the `other` group when you run `./gradlew tasks` (or look in the IDEA Gradle tool window).

Add a group to the definition so they are grouped together for clarity.

https://guides.gradle.org/writing-gradle-tasks/#add_a_task_description

chrisr3:
Fixed in 5.0.7 by #268

PR #268, opened 2019-11-22
Set groups and descriptions for our Gradle tasks.

chrisr3:
Fix issue #245  - Cordformation, API Scanner and Jar Filter.


cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -34,6 +34,11 @@ open class Baseform(objects: ObjectFactory) : DefaultTask() {
 
     private companion object {
         const val nodeJarName = "corda.jar"
+        const val GROUP_NAME = "Cordformation"
```
opticyclic: Why have you added a different group name inline here?

chrisr3: Different to what, exactly?

PR #277, opened 2020-01-28
improved runnodes on LINUX and leveraged rlwrap on Linux and OSX to enable shell history

woggioni:
runnodes script now looks at CORDA_NODE_RUNNER_TERMINAL_COMMAND environmental variable first, to make it possible to
launch a custom terminal emulator (if the aforementioned variable is not set it will fallback to the previous behavior, which was
 to open a new tmux window if run inside tmux or to launch xterm otherwise).

Also if rlwrap is available, it will try, on both Linux and OSX, to wrap crash inside it to enable persistent shell history.

### DIFF
--- a/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
+++ b/cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
@@ -1,7 +1,10 @@
 package net.corda.plugins
 
 import java.awt.GraphicsEnvironment
+import java.io.BufferedWriter
 import java.io.File
+import java.io.FileOutputStream
+import java.io.OutputStreamWriter
 import java.util.*
 
 private const val HEADLESS_FLAG = "--headless"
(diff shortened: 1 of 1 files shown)


cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -102,27 +105,60 @@ private class NodeRunner(baseDebugPort: Int, baseMonitoringPort: Int) {
...
-                } else {
-                    listOf("xterm", "-T", nodeName, "-e", unixCommand)
-                }
+                val tmpFile = createTempScriptFile()
+                val shellScript = """
+                    NODE_NAME='$nodeName'
+                    COMMAND="$tmpFile"
+                    if [ -n ${'$'}{CORDA_NODE_RUNNER_TERMINAL_COMMAND+x} ]
+                    then
+                        eval ${'$'}CORDA_NODE_RUNNER_TERMINAL_COMMAND
```
woggioni: I forgot to mention that it looks for the `$NODE_NAME` and `$COMMAND` placeholders in the value of the environmental variable `CORDA_NODE_RUNNER_TERMINAL_COMMAND`, e.g. I use `tilix` as a terminal emulator and I have
```
[woggioni@19DUB-DEL05 ~]$ echo $CORDA_NODE_RUNNER_TERMINAL_COMMAND
tilix -t "$NODE_NAME" -e "$COMMAND"
```

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -102,27 +105,60 @@ private class NodeRunner(baseDebugPort: Int, baseMonitoringPort: Int) {
...
+            val tmpFile: File = File.createTempFile("corda_node_script_launcher", "")
+            BufferedWriter(OutputStreamWriter(FileOutputStream(tmpFile))).use {
+                //Start shell to keep window open unless java terminated normally or due to SIGTERM:
+                it.write("""
+                        #!/bin/sh
+                        set -v
+                        unlink $tmpFile
+                        cd "$workingDir"
+                        command -v rlwrap
+                        if [ -${'$'}? -eq 0 ]
```
woggioni: the `-` before the dollar sign is actually a typo, it doesn't cause any harm because `0` is `-eq` to `-0`, which makes sense, try to run
```
if [ -0 -eq 0 ]; then echo success; else echo fail; fi
```
to validate my claim..
btw I'm removing the dash in a following revision

woggioni: it could actually cause trouble in case `command` exited with a negative return value but that seems to be impossible
https://unix.stackexchange.com/questions/418784/what-is-the-min-and-max-values-of-exit-codes-in-linux

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -102,27 +105,60 @@ private class NodeRunner(baseDebugPort: Int, baseMonitoringPort: Int) {
     }
 
     private fun startWindowed(command: List<String>, workingDir: File, nodeName: String): Process {
+
+        val createTempScriptFile = {
+            val tmpFile: File = File.createTempFile("corda_node_script_launcher", "")
+            BufferedWriter(OutputStreamWriter(FileOutputStream(tmpFile))).use {
```
chrisr3: Kotlin has its own versions of these functions:
```kotlin
val tmpFile = createTempFile("corda_node_script_launcher", "")
tmpFile.bufferedWriter().use {
   // etc
}
```
Don't forget to remove the unused import statements afterwards ... :wink:

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -102,27 +105,60 @@ private class NodeRunner(baseDebugPort: Int, baseMonitoringPort: Int) {
...
+                //Start shell to keep window open unless java terminated normally or due to SIGTERM:
+                it.write("""
+                        #!/bin/sh
+                        set -v
+                        unlink $tmpFile
+                        cd "$workingDir"
+                        command -v rlwrap
+                        if [ ${'$'}? -eq 0 ]
```
chrisr3: Oook! I think you can jus write `$?` in this _particular_ case because there's no`{` immediately after the `$`.

However, I believe that `$?` is the return code from _the very last_ command executed, which means that **you need to capture its value _immediately_ if you wish to use it again**.
e.g.
```
RC=$?
if [ $RC -eq 0 -o $RC -eq 143 ]
```
I really don't believe that your `$? -eq 0 -o $? -eq 143` lines are doing what you think.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -102,27 +105,60 @@ private class NodeRunner(baseDebugPort: Int, baseMonitoringPort: Int) {
...
+
+        val createTempScriptFile = {
+            val tmpFile: File = File.createTempFile("corda_node_script_launcher", "")
+            BufferedWriter(OutputStreamWriter(FileOutputStream(tmpFile))).use {
+                //Start shell to keep window open unless java terminated normally or due to SIGTERM:
+                it.write("""
+                        #!/bin/sh
+                        set -v
+                        unlink $tmpFile
+                        cd "$workingDir"
```
chrisr3: Yikes! These shenanigans with `workingDir` scare me. There's a standard way for a script to discover its _own_ directory though:
```sh
BASEDIR=$(dirname $0)
```

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
```diff
@@ -102,27 +105,60 @@ private class NodeRunner(baseDebugPort: Int, baseMonitoringPort: Int) {
...
     private fun startWindowed(command: List<String>, workingDir: File, nodeName: String): Process {
+
+        val createTempScriptFile = {
+            val tmpFile: File = File.createTempFile("corda_node_script_launcher", "")
+            BufferedWriter(OutputStreamWriter(FileOutputStream(tmpFile))).use {
+                //Start shell to keep window open unless java terminated normally or due to SIGTERM:
+                it.write("""
+                        #!/bin/sh
+                        set -v
+                        unlink $tmpFile
```
chrisr3: Why does this script want to delete itself?

woggioni: because it a temporary file that needs to be deleted by someone, it cannot be deleted by the Java process using `File#deleteOnExit` because there would be a race condition with the Java process terminating before the shell process is actually fired up (it's a fork so there is no synchronization guarantee)

chrisr3: I think it only "needs" deleting because you keep generating new ones with different values hard-coded into them. Have you tried just creating _one_ script, and passing the values in as parameters?

But this PR has bigger problems than that anyway: The `runnodes` script no longer launches nodes on my machine. Instead, it seems just to dump all of the different `$shellScript` contents to the console and then exit.

cordformation/src/noderunner/kotlin/net/corda/plugins/NodeRunner.kt
chrisr3: OK, this `$?` issue actually looks like a bug in the original script too.

PR #278, opened 2020-01-30
added gradle settings to configure the node's database parameters

woggioni:
this allow to configure the data source parameters in the `node.conf` file in the following way

```
node {
    dataSourceProperties {
        dataSourceClassName "org.postgresql.ds.PGSimpleDataSource"
        url "jdbc:postgresql:corda_bob"
        user "Bob"
        password "mypassword"
    }
}
```

JDBC driver, URL, user and password are supported at the moment

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -248,6 +248,22 @@ open class Node @Inject constructor(private val project: Project) {
         config = config.withValue("h2port", ConfigValueFactory.fromAnyRef(h2Port))
     }
 
+    /**
+     * This section is used to configure the JDBC connection and database driver used for the node’s persistence.
+     */
+    fun dataSourceProperties(options: DataSourceProperties) {
+        config = options.addTo("dataSourceProperties", config)
+    }
+
+    /**
+     * This section is used to configure the JDBC connection and database driver used for the node’s persistence.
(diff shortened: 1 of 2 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -248,6 +248,22 @@ open class Node @Inject constructor(private val project: Project) {
...
+     * This section is used to configure the JDBC connection and database driver used for the node’s persistence.
+     */
+    fun dataSourceProperties(options: DataSourceProperties) {
+        config = options.addTo("dataSourceProperties", config)
+    }
+
+    /**
+     * This section is used to configure the JDBC connection and database driver used for the node’s persistence.
+     */
+    fun dataSourceProperties(configureClosure: Closure<in DataSourceProperties>) {
```
chrisr3: `Closure` is a Groovy-ism. The official Kotlin equivalent these days is more like:
```kotlin
protected val dataSourceProperties: DataSourceProperties = project.objects.newInstance(DataSourceProperties::class.java)

fun dataSourceProperties(action: Action<in DataSourceProperties>) {
    action.execute(dataSourceProperties)
}
```
where `objects` is an instance of Gradle's `ObjectFactory`. Ugh, Getting Gradle DSLs right is _hard work_.

chrisr3:
Closed as abandoned.

PR #289, opened 2020-02-21
replaced `org.apache.commons.io.IOUtils` with `java.nio.file.Files`

woggioni:
`IOUtils.copy` doesn't close the output stream which means that needs to be used inside a **try/finally** block (or **try-with-resource** in Java or **use** block in Kotlin)
The way it was used in the unit test was simply leaking an opened file descriptor with possible unwanted side effects; it actually calls `flush` before exiting, which is the reason why nothing bad happens, but it still doesn't look quite right. In Windows for example you cannot delete a file while it is opened by another process, which is something that could happen here since the unit tests create temporary directories that are deleted afterwards, and there is no synchronization between the leaked `FileOutputStream`s closing (which at this point depends entirely on the garbage collector's mood) and the temporary directory deletion (which happens at the end of every single test case)

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -2,7 +2,6 @@ package net.corda.plugins
 
 import com.typesafe.config.*
 import groovy.lang.Closure
-import org.apache.commons.io.FilenameUtils.removeExtension
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ProjectDependency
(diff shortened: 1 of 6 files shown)


woggioni:
note that there is a subtle difference in the fact that `Files.copy` fails in case the destination file already exists while opening a new `FileOutputStream` over an already existing files simply ends up in overwriting, this difference could simply be removed passing `StandardCopyOption.REPLACE_EXISTING` to `Files.copy` but I deliberately decided not to do it
because in the unit test the files are always copied inside a temporary folder that is created on the fly for every single test case (which means that overwriting should never happen)

woggioni:
oh, yeah.. I'll do that

woggioni:
Turns out the only remaining reason we were depending on `commons-io:commons-io` was to parse file extension from filename, which, in my opinion, is overkill. I just implemented that function in 4 lines and got rid of the dependency completely

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -541,7 +540,14 @@ open class Node @Inject constructor(private val project: Project) {
         Files.createDirectories(configDir)
         for ((jarFile, config) in cordapps) {
             if (config == null) continue
-            val configFile = configDir.resolve("${removeExtension(jarFile.fileName.toString())}.conf")
+            val fileNameWithoutExtension = jarFile.fileName.toString().let {
+                val end = when(val dotIndex = it.lastIndexOf('.')) {
```
chrisr3: Maybe better?
```kotlin
when(val dotIndex = it.lastIndexOf('.')) {
    -1 -> it
    else -> it.substring(0, dotIndex)
}
```
You only need to invoke `substring()` if the filename has an extension.

woggioni: makes total sense...

PR #292, opened 2020-02-28
Use screen on macOS when the user is already running screen

josephzunigadaly:
This allows runnodes to use screen as an alternative to osascript on macOS. screen is included in macOS making it an easy alternative to use.

The change behaves the same as tmux support on Linux. If the user is already running screen, it will be used; otherwise it will run osascript like normal.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -69,6 +69,7 @@ changes to this list.
 * Jonathan Sartin (R3)
 * Jose Coll (R3)
 * Jose Luu (Natixis)
+* Joseph Zuniga-Daly (R3)
 * Josh Lindl (BCS)
 * Justin Chapman (Northern Trust)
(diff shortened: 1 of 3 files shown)


josephzunigadaly:
Thank you. I have now updated the changelog and contributors.

josephzunigadaly:
> OK, one more thing: Does this change have a JIRA number too please? I just need to know what it is, assuming there is one.

The closest thing is issue 181 on this project although they suggested different approaches.

mnesbit:
If there isn't a linked CORDA JIRA number we should create one. All PR's pretty much MUST have a JIRA at the start of the title to help out release managers. Ask Sachin to check your JIRA is in the right buckets

josephzunigadaly:
I have raised CORDA-3661 and sent to Sachin.

PR #294, opened 2020-03-19
EG-848: Extend the Dockerform plugin to default to Postgres database

palade:
Extended Dockerform plugin such that each Corda node which is deployed as a Docker container connects to its own Postgres database which also deployed as a Docker container. The plugin deals with the fact that database is not deployed during bootstrap. The plugin produces a docker-compose file.


cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -1,14 +1,28 @@
...
 
 import org.assertj.core.api.Assertions.assertThat
+import org.gradle.internal.impldep.org.junit.Ignore
 import org.gradle.testkit.runner.TaskOutcome
 import org.junit.jupiter.api.Test
 
 class DockerformTest : BaseformTest() {
 
+
+//    @Test
```
JamesHR3: Let's not leave commented out tests. Either this should work and it should be uncommented (and potentially fixed) or it's an invalid test and it should be removed.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +90,42 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/network-parameters:/opt/corda/network-parameters",
                             "$nodeBuildDir/additional-node-infos:/opt/corda/additional-node-infos",
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
+                    ),
+                    "ports" to listOf(it.rpcPort, it.config.getInt("sshd.port"), it.p2pPort),
```
JamesHR3: Why do we need to add the p2p port to this list?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -497,6 +508,20 @@ open class Node @Inject constructor(private val project: Project) {
         }
     }
 
+    internal fun installPostgresDriver() {
+
+        val driverJar = project.configuration("cordapp").files {
+            (it.group == "org.postgresql") &&
+                    (it.name == "postgresql") &&
+                    (it.version == "42.2.10")
```
JamesHR3: This is the sort of case where having the version number as a constant would be valuable.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -567,19 +598,61 @@ open class Node @Inject constructor(private val project: Project) {
         if (!config.hasPath("sshd.port")) {
             sshdPort(defaultSsh)
         }
-        val configDefaults = ConfigFactory.empty()
-            .withValue("dataSourceProperties.dataSource.url", ConfigValueFactory.fromAnyRef("jdbc:h2:file:./persistence/persistence;DB_CLOSE_ON_EXIT=FALSE;WRITE_DELAY=0;LOCK_TIMEOUT=10000"))
+
+        // TODO Refactor this code to extract the information from the config file
```
JamesHR3: Let's avoid leaving TODOs in the code. If it can be fixed now then let's do so. If it's a serious piece of work that needs a follow up task, then we'll have to raise a JIRA and scope it in.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -56,14 +57,24 @@ open class Node @Inject constructor(private val project: Project) {
         private set
     private var rpcSettings: RpcSettings = RpcSettings()
     private var webserverJar: String? = null
-    private var p2pPort = 10002
+    internal var p2pPort = 10002
```
JamesHR3: Probably want this as `private set` as well (basically saying that only this class is allowed to change the value).

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -56,14 +57,24 @@ open class Node @Inject constructor(private val project: Project) {
...
+        @Input get
     internal var rpcPort = 10003
         @Input get
         private set
     internal var config = ConfigFactory.empty()
         @Internal get
         private set
 
+    lateinit var dbSettings: DbSettings
```
JamesHR3: This should also be `private set` I think. Does it need to be `lateinit`, or do we know what the `DbSettings` are going to be when the class is created?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -567,19 +598,61 @@ open class Node @Inject constructor(private val project: Project) {
         if (!config.hasPath("sshd.port")) {
             sshdPort(defaultSsh)
         }
-        val configDefaults = ConfigFactory.empty()
-            .withValue("dataSourceProperties.dataSource.url", ConfigValueFactory.fromAnyRef("jdbc:h2:file:./persistence/persistence;DB_CLOSE_ON_EXIT=FALSE;WRITE_DELAY=0;LOCK_TIMEOUT=10000"))
+
+        // TODO Refactor this code to extract the information from the config file
+        val p2pHost = config.getString("p2pAddress")
```
JamesHR3: I'm a bit confused with what's going on in this line and below to line 608. I think this should just work the way it did before, so why do you need to do this extra address parsing? If this is required, it should probably be factored out as you're effectively doing the same thing three times (finding out what the port number is from the address string).

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -567,19 +598,61 @@ open class Node @Inject constructor(private val project: Project) {
...
-            .withValue("p2pAddress", ConfigValueFactory.fromAnyRef("$containerName:$p2pPort"))
-            .withValue("rpcSettings.address", ConfigValueFactory.fromAnyRef("$containerName:${rpcSettings.port}"))
-            .withValue("rpcSettings.adminAddress", ConfigValueFactory.fromAnyRef("$containerName:${rpcSettings.adminPort}"))
-            .withValue("detectPublicIp", ConfigValueFactory.fromAnyRef(false))
-            .withFallback(configDefaults)
+                .withValue("dataSourceProperties.dataSource.url", ConfigValueFactory.fromAnyRef("jdbc:h2:file:./persistence/persistence;DB_CLOSE_ON_EXIT=FALSE;WRITE_DELAY=0;LOCK_TIMEOUT=10000"))
+                .withValue("p2pAddress", ConfigValueFactory.fromAnyRef("$containerName:$p2pPort"))
+                .withValue("rpcSettings.address", ConfigValueFactory.fromAnyRef("$containerName:$rpcPort"))
+                .withValue("rpcSettings.adminAddress", ConfigValueFactory.fromAnyRef("$containerName:$rpcAdminPort"))
+                .withValue("detectPublicIp", ConfigValueFactory.fromAnyRef(false))
```
JamesHR3: We've lost the fallback to config defaults here, which might be important.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -567,19 +598,61 @@ open class Node @Inject constructor(private val project: Project) {
...
+        val rpcAdminConfig = config.getString("rpcSettings.adminAddress")
+        val rpcAdminPort = rpcAdminConfig.substring(rpcConfig.indexOf(":") + 1)
+
         val dockerConf = config
-            .withValue("p2pAddress", ConfigValueFactory.fromAnyRef("$containerName:$p2pPort"))
-            .withValue("rpcSettings.address", ConfigValueFactory.fromAnyRef("$containerName:${rpcSettings.port}"))
-            .withValue("rpcSettings.adminAddress", ConfigValueFactory.fromAnyRef("$containerName:${rpcSettings.adminPort}"))
-            .withValue("detectPublicIp", ConfigValueFactory.fromAnyRef(false))
-            .withFallback(configDefaults)
+                .withValue("dataSourceProperties.dataSource.url", ConfigValueFactory.fromAnyRef("jdbc:h2:file:./persistence/persistence;DB_CLOSE_ON_EXIT=FALSE;WRITE_DELAY=0;LOCK_TIMEOUT=10000"))
```
JamesHR3: This is the H2 JDBC URL, so presumably this is the config you use during the bootstrap phase, which is then overridden?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -567,19 +598,61 @@ open class Node @Inject constructor(private val project: Project) {
...
+        setDBSettings(port, address)
+
+        val dockerConf = config
+                .withValue("dataSourceProperties.dataSourceClassName", ConfigValueFactory.fromAnyRef("org.postgresql.ds.PGSimpleDataSource"))
+                .withValue("dataSourceProperties.dataSource.url", ConfigValueFactory.fromAnyRef("jdbc:postgresql://$address:$port/mydb?currentSchema=myschema"))
+                .withValue("dataSourceProperties.dataSource.user", ConfigValueFactory.fromAnyRef("myuser"))
+                .withValue("dataSourceProperties.dataSource.password", ConfigValueFactory.fromAnyRef("mypassword"))
+                .withValue("database.transactionIsolationLevel", ConfigValueFactory.fromAnyRef("READ_COMMITTED"))
+                .withValue("database.runMigration", ConfigValueFactory.fromAnyRef(true))
+                .withValue("database.schema", ConfigValueFactory.fromAnyRef("myschema"))
```
JamesHR3: Again there's no fallback to defaults here, which means anything you don't specify will be unset.

cordformation/src/test/resources/net/corda/plugins/DeployCordappWithDocker.gradle
```diff
@@ -0,0 +1,81 @@
...
+        rpcPort 10003
+        rpcSettings {
+            adminAddress "localhost:10004"
+        }
+        cordapps = ["$corda_group:corda-finance-contracts:$corda_release_version",
+                    "$corda_group:corda-finance-workflows:$corda_release_version"]
+    }
+}
+
+task prepareDockerNodes(type: net.corda.plugins.Dockerform, dependsOn: ['jar']) {
```
JamesHR3: One way around the OS/ENT harmonization issue is to potentially add a bit of configuration to this which is the `externalDatabasePort`. If it's set, use it as the port to a postgres instance, otherwise fall back to H2. We'd then have to document clearly that it won't work in OS because external DBs aren't supported there (again, assuming no harmonization). It might be worth doing this anyway as a work around for port clashes.

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
palade: Test has now been commented out

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -72,6 +72,9 @@ class Cordformation : Plugin<Project> {
         val jolokia = project.dependencies.add(CORDFORMATION_TYPE, "org.jolokia:jolokia-jvm:$jolokiaVersion:agent")
         // The Jolokia agent is a fat jar really, so we don't want its transitive dependencies.
         (jolokia as ModuleDependency).isTransitive = false
+
+        val postgresDriver = project.dependencies.add("cordapp","org.postgresql:postgresql:42.2.10")
```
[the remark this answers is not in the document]
palade: Resolved

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: This code was added to read the p2p/rpc/adminrpc port numbers from the build.gradle. Before it was working fine as long as there was only one Corda node container in the network. This was because the rpc and p2p port numbers from the build.gradle file were ignored, and hardcoded values were used instead. Using the same port number by more than one service in the same network was triggering a port already allocated error, and the Corda node container would shut down.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -72,6 +72,10 @@ class Cordformation : Plugin<Project> {
         val jolokia = project.dependencies.add(CORDFORMATION_TYPE, "org.jolokia:jolokia-jvm:$jolokiaVersion:agent")
         // The Jolokia agent is a fat jar really, so we don't want its transitive dependencies.
         (jolokia as ModuleDependency).isTransitive = false
+
+        val postgresDriverVersion = project.findRootProperty("postgres_driver_version") ?: "42.2.10"
+        val postgresDriver = project.dependencies.add("cordapp","org.postgresql:postgresql:$postgresDriverVersion")
```
chrisr3: Yikes! Why is Cordformation adding extra dependencies to the Gradle project? I wouldn't expect that!

(Yes, I see Jolokia does this too. _**UGH!**_  OTOH, at least no-one _sees_ Jolokia and CorDapp developers hopefully don't know it even exists...)

This is a Gradle plugin, so think about what your `build.gradle` script is going to look like as you try to use this new feature. And _please_ be careful to keep backwards compatibility with older versions of the plugins so that people can upgrade their plugins without breaking their projects.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +90,42 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                    "image" to "entdocker.software.r3.com/corda-enterprise-java1.8-${it.runtimeVersion().toLowerCase()}",
```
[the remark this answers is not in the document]
chrisr3: Agreed - don't hard-code configuration into the plugins, especially if they're very specific to a particular use-case. Best to ask the CorDapp developers themselves for values like this.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
chrisr3: The `Node` object is exposed into people's `build.gradle` files and can be accessed via Groovy, so "yes" - "Lock it down!"

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -497,6 +508,22 @@ open class Node @Inject constructor(private val project: Project) {
         }
     }
 
+    internal fun installPostgresDriver() {
```
chrisr3: I don't understand why PostgreSQL is "special". What about Oracle? And MySQL? Does _everyone_ get built-in support like this?

Supporting PostgreSQL is fine and presumably desirable. But I think the plugin should be allowing people to "Put a JDBC driver of their own choosing in the correct place". We _do not_ want to need to release a new version of the Gradle plugins just to upgrade people's versions of PostgreSQL.

chrisr3:
This is tricky - our Gradle plugins are _public_ and are actively used by CorDapp developers. We must therefore be _very_ careful about how we change them. As I understand this task, we are looking to make it easier for people to use non-H2 databases. The primary use-case here is for `Dockerform`, but `Dockerform` shares the `Node` object with `Cordform` and so we are likely talking about changing both. (Besides, does it make sense for _only_ `Dockerform` to support this?)

So first of all, ask yourself:
- "_Why_ is using a non H2 database difficult?", followed by
- "How can we make it easier _without breaking existing users?_"

People may upgrade their Corda Gradle plugins for any reason, and they don't expect their builds to break when they do.

Then look at your `Dockerform` / `Cordform` block in your `build.gradle` file and ask yourself what you would _exepct_ to need to add here. Forget the plugin code at this point - just try to understand how you would expect to change your `build.gradle` and how you would expect Gradle to behave if you did.

Also appreciate that people using H2 are unlikely to appreciate suddenly having to download a PostgreSQL driver from "somewhere" just because they've upgraded `cordformation`. And people who _do_ want to use PostgreSQL might have a very _specific_ idea of which version of the driver they want to use and where they want to download it from anyway, so both of these need to be configurable. We don't want to be in a position where we need to release a new version of `cordformation` just because we want to upgrade the JDBC driver.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: Removed

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: These are removed because are not fallback configuration anymore. We are using the h2 config to be able to run the network bootstrap procedure, then we append the default Postgres database config

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: There is no need for the fallback because all those properties are specified. Even if they are specified they will be overwritten

cordformation/src/test/resources/net/corda/plugins/DeployCordappWithDocker.gradle
palade: @JamesHR3 If an external Postgres instance exists, then why not use the Corform instead of Dockerform?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: yes

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordappWithDocker.gradle
```diff
@@ -0,0 +1,53 @@
...
+        rpcSettings {
+            adminAddress "localhost:10004"
+        }
+        cordapps = ["$corda_group:corda-finance-contracts:$corda_release_version",
+                    "$corda_group:corda-finance-workflows:$corda_release_version"]
+    }
+}
+
+
+task prepareDockerNodes(type: net.corda.plugins.Dockerform) {
```
JamesHR3: Doesn't this version require you to specify what version of the postgres driver you want to get the databases? I'd have thought there'd be some changes to this task (and the ones below).

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -69,12 +97,44 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+
+            if (it.drivers.isNullOrEmpty() && project.hasProperty(DEFAULT_DB_GRADLE_PROPERTY)) {
+
+                it.dockerImage("entdocker.software.r3.com/corda-enterprise-java1.8-${it.runtimeVersion().toLowerCase()}")
+                it.dbSettings(DEFAULT_DB_STARTING_PORT + index, "${it.containerName}-db")
+                it.installDefaultDatabaseConfig(
+                        createDefaultDBConfig(it.dbSettings.port, it.dbSettings.host), DEFAULT_DB_INIT_FILE
+                )
+                val driverVersion = project.property(DEFAULT_DB_GRADLE_PROPERTY)
+                it.installDriver("org.postgresql", "postgresql", driverVersion)
```
JamesHR3: Where does this get the postgres driver from? Do we specify a repository, do we ask the user for one, or do we rely on the `build.gradle` repositories block?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -497,6 +527,17 @@ open class Node @Inject constructor(private val project: Project) {
         }
     }
 
+    internal fun installDriver(driverGroup: String, driverName: String, driverVersion: Any?) {
```
JamesHR3: There's also `installDrivers` below that seems to do something very similar. Can we reuse that by appending the driver to the drivers configuration instead?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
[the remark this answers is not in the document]
palade: I now addressed this by providing the option to specify the dockerImage in the gradle file

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
palade: I removed this code from Cordformation

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordappWithDocker.gradle
palade: Yes you need to add the postgres dependency version in the build.gradle file. Also you can specify the dockerImage in the same file. I will update the design document to reflect these changes

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
palade: We rely on the repositories specified in the build.gradle block

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: Removed now

palade:
> This is tricky - our Gradle plugins are _public_ and are actively used by CorDapp developers. We must therefore be _very_ careful about how we change them. As I understand this task, we are looking to make it easier for people to use non-H2 databases. The primary use-case here is for `Dockerform`, but `Dockerform` shares the `Node` object with `Cordform` and so we are likely talking about changing both. (Besides, does it make sense for _only_ `Dockerform` to support this?)
> 
> So first of all, ask yourself:
> 
> * "_Why_ is using a non H2 database difficult?", followed by
> * "How can we make it easier _without breaking existing users?_"
> 
> People may upgrade their Corda Gradle plugins for any reason, and they don't expect their builds to break when they do.
> 
> Then look at your `Dockerform` / `Cordform` block in your `build.gradle` file and ask yourself what you would _exepct_ to need to add here. Forget the plugin code at this point - just try to understand how you would expect to change your `build.gradle` and how you would expect Gradle to behave if you did.
> 
> Also understand that people using H2 are unlikely to appreciate suddenly having to download a PostgreSQL driver from "somewhere" just because they've upgraded `cordformation`. And people who _do_ want to use PostgreSQL might have a very _specific_ idea of which version of the driver they want to use and where they want to download it from anyway, so both of these need to be configurable. We don't want to be in a position where we need to release a new version of `cordformation` just because we want to upgrade the JDBC driver.

@chrisr3 Using a non-H2 database I think is more difficult because it requires the database to be deployed before using Dockerform/Cordform plugins and it would require that the developer to either manually supply the database configuration in the build.gradle file or edit each node.conf with the correct configuration. To make it easier for developers to quickly test their cordapp against a non-H2 database, Dockerform can be used. This creates a docker-compose file which can be used to control de deployment of Corda nods and databases by running a single command, whereas in Cordform you need to this manually one-by-one. However, the existing Dockerform supports only H2, and the purpose of this Dockerform extension was to use Postgres. The last commits have addressed the issue of not breaking previous builds. A user has to specify a postgres driver version property in their build.gradle file, and then the Postgres database will be installed. The user can also specify his docker image (of a another corda version) if he chooses to do so. He is not restricted to some pre-defined one.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
chrisr3: Yes, CorDapp developer will expect to use the `repositories` block in `build.gradle` to tell Gradle where to download everything from. I think they'll also expect to specify more than just its version though.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -33,6 +38,10 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
         }
 
         private val YAML_MAPPER = Yaml(YAML_FORMAT_OPTIONS)
+
+        private const val DEFAULT_DB_INIT_FILE = "init.sql";
```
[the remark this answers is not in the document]
palade: removed 😁

cordformation/src/main/kotlin/net/corda/plugins/DbSettings.kt
```diff
@@ -0,0 +1,3 @@
+package net.corda.plugins
+
+data class DbSettings(val port: Int, val host: String)
```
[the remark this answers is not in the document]
palade: This class was removed

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -57,13 +59,23 @@ open class Node @Inject constructor(private val project: Project) {
...
     private var p2pPort = 10002
+        @Internal get
     internal var rpcPort = 10003
         @Input get
         private set
     internal var config = ConfigFactory.empty()
         @Internal get
         private set
 
+    var dbSettings: DbSettings = DbSettings(5432, DEFAULT_HOST)
```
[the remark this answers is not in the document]
palade: This was removed

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -57,13 +59,23 @@ open class Node @Inject constructor(private val project: Project) {
...
+
+    /**
+     * Sets the database configuration
+     */
+    fun dbSettings(port: Int, address: String) {
```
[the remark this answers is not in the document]
palade: Removed now

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +113,45 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+
+                val driverVersion = project.property(DEFAULT_DB_GRADLE_PROPERTY)
+                val driverJar = retrieveDefaultDbDriverDependency(
+                        "org.postgresql", "postgresql", driverVersion)
+                if (driverJar != null) {
+                    it.drivers = listOf(driverJar.path)
+                    it.installDrivers()
+                }
+
+                it.dbSettings(DEFAULT_DB_STARTING_PORT + index, "${it.containerName}-db")
```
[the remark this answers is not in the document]
palade: Removed now

cordformation/src/main/resources/net/corda/plugins/init.sql
```diff
@@ -0,0 +1,7 @@
+CREATE SCHEMA "myschema";
```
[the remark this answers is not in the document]
palade: Removed the sql file and added a Dockerfile to create a wrapper around the base Postgres image that is initialised with the default or user specified values

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -54,6 +62,26 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
             return wantedPath
         }
 
+    private fun retrieveDefaultDbDriverDependency(driverGroup: String, driverName: String, driverVersion: Any?): File? {
+        project.configuration("default").incoming.beforeResolve {
```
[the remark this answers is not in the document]
palade: Added a cordaDriver configuration as described here

palade:
> I've focussed largely on the "Gradle Mechanics" here, i.e. its API usage. I definitely believe that a new `cordaDriver` configuration would be a cleaner way for developers to declare _any_ Maven artifacts that should be copied into nodes' `drivers/` directory, for example.
> 
> I know that _our_ use-case involves PostgreSQL, but the plugins are tools for everyone and should therefore strive to remain agnostic.
> 
> But still, there's a lot of good stuff here too 😸.

Added cordaDriver configuration to default to Postgres database. Replace the init.sql file with a Dockerfile, and refactored the Dockerform to easily add support for other databases in the future.

cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -66,6 +66,7 @@ class Cordformation : Plugin<Project> {
             createCompileConfiguration("cordapp", this)
             val cordaRuntime = createRuntimeConfiguration("cordaRuntime", this)
             createChildConfiguration(CORDFORMATION_TYPE, cordaRuntime, this)
+            createRuntimeConfiguration("cordaDriver", this)
```
chrisr3, line 69: I don't think you want to be using `createRuntimeConfiguration` here because it's equivalent to:
```gradle
configurations {
    runtime.extendsFrom cordaDriver
}
```
i.e. anything added to the `cordaDriver` configuration will automatically be added to the `runtime` configuration too :scream_cat: !

palade: Code changed to address this comment

chrisr3: Eh? I wasn't aware that I'd submitted that review yet... :confused:

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -559,27 +578,58 @@ open class Node @Inject constructor(private val project: Project) {
...
+
+        val options = ConfigRenderOptions
+                .defaults()
+                .setOriginComments(false)
+                .setComments(false)
+                .setFormatted(true)
+                .setJson(false)
+        val configFileText = createNodeConfig(config).root().render(options).split("\n").toList()
+        val nodeConfFile = File(nodeDir, "node.conf")
+        Files.write(nodeConfFile.toPath(), configFileText, StandardCharsets.UTF_8)
```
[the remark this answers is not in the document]
palade: removed

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -559,27 +578,58 @@ open class Node @Inject constructor(private val project: Project) {
         createNodeAndWebServerConfigFiles(config)
     }
 
+    fun installResource(resourceName: String) {
```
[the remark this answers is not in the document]
palade: Added standard copy option

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +116,61 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                    it.drivers = listOf(driverJar.path)
+                    it.installDrivers()
+                }
+
+                it.installResource(DEFAULT_DB_INIT_FILE)
+                it.installResource(DEFAULT_DB_DOCKERFILE)
+
+                val dbPort = DEFAULT_DB_STARTING_PORT + index
+                val dbHost = "${it.containerName}-db"
+                val dbUrl = createDefaultDbURL(DEFAULT_DB, DEFAULT_DB_DRIVER_PATH, dbHost, dbPort, DEFAULT_DB_NAME, DEFAULT_DB_SCHEMA)
```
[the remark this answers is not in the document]
palade: Fixed

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +99,101 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                val dbRunMigration = when {
+                    dockerConfig.hasPath("database.runMigration") -> dockerConfig.getBoolean("database.runMigration")
+                    else -> DEFAULT_DB_RUN_MIGRATION
+                }
+
+                val dbPort = DEFAULT_DB_STARTING_PORT + index
+                val dbHost = "${it.containerName}-db"
+
+                val dbUrl = when {
+                    dockerConfig.hasPath("dataSourceProperties.dataSource.url") -> {
```
chrisr3, line 152: There _is_ actually a way to persuade TypeSafe Config to perform these substitutions, but it requires the `DBURL` value to be (err) _"reformatted"_ first... :scream_cat: 

Ideally, we'd like to do something like this:
```kotlin
val fallback = ConfigFactory.empty()
    .withValue("DBHOSTNAME", ConfigValueFactory.fromAnyRef(dbHost))
    .withValue("DBPORT", ConfigValueFactory.fromAnyRef(dbPort))
    .withValue("DBNAME", ConfigValueFactory.fromAnyRef(dbName))
    .withValue("DBSCHEMA", ConfigValueFactory.fromAnyRef(dbSchema))

// val dbUrl = "jdbc:postgresql://\${DBHOSTNAME}:\${DBPORT}/\${DBNAME}?currentSchema=\${DBSCHEMA}"
val config = ConfigFactory.empty()
    .withValue("DBURL", ConfigValueFactory.fromAnyRef(dbUrl))
    .withFallback(fallback)
    .resolve()
```
However, Typesafe doesn't resolve the `${DBNAME}` etc values for us here because `ConfigValueFactory.fromAnyRef()` adds `DBURL` as a `ConfigString$Quoted` object, i.e. a _literal_ string value :facepalm:. The only way around this  (apparently?!)  is to _parse_ the `DBURL` value instead:
```kotlin
val config = ConfigFactory.parseReader(StringReader("DBURL=$dbUrl"))
    .withFallback(fallback)
    .resolve()
```
However, the parser only accepts valid HCON documents, which means that would then need to wrap the literal elements of `dbUrl` in quotes to escape any ":" or "&" characters (as you already discovered).

I think this should be doable...

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +99,101 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                        var url = dockerConfig.getString("dataSourceProperties.dataSource.url")
+                        url.replace("\${DBHOSTNAME}", dbHost)
```
chrisr3, line 154: The problem now is that we are declaring that we support _exactly_ these four tokens for expansion.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -96,6 +99,10 @@ open class Node @Inject constructor(private val project: Project) {
     @get:Input
     var extraConfig: Map<String, Any> = emptyMap()
 
+    @get:Optional
+    @get:Input
+    var dockerConfig: Map<String, Any> = emptyMap()
```
chrisr3, line 104: `Node` is shared between `Cordform` and `Dockerform`. Does the `dockerConfig` _need_ to be part of `Node`? Could you put it on `Dockerform` and then somehow update the `extraConfig` with the contents of `dockerConfig` at an appropriate time?

cordformation/src/main/resources/net/corda/plugins/Postgres_Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Derive from postgres image
```
chrisr3, line 1: Is this file obsolete?

cordformation/src/main/resources/net/corda/plugins/Postgres_init.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
chrisr3, line 1: Is this file also obsolete?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -559,27 +589,56 @@ open class Node @Inject constructor(private val project: Project) {
         createNodeAndWebServerConfigFiles(config)
     }
 
+    fun installResource(resourceName: String) {
```
chrisr3, line 592: Are we still using this function?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -128,6 +135,24 @@ open class Node @Inject constructor(private val project: Project) {
             }
         }
 
+    /**
```
chrisr3, line 138: Similarly, are we supporting each node having its own `dockerImage`? Or could this property also be moved to `Dockerform` so that all nodes are forced to share the same one?

chrisr3:
For reference, I would _prefer_ that Typesafe Config performs the substitutions in `DBURL` because this allows developers to define their own extra substitution variables beyond the ones we actually need to define ourselves. (I suspect that `DBSCHEMA` is probably something specific to PostgreSQL again...)

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -33,6 +37,17 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+        private const val DEFAULT_DB_DOCKERFILE = "Postgres_Dockerfile"
+        private const val DEFAULT_DB_STARTING_PORT = 5432
+        private const val DEFAULT_DB_USER = "myuser"
+        private const val DEFAULT_DB_PASSWORD = "mypassword"
+        private const val DEFAULT_DB_SCHEMA = "myschema"
+        private const val DEFAULT_DB_NAME = "mydb"
+        private const val DEFAULT_DB_TRANSACTION_ISOLATION_LEVEL = "READ_COMMITTED"
+        private const val DEFAULT_DB_DATA_SOURCE_CLASS_NAME = "org.postgresql.ds.PGSimpleDataSource"
+        private const val DEFAULT_DB_RUN_MIGRATION = true
     }
```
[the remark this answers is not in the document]
palade: removed now

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +99,101 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
             )
-        }.toMap()
 
+            if (it.dockerConfig.isNotEmpty()) {
+
+                var dockerConfig = ConfigFactory.parseMap(it.dockerConfig)
+
+                val dockerfile = when {
+                    dockerConfig.hasPath("dbDockerConfig.dockerfile") -> dockerConfig.getString("dbDockerConfig.dockerfile")
+                    else -> DEFAULT_DB_DOCKERFILE
```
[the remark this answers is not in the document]
palade: removed now

cordformation/src/main/resources/net/corda/plugins/Postgres_Dockerfile
palade: Removed now

cordformation/src/main/resources/net/corda/plugins/Postgres_init.sh
palade: Removed now

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: removed

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: At the moment even though every node has a dockerimage property, all use the same image. Moving this into Dockerform for now

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: Moved into the Dockerform to be added at a later time

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -63,18 +75,18 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
         initializeConfiguration()
         nodes.forEach { it -> it.installDockerConfig(DEFAULT_SSH_PORT) }
```
chrisr3, line 76: Seeing as we're here, the `it ->` part of this lambda is not needed because `it` is the name of the default lambda parameter. So
```kotlin
nodes.forEach { it.installDockerConfig(DEFAULT_SSH_PORT) }
```
is technically better.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -84,13 +96,79 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
                     ),
                     "ports" to listOf(it.rpcPort, it.config.getInt("sshd.port")),
-                    "image" to "corda/corda-zulu-${it.runtimeVersion().toLowerCase()}"
+                    "image" to (dockerImage ?: "corda/corda-zulu-${it.runtimeVersion().toLowerCase()}")
             )
-        }.toMap()
 
+            if (dockerConfig.isNotEmpty()) {
+
+                var dockerConfig = ConfigFactory.parseMap(dockerConfig)
```
chrisr3, line 106: We never change the value of `dockerConfig` once it's been assigned, so
```kotlin
val dockerConfig = ConfigFactory.parseMap(dockerConfig)
```
is better.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -100,5 +178,4 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
 
         Files.write(dockerComposePath, dockerComposeContent.toByteArray(StandardCharsets.UTF_8))
```
chrisr3, line 179: Kotlin's default `Charset` is already `UTF-8` and so you don't need to set this here.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -562,24 +567,47 @@ open class Node @Inject constructor(private val project: Project) {
...
+    private fun updateNodeConfigFile(config: Config) {
+        val options = ConfigRenderOptions
+                .defaults()
+                .setOriginComments(false)
+                .setComments(false)
+                .setFormatted(true)
+                .setJson(false)
+        val configFileText = createNodeConfig(config).root().render(options).split("\n").toList()
+        val nodeConfFile = File(nodeDir, "node.conf")
+        Files.write(nodeConfFile.toPath(), configFileText)
```
chrisr3, line 608: Do we need to specify `REPLACE_EXISTING` here as an option? We _are_ apparently _updating_ the configuration.

cordformation/src/test/resources/postgres.gradle
```diff
@@ -0,0 +1,120 @@
...
+    dbHostName             = 'localhost'
+    dbInit                 = 'Postgres_init.sh'
+    dbDockerfile           = 'Postgres_Dockerfile'
+    postgres = [
+            dataSourceProperties: [
+                    dataSourceClassName: 'org.postgresql.ds.PGSimpleDataSource',
+                    dataSource: [
+                            user    : dbUser,
+                            password: dbPassword,
+                            url     : "jdbc\":\"postgresql\":\"\"//\"\${DBHOSTNAME}\":\"\${DBPORT}/\${DBNAME}\"?\"currentSchema\"=\"\${DBSCHEMA}",
```
chrisr3, line 18: I think this example JDBC URL template can be simplified slightly:
```
url : "\"jdbc:postgresql://\"\${DBHOSTNAME}\":\"\${DBPORT}\"/\"\${DBNAME}\"?currentSchema=\"\${DBSCHEMA}",
```
The "trick" is that Typesafe interprets quoted text literally, and so we only need to put quotes around the prefix text, the suffix text and the text between the substitutions. And the quotes must also balance, of course.

I am slightly concerned that we are exposing Typesafe's limitations in our API here.

cordformation/src/test/resources/postgres.gradle
```diff
@@ -0,0 +1,120 @@
...
+                         DBPORT         : dbPort
+                    ],
+                    dbUser          : dbUser,
+                    dbPassword      : dbPassword,
+                    dbSchema        : dbSchema,
+                    dbName          : dbName,
+                    dbPort          : dbPort,
+                    dbHostName      : dbHostName,
+                    dbDatabase      : dbName
+					]
```
chrisr3, line 48: Am I _completely_ pedantic about whitespace? Err, "yes" actually :wink:. This `]` bracket is _waaaaaaay_ out...

cordformation/src/test/resources/net/corda/plugins/DeployThreeNodeCordappWithNetworkConfigWithDocker.gradle
```diff
@@ -0,0 +1,56 @@
...
+apply from: 'postgres.gradle'
+
+dependencies {
+    cordaRuntime "$corda_group:corda:$corda_release_version"
+    cordaRuntime "$corda_group:corda-node-api:$corda_release_version"
+    cordapp "$corda_group:corda-finance-contracts:$corda_release_version"
+    cordapp "$corda_group:corda-finance-workflows:$corda_release_version"
+}
+
+task prepareDockerNodes(type: net.corda.plugins.Dockerform, dependsOn: ['jar']) {
```
chrisr3, line 15: I think your `Dockerform` task also needs to depend on the `generateInitScripts` task declared in the `postgresql.gradle` plugin. But you only seem to have done this in `DeploySingleNodeWithExternalDBSettingsWithDocker.gradle` :confused:.

chrisr3:
Can I also safely assume that you are developing on a Windows PC and so we can be sure that this PR is working across platforms?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
palade: We are re-writing/updating the same configuration file. Was thinking that option is used only when we copy files

cordformation/src/test/resources/postgres.gradle
palade: Replace with URL template now. One option would be to escape the [forbidden characters ](https://github.com/lightbend/config/blob/master/HOCON.md), but this may clutter the code

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
chrisr3: We're writing bytes to a file, and the Operating System wants us to tell it what to do if there is already a file with the same name on the filing system: Append, Replace or Abort?

palade: The default of Files.write is to create a new file or overwrites an existing file

palade:
> Can I also safely assume that you are developing on a Windows PC and so we can be sure that this PR is working across platforms? (I think I remember seeing a Java classpath from your PC... 😉)

Tested on both Ubuntu 16.04 and Windows 10

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
chrisr3: Fair enough.

cordformation/src/test/resources/postgres.gradle
chrisr3: > One option would be to escape the [forbidden characters ](https://github.com/lightbend/config/blob/master/HOCON.md), but this may clutter the code

For interest, did you explore rewriting the JDBC URL template to "manage" TypeSafe Config's limitations? This would be a "Nice To Have"...

palade: Added a function to add the required escape elements

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -1,15 +1,20 @@
 package net.corda.plugins
 
 import com.typesafe.config.ConfigFactory
+import com.typesafe.config.ConfigParseOptions
+import com.typesafe.config.ConfigSyntax
```
chrisr3, line 5: Please remove unused imports.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -177,4 +182,24 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
 
         Files.write(dockerComposePath, dockerComposeContent.toByteArray())
     }
+
+    private fun encodeTypesafeURL(url: String): String {
+        val matcher = Pattern.compile("\\$\\{([^}]+)\\}").matcher(url);
```
chrisr3, line 187: The `Pattern` only needs to be compiled once, and then you can use it to create as many `Matcher` objects as you like, e.g.
```kotlin
val matcher = PRECOMPILED_PATTERN.matcher(url)
```
IntelliJ also informs me that you don't need to escape the `}` character at the end, and that there's an unnecessary semi-colon there :wink:.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -177,4 +182,24 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            if (startIndex != 0) {
+                builder.append("\"${url.substring(endOfPrevTokenIndex, startIndex)}\"")
+            }
+            val encodedToken = "\${$token}"
+            builder.append(encodedToken)
+            endOfPrevTokenIndex = startIndex + encodedToken.length
+        }
+        builder.append(url.substring(endOfPrevTokenIndex))
+        return builder.toString()
+    }
```
chrisr3, line 204: And the unit tests for this function...? :smiling_imp:
Don't worry, I don't mean full-blown Gradle plugin unit tests for this! Simple tests for this function alone will suffice. So a Kotlin equivalent of:
> When we encode "templateA" and pass it to Typesafe Config, the final string is "valueA" :heavy_check_mark: 
> When we encode "junkTemplateB" and pass it to Typesafe Config, we get "Error B" :skull_and_crossbones:

palade: Moved the encode function to a class and added unit tests 🙏

cordformation/src/main/kotlin/net/corda/plugins/TypesafeUtils.kt
```diff
@@ -0,0 +1,38 @@
...
+        fun encodeString(input: String): String {
+            val matcher = urlTokenPattern.matcher(input)
+            val builder = StringBuilder()
+            var endOfPrevToken = 0
+
+            while (matcher.find()) {
+                val startOfCurToken = matcher.start()
+                val token = matcher.group(1)
+                if (startOfCurToken != 0) {
+                    builder.append("\"${input.substring(endOfPrevToken, startOfCurToken)}\"")
```
chrisr3, line 23: It's better to concatenate your strings using your existing `StringBuilder` here:
```kotlin
builder.append('\"').append(input.substring(endOfPrevToken, startOfCurToken)).append('\"')
```
Note that `StringBuilder` also allows you to add a single character.

cordformation/src/main/kotlin/net/corda/plugins/TypesafeUtils.kt
```diff
@@ -0,0 +1,38 @@
...
+                }
+                val encodedToken = "\${$token}"
+                builder.append(encodedToken)
```
chrisr3, line 26: Again, you can use the `StringBuilder`:
```kotlin
builder.append('\"').append(token).append('\"')
```

cordformation/src/main/kotlin/net/corda/plugins/TypesafeUtils.kt
```diff
@@ -0,0 +1,38 @@
...
+                endOfPrevToken = startOfCurToken + encodedToken.length
+            }
+
+            if (endOfPrevToken == 0) {
+                builder.append("\"${input.substring(endOfPrevToken)}\"")
```
chrisr3, line 31: This _should_ be handling any trailing text after the last token, but it isn't.... (And you can use `builder` to perform your string concatenation too).

cordformation/src/main/kotlin/net/corda/plugins/TypesafeUtils.kt
palade: the token is not escaped, only what is outside of the token

cordformation/src/test/resources/postgres.gradle
```diff
@@ -0,0 +1,120 @@
+ext {
+    postgresql_version     = '42.2.12'
+    postgres_image_version = '12'
```
JamesHR3, line 3: I don't think we actually test against postgres 12, so probably we should default the postgres version to 11 (which we do test against). (I suspect 12 works just fine, but better safe than sorry)

palade: Updated Postgres Docker image version to 11

PR #298, opened 2020-04-07
ENT-4493 Allow setting command line flag to force schema management by the node

blsemo:
Allow projects using cordform to configure whether they need the command line flags to enable database schema management by the node

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Cordform.kt
@@ -2,6 +2,7 @@ package net.corda.plugins
 
 import org.apache.tools.ant.filters.FixCrLfFilter
 import org.gradle.api.model.ObjectFactory
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.TaskAction
 import javax.inject.Inject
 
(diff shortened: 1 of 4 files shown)


blsemo:
Sadly, it needs to run in both places - the bootstrapper will not fire up the full node, namely, it will not fully intialise CorDapps, so the CorDapp schemas won't be set up. Thus the node runner needs to set the flags again. This is completely in line with the current behaviour where `initialiseSchema` always defaults to true and `initaliseAppSchema` to `UPDATE` (no matter what the clever devMode dependent default logic in `NodeConfigurationImpl.kt` makes to want you believe.

chrisr3:
> Sadly, it needs to run in both places...

So where's the code to pass this parameter to the Network Bootstrapper? All CordForm seems to do is transform the `runnodes` script.

blsemo:
I changed the bootstrapper to always set the flag on the Corda side - no point bootstrapping if you can't modify schemas. Thus cordform the bootstrapper invocation and behaviour didn't change. I can't do the same thing for running nodes, hence that requires a Cordform change.

chrisr3:
> I changed the bootstrapper to always set the flag on the Corda side ...

Does this mean that your new `Cordform` option is ineffective for Corda 4.4 and earlier?

blsemo:
Previous versions of Corda (including the tip of 4.5 at the moment) do not support the flags this sets and will barf if you set them - hence this requires a config flag being set in the calling `build.gradle` to change anything at all so I can turn it on in line with the code changes that support it and require it.

PR #299, opened 2020-04-16
EG-1034: Extend the Dockerform plugin to include persistent volumes

palade:
Extended Dockerform to include persistent volumes when deploying (Postgres) databases


cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -162,17 +165,39 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                    File(volumeDir).mkdirs()
+                    volumes[volume] = mapOf(
+                            "driver" to "local",
+                            "driver_opts" to mapOf(
+                                 "type" to "none",
+                                 "device" to "$volumeDir",
+                                 "o" to "bind"
+                            )
+                    )
+                    database["volumes"] = listOf("$volume:/var/lib/postgresql/data:z")
```
chrisr3, line 185: _**EEEEEKK!!**_ It says "postgresql"!!!!! :scream: 

What would I do if I wanted to use a persistent BunnyDB instance instead?

palade: Removed

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -162,17 +165,40 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                // attach database dependency to the node service
                 service["image"] = dockerImage ?: "entdocker.software.r3.com/corda-enterprise-java1.8-${it.runtimeVersion().toLowerCase()}"
                 service["depends_on"] = listOf(dbHost)
+
+                // append persistence volume if it is required
+                if (persistent) {
+                    val dbDataVolume = dockerConfig.getString("dockerConfig.dbDataVolume")
+                    val volume = "$dbHost-volume"
+                    val volumeDir = "$nodeBuildDir/data"
+                    File(volumeDir).mkdirs()
```
chrisr3, line 177: What if the directory cannot be created? Although IIRC the `mkdirs()` function is a bit tricky because it only returns `true` if it _actually creates_ the full directory path, and returns `false` if the directory path already exists. So to check that the directory is **really** there is more like:
```kotlin
volumeDir.mkdirs() || volumeDir.isDirectory()
```
You should probably throw an error if the directory doesn't exist and cannot be created either. Gradle provides both `InvalidUserDataException` and `InvalidUserCodeException` for situations like this.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -146,33 +144,70 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                val dbDockerfile = dockerConfig.getString("dockerConfig.dbDockerfile")
+                // Override the port and hostname parameters in dockerfile
+                val dbDockerfileArgs = defaultUrlArgs.withFallback(
+                        dockerConfig.getConfig("dockerConfig.dbDockerfileArgs")).resolve()
+                val database = mutableMapOf(
                         "build" to mapOf(
                              "context" to project.buildDir.absolutePath,
                              "dockerfile" to project.buildDir.resolve(dbDockerfile).toString(),
-                             "args" to args
+                             "args" to dbDockerfileArgs.entrySet().map { it.key to it.value.unwrapped() }.toMap()
```
chrisr3, line 155: Ouch! :face_with_head_bandage:. How about:
```kotlin
dbDockerfileArgs.entrySet().associate { it.key to it.value.unwrapped() }
```
Kotlin is full of inline lambdas like these :man_facepalming:. Just gotta learn 'em all...

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -146,33 +144,70 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                    }
+
+                    var absoluteHostPath = directoryPath.resolve(hostPath)
+                    val hostDir = File(absoluteHostPath.toUri())
+                    if (hostDir.mkdirs() || hostDir.isDirectory) {
+                        volumes["$dbHost-volume"] = mapOf(
+                                "driver" to "local",
+                                "driver_opts" to mapOf(
+                                        "type" to "none",
+                                        "device" to "$absoluteHostPath",
```
chrisr3, line 183: I think `"$absoluteHostPath"` is equivalent to `absoluteHostPath.toString()` because you're not asking Kotlin to join any strings together.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -146,33 +144,70 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                // append persistence volume if it is required
+                if (dockerConfig.hasPath("dockerConfig.dbDataVolume")) {
+                    var hostPath = dockerConfig.getString("dockerConfig.dbDataVolume.hostPath")
+                    var containerPath = dockerConfig.getString("dockerConfig.dbDataVolume.containerPath")
+
+                    if (dockerConfig.hasPath("dockerConfig.dbDataVolume.pathArgs")) {
+                        val pathArgs = dockerConfig.getConfig("dockerConfig.dbDataVolume.pathArgs")
+
+                        hostPath = ConfigFactory.parseString("HOSTPATH=${TypesafeUtils.encodeString(hostPath)}")
+                                .resolveWith(ConfigFactory.empty().withValue("NODENAME", ConfigValueFactory.fromAnyRef(it.nodeDir.name)))
```
chrisr3, line 169: This `NODENAME` substitution variable is bothering me. As I see it, `hostPath` is either a relative path or an absolute one. And if it's relative then should be relative to `nodeDir`. In other words, I would expect something like:
```kotlin
var hostPathStr = dockerConfig.getString("dockerConfig.dbDataVolume.hostPath")

// etc...

hostPathStr = ConfigFactory.parseString("HOSTPATH=${TypesafeUtils.encodeString(hostPathStr)}")
    .resolveWith(pathArgs)
    .getString("HOSTPATH")

// etc...

val hostPath = Paths.get(hostPathStr)
val absoluteHostPath = if (hostPath.isAbsolute) {
    hostPath
} else {
    nodeBuildPath.resolve(hostPath)
}
```
I realise that some of these objects are `String` and others are `Path`, but I hope you get the general idea. To be honest, a `Path` object is a _lot_ more powerful than a `String`, so I'd be inclined to keep hold of the node's directory, e.g.:
```kotlin
val nodeBuildPath = directoryPath.resolve(it.nodeDir.name).toAbsolutePath()
val nodeBuildDir = nodeBuildPath.toString()
```
I like the reuse of the `TypesafeUtils.encodeString()` function. Looking at your `dbDataVolume` elements from the _outside_, I see two elements called "host**Path**" and "container**Path**", and a map called "**path**Args". So why _wouldn't_ I expect the substitutions to be applied to _both_ elements equally?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -146,33 +143,75 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+
+                    var containerPathStr = dockerConfig.getString("dockerConfig.dbDataVolume.containerPath")
+
+                    if (dockerConfig.hasPath("dockerConfig.dbDataVolume.containerPathArgs")) {
+                        val containerPathArgs = dockerConfig.getConfig("dockerConfig.dbDataVolume.containerPathArgs")
+                        containerPathStr = TypesafeUtils.resolveString(containerPathStr, containerPathArgs)
+                    }
+
+                    val hostPath = Paths.get(hostPathStr)
+                    var absoluteHostPath = when {
```
chrisr3, line 177: This can _still_ be a `val` rather than a `var` :grin:

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -146,33 +143,75 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                        hostPath.isAbsolute -> hostPath
+                        else -> nodeBuildPath.resolve(hostPath).toAbsolutePath()
+                    }
+                    val hostDir = File(absoluteHostPath.toUri())
```
chrisr3, line 181: And this can still be
```kotlin
val hostDir = absoluteHostPath.toFile()
```
which means you don't need to import `java.io.File` any more.

cordformation/src/test/resources/postgres.gradle
```diff
@@ -7,8 +7,15 @@ ext {
...
     dbHostName             = 'localhost'
-    dbInit                 = 'Postgres_init.sh'
     dbDockerfile           = 'Postgres_Dockerfile'
+    dbInit                 = 'Postgres_init.sh'
+    dbDataVolume           =  [
+            hostPath      : 'data',
+            containerPath : '/var/lib/postgresql/data:\${PERMISSIONS}',
+            containerPathArgs   : [
+                    PERMISSIONS : "z"
+            ]
```
chrisr3, line 17: As an aisde, I have _absolutely no idea_ what permissions `z` means?! :exploding_head:

palade: z or rw are suffixes used for a volume mount inside a Docker container which tells Docker that the volume that is being attached will be shared. This option needs to be specified because the Postgres docker image, when initialised, changes the folder permissions of the attached volume (folder) for some reason

PR #305, opened 2020-05-05
EG-1856: Allow users to specify Dockerform ports

palade:
Dockerform allocates predefined port numbers instead of using the ones specified in the `node.conf`. Users should be allowed to specify those ports

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/ConfigurationUtils.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/ConfigurationUtils.kt
@@ -1,0 +1,19 @@
+package net.corda.plugins
+
+import org.gradle.api.InvalidUserDataException
+import java.net.URI
+import java.net.URISyntaxException
+
+class ConfigurationUtils {
+
+    companion object {
+
+        fun parsePort(address: String): Int {
+            return try {
+                URI(null, address, null, null, null).port
+            } catch (ex: URISyntaxException) {
+                throw InvalidUserDataException("Invalid host and port syntax for RPC address, expected host:port. Using default value")
+            }
+        }
+    }
+}
--- a/cordformation/src/main/kotlin/net/corda/plugins/ConfigurationUtils.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/ConfigurationUtils.kt
@@ -12,7 +12,7 @@ class ConfigurationUtils {
             return try {
                 URI(null, address, null, null, null).port
             } catch (ex: URISyntaxException) {
-                throw InvalidUserDataException("Invalid host and port syntax for RPC address, expected host:port. Using default value")
+                throw InvalidUserDataException("Invalid host and port syntax for RPC address, expected host:port")
             }
         }
     }
--- a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
@@ -99,7 +99,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/additional-node-infos:/opt/corda/additional-node-infos",
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
                     ),
-                    "ports" to listOf(it.rpcPort(), it.config.getInt("sshd.port")),
+                    "ports" to listOf(it.rpcPort.get(), it.config.getInt("sshd.port")),
                     "image" to (dockerImage ?: "corda/corda-zulu-${it.runtimeVersion().toLowerCase()}")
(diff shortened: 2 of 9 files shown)


cordformation/src/main/kotlin/net/corda/plugins/RpcSettings.kt
```diff
@@ -34,6 +39,9 @@ class RpcSettings {
      * RPC admin address for the node (necessary if [useSsl] is false or unset).
      */
     fun adminAddress(value: String) {
+        try {
+            adminPort = URI(null, value, null, null, null).port
+        } catch (ex: URISyntaxException) { }
```
chrisr3, line 44: Same as above. And maybe make a function that both properties can use for this?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -99,7 +99,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/additional-node-infos:/opt/corda/additional-node-infos",
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
                     ),
-                    "ports" to listOf(it.rpcPort, it.config.getInt("sshd.port")),
+                    "ports" to listOf(it.rpcPort(), it.config.getInt("sshd.port")),
```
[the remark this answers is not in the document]
palade: > Does `sshd.port` need to be different for each node too?

Not necessarily, but a user may want to see the port number that he set in the `build.conf`

palade:
> Looks OK. Is there any way that your new test can also check whether the RPC ports have been set correctly? After all, that _is_ the point of this PR.

Added a unit test to check for correct parsing of the RPC  port. Moved the parsing function in separate file and added several unit tests for that function as well

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -111,5 +114,23 @@ class DockerformTest : BaseformTest() {
...
         assertThat(getNodeCordappJar(bankOfCordaNodeName, cordaFinanceContractsJarName)).isRegularFile()
         assertThat(getNetworkParameterOverrides(bankOfCordaNodeName)).isRegularFile()
+
+        val notaryConfigPath = getNodeConfig(notaryNodeName)
+        assertThat(notaryConfigPath).isRegularFile()
+
+        val notaryConfig = ConfigFactory.parseFile(notaryConfigPath.toFile())
+        assertTrue(notaryConfig.hasPath("rpcSettings.address"))
+        assertTrue(notaryConfig.hasPath("rpcSettings.adminAddress"))
+        assertEquals(ConfigurationUtils.parsePort(notaryConfig.getString("rpcSettings.address")), 10003)
```
chrisr3, line 124: Funny thing - this `assertEquals()` function is really:
```kotlin
assertEquals(expectedValue, actualValue)
```
i.e. the assertion should be written as:
```kotlin
assertEquals(10003, ConfigurationUtils.parsePort(notaryConfig.getString("rpcSettings.address")))
```
(FYI different frameworks put these parameters in different orders. I don't know why....)

cordformation/src/test/kotlin/net/corda/plugins/ConfigurationUtilsTest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+
+import org.junit.jupiter.api.Test
+import kotlin.test.assertEquals
+
+class ConfigurationUtilsTest {
+
+    @Test
+    fun `check correct port value parsing`() {
+        assertEquals(ConfigurationUtils.parsePort("localhost:10000"), 10000)
```
chrisr3, line 11: Oops, `assertEquals(expectedValue, actualValue)`...

cordformation/src/main/kotlin/net/corda/plugins/ConfigurationUtils.kt
```diff
@@ -0,0 +1,19 @@
...
+
+class ConfigurationUtils {
+
+    companion object {
+
+        fun parsePort(address: String): Int {
+            return try {
+                URI(null, address, null, null, null).port
+            } catch (ex: URISyntaxException) {
+                throw InvalidUserDataException("Invalid host and port syntax for RPC address, expected host:port. Using default value")
```
chrisr3, line 15: This exception will kill Gradle, which is fine. It's just that the message shouldn't therefore say that the default value will be used.

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
palade: Fixed now :-)

Issue #275, opened 2019-12-22
Dockerform generated docker image version "corda/corda-zulu-4.3" is not available in docker hub


RimashMohomed:
Description
docker-compose file is created via  net.corda.plugins.Dockerform gradle plugin using following task from the documentation contains the docker image ```image: corda/corda-zulu-4.3``` not available in the docker hub which results below error. 

Only by downgraing the app to 4.1 solved the issue.

**Error**

```Pulling partyb (corda/corda-zulu-4.3:latest)...
ERROR: Get https://registry-1.docker.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

**gradle task**

```task prepareDockerNodes(type: net.corda.plugins.Dockerform, dependsOn: ['jar']) {
    nodeDefaults {
        cordapp project(":contracts")
    }
    node {
        name "O=Notary,L=London,C=GB"
        notary = [validating : false]
        p2pPort 10002
        rpcSettings {
            address("localhost:10003")
            adminAddress("localhost:10023")
        }
        projectCordapp {
            deploy = false
        }
        cordapps.clear()
    }
    node {
        name "O=PartyA,L=London,C=GB"
        p2pPort 10002
        rpcSettings {
            address("localhost:10003")
            adminAddress("localhost:10023")
        }
        rpcUsers = [[user: "user1", "password": "test", "permissions": ["ALL"]]]
    }
    node {
        name "O=PartyB,L=New York,C=US"
        p2pPort 10002
        rpcSettings {
            address("localhost:10003")
            adminAddress("localhost:10023")
        }
        rpcUsers = [[user: "user1", "password": "test", "permissions": ["ALL"]]]
    }
}`

**generated docker-compose.yml**

```version: '3'
services:
  notary:
    volumes:
    - /home/rimash/Work/Corda/cordapp-template-java/build/nodes/Notary/node.conf:/etc/corda/node.conf
    - /home/rimash/Work/Corda/cordapp-template-java/build/nodes/Notary/certificates:/opt/corda/certificates
    - /home/rimash/Work/Corda/cordapp-template-java/build/nodes/Notary/logs:/opt/corda/logs
    - /home/rimash/Work/Corda/cordapp-template-java/build/nodes/Notary/persistence:/opt/corda/persistence
    - /home/rimash/Work/Corda/cordapp-template-java/build/nodes/Notary/cordapps:/opt/corda/cordapps
[...]

PR #307, opened 2020-05-07
EG-1902: Adjust the default Corda Docker image name

palade:
[Most of the recent](https://hub.docker.com/search?q=corda%2Fcorda-&sort=updated_at&order=desc) Corda Docker images on DockerHub follow the format corda/corda-zulu-java1.8-*. This also has been reported as an [issue](https://github.com/corda/corda-gradle-plugins/issues/275). The default Docker image name in Dockerform is adjusted to address this issue.

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
@@ -124,21 +124,12 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                 }
 
                 // Install the database configuration
-                val dbUser = dockerConfig.getString("dockerConfig.dbUser")
-                val dbPassword = dockerConfig.getString("dockerConfig.dbPassword")
-                val dbDataSourceClassName = dockerConfig.getString("dataSourceProperties.dataSourceClassName")
-                val dbTransactionIsolationLevel = dockerConfig.getString("database.transactionIsolationLevel")
-                val dbRunMigration = dockerConfig.getBoolean("database.runMigration")
-                val dbSchema= dockerConfig.getString("dockerConfig.dbSchema")
-
                 val dbConfig = ConfigFactory.empty()
-                        .withValue("dataSourceProperties.dataSourceClassName", ConfigValueFactory.fromAnyRef(dbDataSourceClassName))
                         .withValue("dataSourceProperties.dataSource.url", ConfigValueFactory.fromAnyRef(dbUrl))
-                        .withValue("dataSourceProperties.dataSource.user", ConfigValueFactory.fromAnyRef(dbUser))
-                        .withValue("dataSourceProperties.dataSource.password", ConfigValueFactory.fromAnyRef(dbPassword))
-                        .withValue("database.transactionIsolationLevel", ConfigValueFactory.fromAnyRef(dbTransactionIsolationLevel))
-                        .withValue("database.runMigration", ConfigValueFactory.fromAnyRef(dbRunMigration))
-                        .withValue("database.schema", ConfigValueFactory.fromAnyRef(dbSchema))
+                        .withValue("dataSourceProperties.dataSourceClassName", dockerConfig.getValue("dataSourceProperties.dataSourceClassName"))
+                        .withValue("dataSourceProperties.dataSource.user", dockerConfig.getValue("dockerConfig.dbUser"))
+                        .withValue("dataSourceProperties.dataSource.password", dockerConfig.getValue("dockerConfig.dbPassword"))
+                        .withValue("database", ConfigValueFactory.fromMap(dockerConfig.getObject("database")))
 
                 it.installDefaultDatabaseConfig(dbConfig)
 
--- a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
@@ -60,9 +60,8 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
     var dockerConfig: Map<String, Any> = emptyMap()
 
     /**
-     * Docker image for this node, or null if one hasn't been specified.
+     * Docker image for this node
      */
-    @get:Optional
     @get:Input
     var dockerImage: String? = null
 
@@ -80,7 +79,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
         bootstrapNetwork()
         nodes.forEach(Node::buildDocker)
 
-        val services = mutableMapOf<String, MutableMap<String, Any>>()
+        val services = mutableMapOf<String, MutableMap<String, Any?>>()
         val volumes = mutableMapOf<String, Map<String, Any>>()
 
         nodes.forEachIndexed {index, it ->
@@ -100,7 +99,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
                     ),
                     "ports" to listOf(it.rpcPort.get(), it.config.getInt("sshd.port")),
-                    "image" to (dockerImage ?: "corda/corda-zulu-java1.8-${it.runtimeVersion().toLowerCase()}")
+                    "image" to dockerImage
             )
 
             if (dockerConfig.isNotEmpty()) {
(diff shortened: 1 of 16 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -100,7 +100,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
                     ),
                     "ports" to listOf(it.rpcPort.get(), it.config.getInt("sshd.port")),
-                    "image" to (dockerImage ?: "corda/corda-zulu-${it.runtimeVersion().toLowerCase()}")
+                    "image" to (dockerImage ?: "corda/corda-zulu-java1.8-${it.runtimeVersion().toLowerCase()}")
```
chrisr3, line 103: This kind of thing shows why it is better to _force_ the user to specify the image they want to use in the first place. If some people had their way, the plugins would be responsible for :sparkles: **automagically** :sparkles: configuring everything from CorDapp developers' compiler settings onwards. And consequently, every subsequent misconfiguration problem would become the plugins' problem too.

palade: The users can now specify the Docker image by using dockerImage property, but is provided as optional. Should we make this change this?

JamesHR3: I think that's a breaking change so you probably can't - unless we have a different policy around the plugins.

chrisr3: To be blunt: _Yes_, I would prefer that the image were a mandatory `@Input` property without a default. Because otherwise we'll be right back here when (say) we switch to Java 11 or someone decides that the default should be different for any other reason.

chrisr3: > I think that's a breaking change so you probably can't - unless we have a different policy around the plugins.

It was broken back when the name of the base image changed. And we really don't want to open the Pandora's Box of trying to hard-code different default values for different versions of Corda into the plugins. So maybe it's better to leave it broken for now and help people migrate? And _then_ we kill the default entirely.

chrisr3: Come to think of it, this PR is _itself_ a breaking change for those using Corda 4.3 and earlier.

palade: Made the dockerImage property mandatory and also update the docs to reflect this change

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithExternalDBSettingsWithDocker.gradle
```diff
@@ -21,5 +21,8 @@ task prepareDockerNodes(type: net.corda.plugins.Dockerform, dependsOn: 'generate
         cordapps = ["$corda_group:corda-finance-contracts:$corda_release_version",
                     "$corda_group:corda-finance-workflows:$corda_release_version"]
     }
+
     dockerConfig = postgres
+
+    dockerImage = "corda/corda-zulu-java1.8-4.3"
```
chrisr3, line 27: You could add this value to the tests' `gradle.properties` file:
```
docker_image_name=corda/corda-zulu-java1.8-4.3
```
and then your `.gradle` test files could all refer to it:
```
dockerImage = docker_image_name
```

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -60,9 +60,8 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
     var dockerConfig: Map<String, Any> = emptyMap()
 
     /**
-     * Docker image for this node, or null if one hasn't been specified.
+     * Docker image for this node
      */
-    @get:Optional
     @get:Input
     var dockerImage: String? = null
```
chrisr3, line 66: The `dockerImage` property should still never be `null` though. Maybe use a "lazy" Gradle property here?
```kotlin
@get:Input
val dockerImage: Property<String> = objects.property(String::class.java)
```
And then you wouldn't need to change the types of your `MutableMap<K,V>` objects either.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -100,7 +99,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
                     ),
                     "ports" to listOf(it.rpcPort.get(), it.config.getInt("sshd.port")),
-                    "image" to (dockerImage ?: "corda/corda-zulu-${it.runtimeVersion().toLowerCase()}")
+                    "image" to dockerImage
```
chrisr3, line 102: Using a "lazy" property would make this:
```kotlin
"image" to dockerImage.get()
```

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -3,24 +3,23 @@ package net.corda.plugins
...
 class DockerformTest : BaseformTest() {
 
     @Test
-    fun `deploy a node with cordapp dependency`() {
+    fun `fail to deploy a node with cordapp dependency with missing dockerImage tag`() {
         val runner = getStandardGradleRunnerFor(
                 "DeploySingleNodeWithCordappWithDocker.gradle",
                 "prepareDockerNodes")
 
-        val result = runner.build()
```
chrisr3, line 18: I like the extra test. Gradle's test-kit also allows you to test error paths like this:
```
val result = runner.buildAndFail()
```
This allows you to inspect the outcome of the `prepareDockerNodes` task so you can be _sure_ that the test failed for the correct reason.

chrisr3:
The build output from each `GradleRunner` can be read from `BuildResult.getOutput()`, which means you can do this:
```gradle
val result = runner.build()
println(result.output)
```
This adds the output to the JUnit report.

cordformation/src/test/resources/gradle.properties
```diff
@@ -3,4 +3,5 @@ org.gradle.jvmargs=-XX:+UseG1GC -Xmx512m
 org.gradle.caching=false
 corda_group=net.corda
 corda_release_version=4.0
-jolokia_version=1.6.0
\ No newline at end of file
+jolokia_version=1.6.0
+docker_image_name=corda/corda-zulu-java1.8-4.3
```
chrisr3, line 7: What's worrying me here is that `docker_image_name` is _really_
```
corda/corda-zulu-java1.8-$corda_release_version
```
but `corda_release_version` is still `4.0`. So these properties are not consistent. Even worse, this test is probably deploying CorDapps from 4.0 into a Corda 4.3 Docker image. Think about how you could make these tests consistent.

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithExternalDBSettingsWithDocker.gradle
```diff
@@ -21,5 +21,8 @@ task prepareDockerNodes(type: net.corda.plugins.Dockerform, dependsOn: 'generate
         cordapps = ["$corda_group:corda-finance-contracts:$corda_release_version",
                     "$corda_group:corda-finance-workflows:$corda_release_version"]
     }
+
     dockerConfig = postgres
+
+    dockerImage = docker_image_name
```
chrisr3, line 27: Is this the correct Docker image for use with an external database? I see you have removed a reference to this image from the plugin:
```
entdocker.software.r3.com/corda-enterprise-java1.8-4.3
```
I had been assuming that this image has extra added "databaseyness". However, if it is just the basic Docker image for Corda Enterprise 4.3 then what you have here is already fine, with no action required.

FYI the Corda Gradle Plugins are Open Source, and are used equally by both Corda Enterprise and Corda OS. They therefore should not have Enterprise-specific default values - something I probably should have picked up on when I reviewed the previous PR.

cordformation/src/test/resources/gradle.properties
palade: I'm changing this to `corda/corda-zulu-4.0` to highlight the consistency between the Corda Docker image and Corda version values. While this Docker image name points to a valid Docker image on DockerHub, the tests are to ensure that the dockerImage tag must provided as part of the Dockerform task.

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithExternalDBSettingsWithDocker.gradle
palade: That is just a basic Docker image for Corda Enterprise 4.3. Also, in the last commit I changed the Dockerform slightly to remove as everything ENT specific. It is now up to the user to provide the correct configuration and the correct Docker image

PR #308, opened 2020-05-12
ENT-5258 Handle db migration in Cordform

blsemo:
In Corda 4.6, database schema migration/set-up will have to be run explicitly via a new subcommand - this change brings in the code to do this during cordformation, currently guarded behind a property that defaults to off.

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -145,10 +145,11 @@ open class Node @Inject constructor(private val project: Project) {
         @Optional @Input get
         private set
 
-    // Should the cordform set up run data base migration scripts after installation
+    // Should the cordform set up run data base migration scripts after installation - defaults to false for compatibility
+    // with current and previous Corda versions
     @get:Optional
     @get:Input
-    var runSchemaMigration: Boolean = true
+    var runSchemaMigration: Boolean = false
 
     // If generating schemas, should app schema be generated using hibernate if missing migration scripts
     @get:Optional
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -5,6 +5,7 @@ import groovy.lang.Closure
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ProjectDependency
+import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
@@ -149,12 +150,17 @@ open class Node @Inject constructor(private val project: Project) {
     // with current and previous Corda versions
     @get:Optional
     @get:Input
-    var runSchemaMigration: Boolean = false
+    var runSchemaMigration: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType).convention(false)
 
     // If generating schemas, should app schema be generated using hibernate if missing migration scripts
     @get:Optional
     @get:Input
-    var allowHibernateToManageAppSchema: Boolean = false
+    var allowHibernateToManageAppSchema: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType).convention(false)
+
+    //Configure the timeout for schema generation runtime
+    @get:Optional
+    @get:Input
+    var nodeJobTimeOut: Property<Long> = project.objects.property(Long::class.javaObjectType).convention(3)
 
     /**
      * Set the name of the node.
@@ -491,11 +497,11 @@ open class Node @Inject constructor(private val project: Project) {
             "-jar",
             "corda.jar",
             "run-migration-scripts",
-            if (allowHibernateToManageAppSchema) "--allow-hibernate-to-manage-app-schema" else null)
+            if (allowHibernateToManageAppSchema.get()) "--allow-hibernate-to-manage-app-schema" else null)
 
     private fun runSchemaMigration(){
-        if (!runSchemaMigration) return
-        project.logger.lifecycle("Run database schema migration scripts${if(allowHibernateToManageAppSchema) " - managing CorDapp schemas with hibernate" else ""}")
+        if (!runSchemaMigration.get()) return
+        project.logger.lifecycle("Run database schema migration scripts${if(allowHibernateToManageAppSchema.get()) " - managing CorDapp schemas with hibernate" else ""}")
         runNodeJob(createSchemasCmd(), "node-schema-cordform.log")
     }
 
@@ -511,11 +517,11 @@ open class Node @Inject constructor(private val project: Project) {
                 .apply { environment()["CAPSULE_CACHE_DIR"] = "../.cache" }
                 .start()
         try {
-            if (!process.waitFor(3, TimeUnit.MINUTES)) {
+            if (!process.waitFor(nodeJobTimeOut.get(), TimeUnit.MINUTES)) {
                 process.destroyForcibly()
-                printNodeOutputToConsoleAndThrow(nodeRedirectFile)
+                printNodeOutputAndThrow(nodeRedirectFile)
             }
-            if (process.exitValue() != 0) printNodeOutputToConsoleAndThrow(nodeRedirectFile)
+            if (process.exitValue() != 0) printNodeOutputAndThrow(nodeRedirectFile)
         } catch (e: InterruptedException) {
             // Don't leave this process dangling if the thread is interrupted.
             process.destroyForcibly()
@@ -523,11 +529,10 @@ open class Node @Inject constructor(private val project: Project) {
         }
     }
 
-    private fun printNodeOutputToConsoleAndThrow(stdoutFile: File) {
-        val nodeDir = stdoutFile.parent
-        System.err.println("#### Error while generating node info file $name ####")
-        stdoutFile.inputStream().copyTo(System.err)
-        throw IllegalStateException("Error while generating node info file. Please check the logs in $nodeDir.")
+    private fun printNodeOutputAndThrow(stdoutFile: File) {
+        project.logger.error("#### Error while generating node info file $name ####")
+        project.logger.error(stdoutFile.readText())
+        throw IllegalStateException("Error while generating node info file. Please check the logs in ${stdoutFile.parent}.")
     }
 
     fun runtimeVersion(): String {
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -150,17 +150,17 @@ open class Node @Inject constructor(private val project: Project) {
     // with current and previous Corda versions
     @get:Optional
     @get:Input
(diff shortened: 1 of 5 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -470,6 +485,51 @@ open class Node @Inject constructor(private val project: Project) {
...
+    val LOGS_DIR_NAME: String = "logs"
+
+    private fun runNodeJob(command: List<String>, logfileName: String) {
+        val logsDir = Files.createDirectories(nodeDir.toPath().resolve(LOGS_DIR_NAME))
+        val nodeRedirectFile = logsDir.resolve(logfileName).toFile()
+        val process = ProcessBuilder(command)
+                .directory(nodeDir)
+                .redirectErrorStream(true)
+                .redirectOutput(nodeRedirectFile)
+                .apply { environment()["CAPSULE_CACHE_DIR"] = "../.cache" }
```
woggioni, line 511: may I ask what is this for?

blsemo: Hmm - that is copied and pasted 😃 
It creates the capsule cache for this run locally rather than in the user's home directory, which means it gets wiped when running e.g. `./gradlew clean`. But I am not overly attached to it.

chrisr3: But it does also mean that your `corda-4.5-SNAPSHOT.jar` is the SNAPSHOT you actually asked for, and hasn't been contaminated by (say) last week's 4.5-SNAPSHOT which was lingering in your `$HOME/.capsule` directory.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -143,6 +145,17 @@ open class Node @Inject constructor(private val project: Project) {
         @Optional @Input get
         private set
 
+    // Should the cordform set up run data base migration scripts after installation - defaults to false for compatibility
+    // with current and previous Corda versions
+    @get:Optional
+    @get:Input
+    var runSchemaMigration: Boolean = false
```
chrisr3, line 152: Modern Gradle properties are supposed to be "lazy" properties whose values are computed when Gradle's configuration phase is complete.
```kotlin
@get:Optional
@get:Input
val runSchemaMigration: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType).convention(false)
```
or something like that.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -143,6 +145,17 @@ open class Node @Inject constructor(private val project: Project) {
...
+
+    // If generating schemas, should app schema be generated using hibernate if missing migration scripts
+    @get:Optional
+    @get:Input
+    var allowHibernateToManageAppSchema: Boolean = false
```
chrisr3, line 157: Similarly for this property.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -470,6 +485,51 @@ open class Node @Inject constructor(private val project: Project) {
...
+        val logsDir = Files.createDirectories(nodeDir.toPath().resolve(LOGS_DIR_NAME))
+        val nodeRedirectFile = logsDir.resolve(logfileName).toFile()
+        val process = ProcessBuilder(command)
+                .directory(nodeDir)
+                .redirectErrorStream(true)
+                .redirectOutput(nodeRedirectFile)
+                .apply { environment()["CAPSULE_CACHE_DIR"] = "../.cache" }
+                .start()
+        try {
+            if (!process.waitFor(3, TimeUnit.MINUTES)) {
```
chrisr3, line 514: Should this interval be configurable?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -470,6 +485,51 @@ open class Node @Inject constructor(private val project: Project) {
...
+        } catch (e: InterruptedException) {
+            // Don't leave this process dangling if the thread is interrupted.
+            process.destroyForcibly()
+            throw e
+        }
+    }
+
+    private fun printNodeOutputToConsoleAndThrow(stdoutFile: File) {
+        val nodeDir = stdoutFile.parent
+        System.err.println("#### Error while generating node info file $name ####")
```
chrisr3, line 528: Can you use Gradle's own `logger` rather that `System.err.println()`?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
blsemo: Maybe? It isn't in the bootstrapper, and I think 3 minutes is generous for schema creation - but I can make it configurable for good measure.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
blsemo: Yes, sure. Copy and paste error.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
blsemo: I've done that change - I don't like the required changes to the build.gradle file though - I am not sure that the realised benefits are worth changing
```
runSchemaMigration = true
```
to 
```
runSchemaMigration.set(true)
```

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -143,6 +146,22 @@ open class Node @Inject constructor(private val project: Project) {
         @Optional @Input get
         private set
 
+    // Should the cordform set up run data base migration scripts after installation - defaults to false for compatibility
+    // with current and previous Corda versions
+    @get:Optional
+    @get:Input
+    var runSchemaMigration: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType).convention(false)
```
chrisr3, line 153: Use `val` instead of `var`.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -143,6 +146,22 @@ open class Node @Inject constructor(private val project: Project) {
...
+
+    // If generating schemas, should app schema be generated using hibernate if missing migration scripts
+    @get:Optional
+    @get:Input
+    var allowHibernateToManageAppSchema: Property<Boolean> = project.objects.property(Boolean::class.javaObjectType).convention(false)
```
chrisr3, line 158: Use `val` instead of `var`.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -143,6 +146,22 @@ open class Node @Inject constructor(private val project: Project) {
...
+
+    //Configure the timeout for schema generation runtime
+    @get:Optional
+    @get:Input
+    var nodeJobTimeOut: Property<Long> = project.objects.property(Long::class.javaObjectType).convention(3)
```
chrisr3, line 163: Use `val` instead of `var`.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -470,6 +491,50 @@ open class Node @Inject constructor(private val project: Project) {
...
+            }
+            if (process.exitValue() != 0) printNodeOutputAndThrow(nodeRedirectFile)
+        } catch (e: InterruptedException) {
+            // Don't leave this process dangling if the thread is interrupted.
+            process.destroyForcibly()
+            throw e
+        }
+    }
+
+    private fun printNodeOutputAndThrow(stdoutFile: File) {
```
chrisr3, line 532: Perhaps also give this function a return type of `Nothing` so that Kotlin knows it will not return?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
blsemo: I changed it to `val` locally - works exactly the same.
Still refuses to accept `runSchemaMigration = true`.

chrisr3:
Would you also care to create a unit test for this new Cordformation feature and to update the changelog please?

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -470,6 +491,50 @@ open class Node @Inject constructor(private val project: Project) {
...
+            "run-migration-scripts",
+            if (allowHibernateToManageAppSchema.get()) "--allow-hibernate-to-manage-app-schema" else null)
+
+    private fun runSchemaMigration(){
+        if (!runSchemaMigration.get()) return
+        project.logger.lifecycle("Run database schema migration scripts${if(allowHibernateToManageAppSchema.get()) " - managing CorDapp schemas with hibernate" else ""}")
+        runNodeJob(createSchemasCmd(), "node-schema-cordform.log")
+    }
+
+    val LOGS_DIR_NAME: String = "logs"
```
chrisr3, line 508: This property should probably be `private`, or else Gradle will expect it to be annotated with `@Input`, `@Internal`etc.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -5,6 +5,7 @@ import groovy.lang.Closure
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.artifacts.ProjectDependency
+import org.gradle.api.provider.Property
```
chrisr3, line 8: Unused import.

cordformation/src/main/kotlin/net/corda/plugins/Node.kt
```diff
@@ -470,6 +491,50 @@ open class Node @Inject constructor(private val project: Project) {
...
+            "run-migration-scripts",
+            if (allowHibernateToManageAppSchema) "--allow-hibernate-to-manage-app-schema" else null)
+
+    private fun runSchemaMigration(){
+        if (!runSchemaMigration) return
+        project.logger.lifecycle("Run database schema migration scripts${if(allowHibernateToManageAppSchema) " - managing CorDapp schemas with hibernate" else ""}")
+        runNodeJob(createSchemasCmd(), "node-schema-cordform.log")
+    }
+
+    val LOGS_DIR_NAME: String = "logs"
```
chrisr3, line 508: This still needs to be private as it is causing these build warnings:
```
  - Warning: Task type 'net.corda.plugins.Baseform': property 'nodes*.LOGS_DIR_NAME' is not annotated with an input or output annotation.
  - Warning: Task type 'net.corda.plugins.Cordform': property 'nodes*.LOGS_DIR_NAME' is not annotated with an input or output annotation.
  - Warning: Task type 'net.corda.plugins.Dockerform': property 'nodes*.LOGS_DIR_NAME' is not annotated with an input or output annotation.
```

blsemo:
After a lot of experimenting we have concluded that the cordform plug-in is currently locked into the old way of configuration properties, and it would take some extra effort to move it on (ticket has been created https://r3-cev.atlassian.net/browse/CORDA-3813).

PR #309, opened 2020-05-26
ENT-5364: Added ACCEPT_LICENSE environment variable

palade:
This change allows ACCEPT_LICENSE environment variable to be passed to the Corda Docker Image. This variable is important for the ENT version and has no effect on the OS version.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 
 ### Version 5.0.10
 
+* `cordformation`: Add option to accept license agreements for Corda enterprise Docker images 
 * `cordformation`: Add properties `runSchemaMigration` and `allowHibernateToManageAppSchema` in order to run schema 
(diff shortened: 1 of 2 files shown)


palade:
> I don't know Docker well enough to understand how this should work. I see you adding a line
> 
> ```
> ACCEPT_LICENCE=${ACCEPT_LICENCE}
> ```
> 
> to the `environment` elements of the `services` map, but where/how does the user set the actual value for `ACCEPT_LICENCE`?

There are a few ways to set the value of that variable, but these are the most common ones:
1. As an environment variable. 
2. When running the docker-compose task:   `ACCEPT_LICENSE=Y docker-compose up`

The latest version ENT Docker imager requires this value to be set using an environment variable to show that the user has accepted the license agreements (please see licenseAgreementCheck.sh in the ent repository).

PR #311, opened 2020-05-31
NOTICK: Make Gradle scripts for cordformation tests consistent.

chrisr3:
- Use common `gradle.properties` instead of `buildscript` wherever possible.
- Replace deprecated `runtime` configuration with `cordaRuntime`
- Provide simple SLF4J implementation for Network Bootstrapper.

### DIFF
--- a/cordformation/src/test/resources/gradle.properties
+++ b/cordformation/src/test/resources/gradle.properties
@@ -5,3 +5,5 @@ corda_group=net.corda
 corda_release_version=4.0
 jolokia_version=1.6.0
 docker_image_name=corda/corda-zulu-4.0
+
+slf4j_version=1.7.26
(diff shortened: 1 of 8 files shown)


chrisr3:
> Likely unrelated to these changes, but two tests from `CordformTest` fail on my machine: `network parameter overrides()` and `a node that requires an extra command to create schema()`.

Cannot reproduce these problems here. Are you running the tests from within IntelliJ? Because if so, please ensure that IntelliJ is delegating all build operations to Gradle because the plugins project does not support IntelliJ's own funky way of doing things.

palade:
Yes I was running the tests from Intellij. There was only one log message:

> Error: Could not find or load main class net.corda.node.Corda

Works when I build it from command line

PR #312, opened 2020-06-05
[CORDA-3829] extended NetworkParameterOverrides object

woggioni:
extended `NetworkParameterOverrides` object to support `minimumPlatformVersion`, `maxMessageSize`, `maxTransactionSize` and `eventHorizon`

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
@@ -283,7 +283,7 @@ open class Baseform(objects: ObjectFactory) : DefaultTask() {
 
     private fun invokeBootstrap(networkBootstrapperClass: Class<*>, rootDir: Path, allCordapps: List<Path>) {
         try {
-            if (networkParameterOverrides.packageOwnership.isEmpty()) {
+            if (networkParameterOverrides.isEmpty()) {
                 val bootstrapMethod = networkBootstrapperClass.getMethod("bootstrapCordform", Path::class.java, List::class.java).apply { isAccessible = true }
                 bootstrapMethod.invoke(networkBootstrapperClass.newInstance(), rootDir, allCordapps)
             } else {
--- a/cordformation/src/main/kotlin/net/corda/plugins/NetworkParameterOverrides.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/NetworkParameterOverrides.kt
@@ -1,16 +1,29 @@
 package net.corda.plugins
 
-import com.typesafe.config.Config
-import com.typesafe.config.ConfigFactory
-import com.typesafe.config.ConfigObject
-import com.typesafe.config.ConfigValueFactory
+import com.typesafe.config.*
 import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Project
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Nested
+import org.gradle.api.tasks.Optional
+import java.time.Duration
 import javax.inject.Inject
 
 open class NetworkParameterOverrides @Inject constructor(project: Project) {
+
+    @get:Optional
+    @get:Input
+    val minimumPlatformVersion: Property<Int> = project.objects.property(Int::class.java)
+
+    @get:Optional
+    @get:Input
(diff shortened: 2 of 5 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
```diff
@@ -283,13 +283,8 @@ open class Baseform(objects: ObjectFactory) : DefaultTask() {
...
         try {
-            if (networkParameterOverrides.packageOwnership.isEmpty()) {
-                val bootstrapMethod = networkBootstrapperClass.getMethod("bootstrapCordform", Path::class.java, List::class.java).apply { isAccessible = true }
-                bootstrapMethod.invoke(networkBootstrapperClass.newInstance(), rootDir, allCordapps)
-            } else {
-                val bootstrapMethod = networkBootstrapperClass.getMethod("bootstrapCordform", Path::class.java, List::class.java, String::class.java).apply { isAccessible = true }
-                bootstrapMethod.invoke(networkBootstrapperClass.newInstance(), rootDir, allCordapps, networkParameterOverrides.toConfig().root().render(ConfigRenderOptions.concise()))
-            }
+            val bootstrapMethod = networkBootstrapperClass.getMethod("bootstrapCordform", Path::class.java, List::class.java, String::class.java).apply { isAccessible = true }
+            bootstrapMethod.invoke(networkBootstrapperClass.newInstance(), rootDir, allCordapps, networkParameterOverrides.toConfig().root().render(ConfigRenderOptions.concise()))
```
chrisr3, line 287: The Network Bootstrapper entry point supporting `networkParameterOverrides` was only added in Corda 4.3. Why have you removed support for the previous entry point? This breaks compatibility with earlier versions of Corda.

woggioni: ok, I didn't know that. That's probably what broke the unit tests. The reason I removed it it because it forces to write boilerplate code like 
```
if this is empty and that is null and that_other is null ...
then do this
```
but if it is necessary I'll do that

woggioni:
there is no specific ticket, I just needed to specify the minimum platform version in Cordformation

chrisr3:
> there is no specific ticket, I just needed to specify the minimum platform version in Cordformation

Then please _raise_ a JIRA ticket for this piece of work,

cordformation/src/main/kotlin/net/corda/plugins/NetworkParameterOverrides.kt
```diff
@@ -1,29 +1,58 @@
...
 
     fun packageOwnership(action: Action<in NamedDomainObjectContainer<PackageOwnership>>) {
         action.execute(packageOwnership)
     }
 
+    @get:Optional
+    @get:Input
+    var eventHorizon: Duration? = null
+
+    fun isEmpty() = minimumPlatformVersion == null &&
```
chrisr3, line 38: This `isEmpty()` function is missing the `eventHorizon` property. You can check whether a "lazy" property has been assigned a value using `isPresent`.

cordformation/src/main/kotlin/net/corda/plugins/NetworkParameterOverrides.kt
```diff
@@ -1,29 +1,58 @@
...
+            maxMessageSize == null &&
+            maxTransactionSize == null &&
+            packageOwnership.isEmpty()
+
     fun toConfig(): Config {
-        val packageOwnershipsList = mutableListOf<ConfigObject>()
```
chrisr3, line 23: This function would also need modifying to support "lazy" Gradle properties.

cordformation/src/main/kotlin/net/corda/plugins/NetworkParameterOverrides.kt
```diff
@@ -1,29 +1,60 @@
...
 import org.gradle.api.tasks.Nested
+import org.gradle.api.tasks.Optional
+import java.time.Duration
 import javax.inject.Inject
 
 open class NetworkParameterOverrides @Inject constructor(project: Project) {
 
+    @get:Optional
+    @get:Input
+    var minimumPlatformVersion: Property<Int> = project.objects.property(Int::class.java)
```
chrisr3, line 18: These should all be `val` instead of `var`. A "lazy" property is a container object for a value, and the _container_ will not change.

PR #315, opened 2020-06-11
POC - record hashes of CorDapp dependencies inside META-INF.

chrisr3:
Modify the `cordapp` plugin to record the SHA-256 hashes of everything added to Gradle's `cordapp` configuration into a `META-INF/Cordapp-Dependencies` file.

These SHA-256 hashes are equivalent to Corda attachment IDs, and therefore allow Corda to reassemble the exact set of compile-time attachment dependencies that this CorDapp used at build-time.


chrisr3:
Superseded.

PR #316, opened 2020-06-11
CORDA-3838: Make cordapp plugin properties lazy.

chrisr3:
Update the `cordapp` plugin to use Gradle's "lazy" property APIs. This fixes some technical debt, and should have no functional impact.

Note: Gradle considers a `Property` to be "present" if `get()` will return a value, _and this includes having a value set by `convention()`_. This doesn't include setting a `Provider` as a `convention()` if that `Provider` also has no value.

### DIFF
--- a/cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
+++ b/cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
@@ -26,6 +26,7 @@ class CordappTest {
     fun setup() {
         buildFile = testProjectDir.resolve("build.gradle")
         installResource(testProjectDir, "settings.gradle")
+        installResource(testProjectDir, "gradle.properties")
         installResource(testProjectDir, "repositories.gradle")
     }
 
@@ -47,6 +48,7 @@ class CordappTest {
         val jarTaskRunner = jarTaskRunner("CorDappWithInfo.gradle", extraArgs)
 
         val result = jarTaskRunner.build()
+        println(result.output)
 
         assertThat(result.task(":jar")!!.outcome).isEqualTo(TaskOutcome.SUCCESS)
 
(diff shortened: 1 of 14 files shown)


cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
```diff
@@ -47,6 +48,7 @@ class CordappTest {
         val jarTaskRunner = jarTaskRunner("CorDappWithInfo.gradle", extraArgs)
 
         val result = jarTaskRunner.build()
+        println(result.output)
```
palade, line 51: Maybe this `println` can be removed

cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
```diff
@@ -83,6 +85,7 @@ class CordappTest {
         val jarTaskRunner = jarTaskRunner("CorDappWithContractInfo.gradle", extraArgs)
 
         val result = jarTaskRunner.build()
+        println(result.output)
```
palade, line 88: This one as well

cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
```diff
@@ -119,6 +122,7 @@ class CordappTest {
         val jarTaskRunner = jarTaskRunner("CorDappWithWorkflowInfo.gradle", extraArgs)
 
         val result = jarTaskRunner.build()
+        println(result.output)
```
palade, line 125: and this one

cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
```diff
@@ -163,6 +167,7 @@ class CordappTest {
         val jarTaskRunner = jarTaskRunner("CorDappWithContractAndWorflowInfo.gradle", extraArgs)
 
         val result = jarTaskRunner.build()
+        println(result.output)
```
palade, line 170: this one and two more in lines 220 and 255

cordapp/src/test/kotlin/net/corda/plugins/CordappTest.kt
chrisr3: I added those `println()` deliberately so that the output from the test build becomes visible in the JUnit output :grinning:.

PR #318, opened 2020-06-18
NOTICK: Close input streams as JarFilter scans jar entries.

chrisr3:
Ensure that we close every `InputStream` created for the JAR file entries we are scanning. This is just good housekeeping, but otherwise we _could_ theoretically run out of file descriptors if required to scan a particularly large JAR.

### DIFF
--- a/jar-filter/src/main/kotlin/net/corda/gradle/jarfilter/JarFilterTask.kt
+++ b/jar-filter/src/main/kotlin/net/corda/gradle/jarfilter/JarFilterTask.kt
@@ -224,20 +224,21 @@ open class JarFilterTask @Inject constructor(objects: ObjectFactory, layouts: Pr
                 outJar.setComment(inJar.comment)
 
                 for (entry in inJar.entries()) {
-                    val entryData = inJar.getInputStream(entry)
-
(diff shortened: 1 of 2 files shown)


jar-filter/src/main/kotlin/net/corda/gradle/jarfilter/JarFilterTask.kt
```diff
@@ -224,20 +224,21 @@ open class JarFilterTask @Inject constructor(objects: ObjectFactory, layouts: Pr
...
-                        outJar.putNextEntry(entry.copy().withFileTimestamps(preserveTimestamps.get()))
-                        entryData.copyTo(outJar)
-                    } else {
-                        val classData = transform(entryData.readBytes())
-                        if (classData.isNotEmpty()) {
-                            // This entry's byte contents have almost certainly
-                            // changed, and will be stored compressed.
-                            outJar.putNextEntry(entry.asCompressed().withFileTimestamps(preserveTimestamps.get()))
-                            outJar.write(classData)
+                    inJar.getInputStream(entry).use { entryData ->
```
palade, line 227: Would inJar.getInputStream(entry)**.bufferedReader()**.use { .. } make sense in case of a large entry?

PR #321, opened 2020-06-25
EG-1982: <host:container> ssh port mapping in Dockerform

palade:
The purpose of this PR is to allow Dockerform to allow for <host:container> ssh port mapping. At the moment docker-compose will allocate an available port on the host machine when a Docker container is started. However, this has reduced user experience since it requires to check the port mapping before being able to ssh. The allocated port will be different from the one specified in the `docker-compose.yml` and `build.gradle` files.


palade:
@chrisr3  Would be possible to have the line for the port mapping without the single quotes?

```
  notary-service:
    volumes:
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/node.conf:/etc/corda/node.conf
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/certificates:/opt/corda/certificates
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/logs:/opt/corda/logs
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/persistence:/opt/corda/persistence
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/cordapps:/opt/corda/cordapps
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/network-parameters:/opt/corda/network-parameters
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/additional-node-infos:/opt/corda/additional-node-infos
    - C:\Users\Andrei Palade\Documents\code\corda-gradle-plugins-ssh\cordformation\build\junit1302795633451947092\build\nodes\NotaryService/drivers:/opt/corda/drivers
    environment:
    - ACCEPT_LICENSE=${ACCEPT_LICENSE}
    ports:
    - 10003
    - '"22234:22234"'      <- This should be "22234:22234"  without the single quotes
    image: corda/corda-zulu-4.0
```

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -100,7 +100,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
                             "$nodeBuildDir/drivers:/opt/corda/drivers"
                     ),
                     "environment" to listOf("ACCEPT_LICENSE=\${ACCEPT_LICENSE}"),
-                    "ports" to listOf(it.rpcPort.get(), it.config.getInt("sshd.port")),
+                    "ports" to listOf(it.rpcPort.get(), "\"${it.config.getInt("sshd.port")}:${it.config.getInt("sshd.port")}\""),
```
chrisr3, line 103: Why does the `ports` element suddenly need to be surrounded by quotes, i.e. the `\"` characters at the beginning and end?

> @chrisr3 Would be possible to have the line for the port mapping without the single quotes?

Funny you should ask.... :wink:

chrisr3:
>     - '"22234:22234"'      <- This should be "22234:22234"  without the single quotes

Are you _sure_ that you need the quotes at all?

palade:
> > ```
> > - '"22234:22234"'      <- This should be "22234:22234"  without the single quotes
> > ```
> 
> Are you _sure_ that you need the quotes at all?

According to [docker-compose documentation:
](https://docs.docker.com/compose/compose-file/)
> When mapping ports in the HOST:CONTAINER format, you may experience erroneous results when using a container port lower than 60, because YAML parses numbers in the format xx:yy as a base-60 value. For this reason, we recommend always explicitly specifying your port mappings as strings.

chrisr3:
I have spoken with _"The Wise Woman"_ (you probably haven't met her - she lives in Putney. Perhaps @fenryka can introduce you? :wink:). Anyway, her advice is to _**"Put quotes around everything in the Whole World!!!"**_

To that end, you can modify the `YAML_FORMAT_OPTIONS` thus:
```diff
diff --git a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
index 8bd91f479..73e1c8a5c 100644
--- a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
@@ -33,6 +33,7 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
         private val YAML_FORMAT_OPTIONS = DumperOptions().apply {
             indent = 2
             defaultFlowStyle = DumperOptions.FlowStyle.BLOCK
+            defaultScalarStyle = DumperOptions.ScalarStyle.SINGLE_QUOTED
         }
 
         private val YAML_MAPPER = Yaml(YAML_FORMAT_OPTIONS)
```
This may seem extreme, but it should automatically resolve other cases of "special characters" in the Dockerform going forward too,

You'd think YAML might have a way to set the "scalar type" to `SINGLE_QUOTED` _specifically_ for the `ports` tag, but I haven't found one yet.

chrisr3:
On the other hand, you can also try ignoring _"The Wise Woman"_ completely:
```diff
diff --git a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
index 8bd91f479..d4afdf283 100644
--- a/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
@@ -9,6 +9,7 @@ import org.gradle.api.tasks.*
 import org.gradle.api.tasks.PathSensitivity.RELATIVE
 import org.yaml.snakeyaml.DumperOptions
 import org.yaml.snakeyaml.Yaml
+import org.yaml.snakeyaml.nodes.Tag
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
@@ -35,7 +36,9 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
             defaultFlowStyle = DumperOptions.FlowStyle.BLOCK
         }
 
-        private val YAML_MAPPER = Yaml(YAML_FORMAT_OPTIONS)
+        private val YAML_MAPPER = Yaml(YAML_FORMAT_OPTIONS).apply {
+            addImplicitResolver(Tag("ports"), "\\d+:\\d+".toPattern(), null)
+        }
     }
 
     init {
```
The _**massive**_ problem here is that I don't have the faintest idea why or how this is working :exploding_head:!?

palade:
> ```diff
> addImplicitResolver
> ```


I was trying something like below, but this does not work. 
```
 class CustomRepresenter : Representer() {

        override fun representMapping(tag: Tag?, mapping: MutableMap<*, *>?, flowStyle: Boolean?): org.yaml.snakeyaml.nodes.Node {
            if (tag != null) {
                if (tag.startsWith("ports"))
                    setDefaultScalarStyle(DumperOptions.ScalarStyle.SINGLE_QUOTED)
            }
            return super.representMapping(tag, mapping, flowStyle)
        }
    }
```
I will use the addImplicitresolver approach as it does work

chrisr3:
> I will use the addImplicitresolver approach as it does work

What _seems_ to be happening is that YAML is recognising the `<portA>:<portB>` value as a `Tag.STR` which isn't supposed to be handled as `Tag.STR` because it's been mapped to a custom `Tag("ports")` :thinking:.

palade:
> > I will use the addImplicitresolver approach as it does work
> 
> What _seems_ to be happening is that YAML is recognising the `<portA>:<portB>` value as a `Tag.STR` which isn't supposed to be handled as `Tag.STR` because it's been mapped to a custom `Tag("ports")` 🤔.

I think does overwrite (the solver) the default representation style when it finds a particular node with that tag which is what we want in this case

chrisr3:
In the YML `Serializer` class:
```java
    case scalar:
        ScalarNode scalarNode = (ScalarNode) node;
        Tag detectedTag = this.resolver.resolve(NodeId.scalar, scalarNode.getValue(), true);
        Tag defaultTag = this.resolver.resolve(NodeId.scalar, scalarNode.getValue(), false);
        ImplicitTuple tuple = new ImplicitTuple(node.getTag().equals(detectedTag), node
                .getTag().equals(defaultTag));
        ScalarEvent event = new ScalarEvent(tAlias, node.getTag().getValue(), tuple,
                scalarNode.getValue(), null, null, scalarNode.getStyle());
        this.emitter.emit(event);
        break;
```
The `addImplicitResolver()` is changing the `ImplicitTuple` value in our particular case so that:
- `detectedTag` is our custom "ports" `Tag`
- `defaultTag` is `Tag.STR`
- `node.getTag()` is `Tag.STR`

which gives us
```kotlin
ImplicitTuple(plain = false, nonPlain = true)
```

cordformation/src/test/resources/net/corda/plugins/DeployTwoNodeCordappWithDocker.gradle
```diff
@@ -29,6 +29,7 @@ task prepareDockerNodes(type: net.corda.plugins.Dockerform, dependsOn: ['jar'])
             address("localhost:10003")
             adminAddress("localhost:10043")
         }
+        sshdPort 22234
```
chrisr3, line 32: Is there any way that you can test what has been written into the YAML please? I am worried that we have strayed into "undocumented behaviour" territory and so want to keep a close eye on the output :worried:.

You should be able to find the `docker-compose.yml` file relative to `testProjectDir`.

palade: Extended one of the unit tests to check if the ports are correctly set

palade:
> Looking OK, but we need to be _sure_ the YAML has been written correctly.

Moved to the custom representer approach and moved this to a DockerformUtils class seems

cordformation/src/main/kotlin/net/corda/plugins/DockerformUtils.kt
```diff
@@ -0,0 +1,41 @@
+package net.corda.plugins
+
+import org.yaml.snakeyaml.DumperOptions
+import org.yaml.snakeyaml.Yaml
+import org.yaml.snakeyaml.nodes.Tag
+import org.yaml.snakeyaml.representer.Represent
+import org.yaml.snakeyaml.representer.Representer
+import java.io.InputStream
+
+class DockerformUtils {
```
chrisr3, line 10: If this approach is more reliable / **documented** than the `addImplicitResolver()` approach then fair enough. But in that case we don't need to create a `DockerformUtils` wrapper class.

cordformation/src/main/kotlin/net/corda/plugins/DockerformUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+                val str = data as QuotedString
+                return representScalar(Tag.STR, str.value, DumperOptions.ScalarStyle.DOUBLE_QUOTED.char)
+            }
+        }
+
+        init {
+            this.representers[QuotedString::class.java] = RepresentQuotedString()
+        }
+    }
+}
```
chrisr3, line 41: I'd move all this back into `Dockerform.kt`. You can put the `QuotedString` and `DockerComposeRepresenter` classes at the bottom of the file and declare them both as `private class`, which means that they are "package private" in Java terms. (`internal` scope only means something to Kotlin.)

cordformation/src/main/kotlin/net/corda/plugins/DockerformUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        private val YAML_MAPPER = Yaml(DockerComposeRepresenter(), YAML_FORMAT_OPTIONS)
+
+        fun dump(data: Any): String =  YAML_MAPPER.dump(data)
+
+        fun load(io: InputStream): Map<Any, Any> = YAML_MAPPER.load(io)
+    }
+
+    internal class QuotedString(var value: String)
```
chrisr3, line 26: `QuotedString` should use `val` instead of `var`.

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -132,5 +133,24 @@ class DockerformTest : BaseformTest() {
...
+        assertThat(yaml.containsKey("services")).isTrue()
+
+        val services = yaml["services"] as Map<Any, Any>
+        assertThat(services.containsKey("notary-service")).isTrue()
+        assertThat(services.containsKey("bankofcorda")).isTrue()
+
+        val notaryService = services["notary-service"] as Map<Any, Any>
+        assertThat(notaryService.containsKey("ports")).isTrue()
+        val notaryServicePorts = notaryService["ports"] as List<Any>
+        assertThat(notaryServicePorts.contains("22234:22234")).isTrue()
```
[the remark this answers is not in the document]
palade: Please note that `notaryServicePorts` is an array not a string

cordformation/src/main/kotlin/net/corda/plugins/DockerformUtils.kt
palade: Removed DockerformUtils and put everything in back in Dockerform

cordformation/src/main/kotlin/net/corda/plugins/DockerformUtils.kt
palade: Changed to `val`

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -132,5 +133,24 @@ class DockerformTest : BaseformTest() {
         assertTrue(bankOfCordaConfig.hasPath("rpcSettings.adminAddress"))
         assertEquals(10006, ConfigurationUtils.parsePort(bankOfCordaConfig.getString("rpcSettings.address")))
         assertEquals(10046, ConfigurationUtils.parsePort(bankOfCordaConfig.getString("rpcSettings.adminAddress")))
+
+        val dockerComposePath = getDockerCompose()
+
+        val yaml = DockerformUtils.load(dockerComposePath.toFile().inputStream());
```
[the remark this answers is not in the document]
palade: Replaced with a standard YAML project without any required formatting

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -132,5 +134,24 @@ class DockerformTest : BaseformTest() {
         assertTrue(bankOfCordaConfig.hasPath("rpcSettings.adminAddress"))
         assertEquals(10006, ConfigurationUtils.parsePort(bankOfCordaConfig.getString("rpcSettings.address")))
         assertEquals(10046, ConfigurationUtils.parsePort(bankOfCordaConfig.getString("rpcSettings.adminAddress")))
+
+        val dockerComposePath = getDockerCompose()
+
+        val yaml = Yaml().load(dockerComposePath.toFile().bufferedReader()) as Map<Any, Any>
```
chrisr3, line 140: I think this should be `Map<String, Any>`

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -132,5 +134,24 @@ class DockerformTest : BaseformTest() {
...
+        assertThat(yaml.containsKey("services")).isTrue()
+
+        val services = yaml["services"] as Map<Any, Any>
+        assertThat(services.containsKey("notary-service")).isTrue()
+        assertThat(services.containsKey("bankofcorda")).isTrue()
```
chrisr3, line 145: AssertJ is very flexible in its assertions, and it understands both maps and lists. Hence:
```kotlin
assertThat(services).containsKey("notary-service")
```

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -132,5 +134,24 @@ class DockerformTest : BaseformTest() {
...
+
+        val notaryService = services["notary-service"] as Map<Any, Any>
+        assertThat(notaryService.containsKey("ports")).isTrue()
+        val notaryServicePortsList = notaryService["ports"] as List<Any>
+        assertThat("22234:22234" in notaryServicePortsList).isTrue()
```
chrisr3, line 150: And similarly...
```kotlin
assertThat(notaryServicePortsList).contains("22234:22234")
```

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -132,5 +134,24 @@ class DockerformTest : BaseformTest() {
...
+
+        val bankOfCorda = services["bankofcorda"] as Map<Any, Any>
+        assertThat(bankOfCorda.containsKey("ports")).isTrue()
+        val bankOfCordaPortsList = bankOfCorda["ports"] as List<Any>
+        assertThat("22235:22235" in bankOfCordaPortsList).isTrue()
     }
```
chrisr3, line 156: Etc... :wink:

PR #322, opened 2020-06-25
CORDA-3863: Refresh certificate for development CorDapp signing key.

chrisr3:
The current CorDapp signing key expired on 19th November 2019, so create a new one that will expire in 2030. Also include the script that generated it (mainly for reference).


cordapp/src/main/scripts/generate-certificate.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
```
szymonsztuka, line 2: Maybe add explanation ? :
```suggestion
# For reference, this script was used to generate the  CorDapp signing key.
```

mnesbit:
Pretty sure we have to preserve the keypair, but generate a new cert

PR #330, opened 2020-07-03
CORDA-3884: Use Gradle's task configuration avoidance APIs for cordapp.

chrisr3:
This is not a user-visible change, unless you perform a build scan on your Gradle project. The `configureEach {}` API was introduced in Gradle 4.9.

### DIFF
--- a/cordapp/src/main/kotlin/net/corda/plugins/CordappUtils.kt
+++ b/cordapp/src/main/kotlin/net/corda/plugins/CordappUtils.kt
@@ -3,6 +3,8 @@ package net.corda.plugins
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.ConfigurationContainer
+import org.gradle.api.plugins.JavaPlugin.COMPILE_CONFIGURATION_NAME
+import org.gradle.api.plugins.JavaPlugin.RUNTIME_CONFIGURATION_NAME
 import java.nio.file.Files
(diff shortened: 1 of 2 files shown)


cordapp/src/main/kotlin/net/corda/plugins/CordappUtils.kt
```diff
@@ -23,11 +25,13 @@ class CordappUtils {
...
 
         fun createCompileConfiguration(name: String, configurations: ConfigurationContainer): Configuration {
-            return createChildConfiguration(name, configurations.single { it.name == "compile" }, configurations)
+            @Suppress("deprecation")
+            return createChildConfiguration(name, configurations.single { it.name == COMPILE_CONFIGURATION_NAME }, configurations)
         }
 
         fun createRuntimeConfiguration(name: String, configurations: ConfigurationContainer): Configuration {
-            return createChildConfiguration(name, configurations.single { it.name == "runtime" }, configurations)
+            @Suppress("deprecation")
```
palade, line 33: The `@Suppress` element was not here before the change. Was it?

chrisr3: No, it wasn't here before. However, I'm now using Gradle's `RUNTIME_CONFIGURATION_NAME` identifier, which is marked as "deprecated" in the Javadoc because Gradle's "runtime" configuration _is_ deprecated. Hence I need to suppress the warning for referencing a deprecated field.

PR #331, opened 2020-07-06
INFRA-446 Add CODEOWNERS file

rnicoll:
Add initial code owners for Gradle plugins


rnicoll:
Ah don't worry, it doesn't require you yet, this is just saying it's all yours.

PR #335, opened 2020-07-08
CORDA-3902: Update publish-utils to be a semi-respectable Gradle citizen.

chrisr3:
The `publish-utils` plugin is one of our earliest Gradle plugins that helps configure  `maven-publish` for the sake of Artifactory and Bintray. Unfortunately, its requirement to configure the Bintray plugin has resulted in some "questionable" yet difficult to fix code that has not been maintained as Gradle itself has evolved.

Bring `publish-utils` as close to "up-to-date" as its dependence on the Bintray plugin currently allows.

Tested _locally_  by setting `bintrayConfig.dryRun=true` and then executing:
```
$ ./gradlew install
$ ./gradlew bintrayUpload
$ ./gradlew artifactoryPublish
```


roastario:
👀 - looking through

chrisr3:
> All looks very sensible - only thing I'd ask, is there not an existing maven model library out there for writing maven plugins we could use instead of recreating parts of the Pom model as pojos?

Well, it would have to integrate with Gradle too, although tbh the XML part of the code hasn't really changed here - I just moved it somewhere else to make Gradle happier.

PR #348, opened 2020-08-14
Add gradle extension to Dockerform

J-Gerakis:
EG-849 Dockerform enhancement

A new class (External) hold configuration for an external service described in external-service.gradle and the formatting is handled in the Dockerform class

I also created 2 extra unit test and test data.

Note: there's an existing, failing test (CordformTest -> a node that requires an extra command to create schema()) that will need to be addressed properly


cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +205,28 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+                extra["environment"] = external.environment.toList().map { "${it.first}=${it.second}" }
+            }
+            if(external.containerName.isNotEmpty()){
+                extra["container_name"] = external.containerName
+            }
+            if(external.privileged) {
+                extra["privileged"] = true
+            }
+            if(external.commands.isNotEmpty()){
+                extra["command"] = "bash -c \"${external.commands}\""
```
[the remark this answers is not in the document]
J-Gerakis: As far as I can tell, Docker images are unix-based (I'm not a Docker expert, this just from my own limited experience with them) Containers only accept only one command, so if you have two or more, you need to "cheat" by chaining them like this: 
> cd home/app/ && java -jar blah.jar 

If there's only one command, then yes, the bash -c part is extraneous. If you have chain, you need to have it (again, empirical experience from my part)

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -69,6 +69,10 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
     @get:Input
     val dockerImage: Property<String> = objects.property(String::class.java)
 
+    @get:Optional
+    @get:Input
+    var external: External = objects.newInstance(External::class.java)
```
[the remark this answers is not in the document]
J-Gerakis: Agree on the `@optional` this is unnecessary. But it needs to be a `var` and not a `val` or else the affectation in the gradle `prepareDockerNodes` will fail.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,40 @@
...
+    @get:Input
+    var commands: String = ""
+
+    @get:Optional
+    @get:Input
+    var privileged: Boolean = false
+
+    fun isEmpty(): Boolean {
+        return volumes.isEmpty() && containerImage.isEmpty()
+    }
```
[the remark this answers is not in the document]
J-Gerakis: My reasoning for the existence of this method was that without volumes or image, nothing can be done and there no point on processing further. Those two are the minimum needed for a docker-compose file to be viable.

However, now that I'm thinking about it, volumes may not even be needed if the user comes with a custom image already containing all that is need... So I should replace this isEmpty() method with just External.containerImage.isEmpty()

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
+package net.corda.plugins;
+
+import org.gradle.api.tasks.Input
+
+open class External {
+
+    @get:Input
```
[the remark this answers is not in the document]
J-Gerakis: Those are elements that, while not needed 100% of the time, are required often enough. Not every user is going to come in with a customized image, I think a lot re going to tinker with out of the hub standard image.

With a web container like tomcat, you need ports and volumes. If you want to run a .jar, you need command and ports. Only "privileged" is a bit of a special case, but since we know it happens, why not have it?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -69,6 +69,12 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
     @get:Input
     val dockerImage: Property<String> = objects.property(String::class.java)
 
+    /**
+     * External service config
+     */
+    @get:Nested
+    var external: External = objects.newInstance(External::class.java)
```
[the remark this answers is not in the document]
J-Gerakis: That's not necessary. You can use whatever name you want for your credentials variables. Like node1-pwd, node1-user, node2-pwd, node2-user, etc.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
...
+    var containerName: String = "external-service"
+
+    @get:Input
+    var servicePorts: List<Int> = listOf(8080)
```
chrisr3, line 11: This becomes:
```kotin
 @get:Input
val servicePorts: ListProperty<Int> = objects.listProperty(Int::class.java).convention(listOf(8080))
```
Again, it doesn't need to be `@Optional`. However, I don't think a `ListProperty` needs to be `@Optional` anyway because its default value is an empty list.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @get:Input
+    var containerImage = ""
```
chrisr3, line 14: I think this one should be  an `@Optional` `Property<String>` without a convention.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @get:Input
+    var volumes: List<Map<String, String>> = emptyList()
```
chrisr3, line 17: This is tricky - I suspect it should look like this:
```kotlin
@Suppress("unchecked_cast")
@get:Input
val volumes: ListProperty<Map<String, String>> = objects.listProperty(Map::class.java) as ListProperty<Map<String, String>>
```
Again, this `ListProperty` is empty by convention and so doesn't need to be declared as `@Optional`.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @get:Input
+    var environment: Map<String, Any> = emptyMap()
```
chrisr3, line 20: This should be a `MapProperty`, which is also automatically empty by default and so doesn't need to be `@Optional`.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @get:Input
+    var commands: String = ""
```
chrisr3, line 23: This `String` property doesn't need a default value and so _can_ be declared as `@Optional`.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @get:Input
+    var privileged: Boolean = false
```
chrisr3, line 26: This `Boolean` property has a convention of `false` and so doesn't need to be `@Optional`.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +207,30 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
             services[it.containerName] = service
         }
 
+        if (external.containerImage.isNotEmpty()) {
```
chrisr3, line 210: If `containerImage` is an `@Optional` property then this line would become:
```kotlin
if (external.containerImage.isPresent()) {
```

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +207,30 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            val extra = mutableMapOf(
+                    "image" to external.containerImage,
+                    "ports" to external.servicePorts.toList().map{QuotedString("${it}:${it}")},
+                    "expose" to external.servicePorts
+            )
```
chrisr3, line 215: You need to invoke `get()` to retrieve a lazy property's value. And I think the `toList()` is unnecessary. So this would become:
```kotlin
val extra = mutableMapOf(
    "image" to external.containerImage.get(),
    "ports" to external.servicePorts.get().map {QuotedString("${it}:${it}") },
    "expose" to external.servicePorts.get()
)
```

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +207,30 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            if(external.volumes.isNotEmpty()){
+                extra["volumes"] = external.volumes.toList().map{ "${it["sourceFile"]}:${it["deploymentPath"]}" }
+            }
```
chrisr3, line 218: The `volumes` property is a `ListProperty` and so has a default value of an empty list. This means that `.isPresent()` will always return `true`, and so you can do this:
```kotlin
val volumes = external.volumes.get()
if (volumes.isNotEmpty()) {
   ...
}
```
Again, you don't need `toList()` here.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +207,30 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            if(external.environment.isNotEmpty()){
+                extra["environment"] = external.environment.toList().map { "${it.first}=${it.second}" }
+            }
```
chrisr3, line 221: `environment` is a `MapProperty` and so also has an empty default value.
```kotlin
val environment = external.environment.get()
if (environmen.isNotEmpty()) {
   ...
}
```
Superfluous `toList()` again. This should be enough:
```kotlin
environment.map { "${it.key}=${it.value}" }
```

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +207,30 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            if(external.containerName.isNotEmpty()){
+                extra["container_name"] = external.containerName
+            }
```
chrisr3, line 224: `containerName` has a convention of `external-service`, so do you need to check whether it has a value? You seem to be assuming it _always_ has a value here:
```kotlin
services["${external.containerName}"] = extra
```
Shouldn't the expectation be consistent?

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +207,30 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            if(external.privileged) {
+                extra["privileged"] = true
+            }
+            if(external.commands.isNotEmpty()){
+                extra["command"] = external.commands
+            }
+            services["${external.containerName}"] = extra
```
chrisr3, line 231: You will need to use `.isPresent()` and `.get()` correctly here.

cordformation/src/test/kotlin/net/corda/plugins/BaseformTest.kt
```diff
@@ -43,6 +43,8 @@ open class BaseformTest {
         installResource("repositories.gradle")
         installResource("gradle.properties")
         installResource("postgres.gradle")
+        installResource("external-service.gradle")
+        installResource("external-service-2.gradle")
```
chrisr3, line 47: Are you replicating the pattern we used for `postgres.gradle`? The external database configuration has _a lot_ of complicated elements that is still largely "boilerplate", and so we separated out all of the "database dependent" parts into a script plugin. But I don't think we need this level of complexity for your `external` block.

cordformation/src/test/resources/external-service-2.gradle
```diff
@@ -0,0 +1,5 @@
+ext {
+    externalServiceConfig = [
+            containerImage: "docker/test/customTomcat"
+    ]
+}
```
chrisr3, line 5: You don't need this file.You can apply this external configuration _directly_ into `DeployTwoNodeCordappWithExternalServiceNoOption.gradle`

cordformation/src/test/resources/external-service.gradle
```diff
@@ -0,0 +1,23 @@
...
+            ],
+            environment: [
+                    "rpcUsername" : "user1",
+                    "rpcPassword" : "pass2"
+            ],
+            commands: "bash -c \"cd home && java -jar CordaDevTestAPI.jar\"",
+            privileged: true
+
+    ]
+}
```
chrisr3, line 23: You don't need this file either. Please apply these settings directly into `DeployTwoNodeCordappWithExternalService.gradle`.

chrisr3 (changes requested):
These are the rest of my "Gradle mechanics" comments. But we need to put Gradle to one side and ask ourselves:
> "Does this DSL make sense for what I am allowing CorDapp developers to do?"

because Gradle is fundamentally just a _means_ to an end, and not an end in itself.

A developer will arrive here with preconceived ideas of what they want to achieve, and the configuration they will need to provide - _and not provide!_ - to do it. Does this DSL align with those expectations? (This is a UI design question, not a criticism. I'm not really familiar with Corda's Docker deployments.)

cordformation/src/test/kotlin/net/corda/plugins/BaseformTest.kt
J-Gerakis: Initial story requirement called for a separate file. While it's true the external block will be, on average, rather short, there's some value in keeping it separated, in my opinion.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
J-Gerakis: Yeah, I just stuttered here.

cordformation/src/test/kotlin/net/corda/plugins/BaseformTest.kt
chrisr3: > Initial story requirement called for a separate file. While it's true the external block will be, on average, rather short, there's some value in keeping it separated, in my opinion.

I'm not convinced that this separate file is going to be compatible with using Gradle's lazy APIs. In which case the separate file is going to have to lose, I'm afraid.

J-Gerakis: From what I'm testing right now, it seems indeed incompatible with Gradle lazy init.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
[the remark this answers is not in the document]
J-Gerakis: What exactly is the benefit of lazy properties here? It increases significantly the code complexity.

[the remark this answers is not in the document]
chrisr3: > What exactly is the benefit of lazy properties here? It increases significantly the code complexity.

That's Gradle for you! :grin:

A lazy property is not evaluated until the configuration phase is complete, where the configuration phase includes applying any third party plugins and parsing the contents of the `build.gradle` file. Not using lazy properties can lead to situations where tasks don't execute with the configuration values that you expect.

I know this is fiddlier than using POJOs, but _this is how Gradle plugins are now expected to work._ The `cordformation` plugin may be one of the last hold-outs of obsolete Gradle practices, but that is no excuse to make the problem even worse.

J-Gerakis:
@chrisr3 : I followed your recommendations and reworked it to use gradle lazy properties. I also removed external files to move the declaration in the build.gradle I'm not 100% sure lazy properties are incompatible with external files, maybe there's a workaround, but nothing I could find in a reasonable amount of time.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +212,29 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
             services[it.containerName] = service
         }
 
+        if (external.containerImage.isPresent()) {
```
chrisr3, line 215: You have given the `containerImage` property a convention of `""`. Gradle will therefore _always_ consider this property to be present. I think what you need to do here is remove the convention and declare the property as `@Optional`.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +212,29 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
...
+            if(external.volumes.get().isNotEmpty()){
+                extra["volumes"] = external.volumes.get().map{ "${it["sourceFile"]}:${it["deploymentPath"]}" }
+            }
+            if(external.environment.get().isNotEmpty()){
+                extra["environment"] = external.environment.get().map { "${it.key}=${it.value}" }
+            }
+            if(external.privileged.get() == true) {
+                extra["privileged"] = external.privileged.get()
+            }
+            if(external.commands.get().isNotEmpty()){
```
chrisr3, line 232: I would remove the convention from `commands` too, and then say
```kotlin
if (external.commands.isPresent()) {
    extra["command"] = external.commands.get()
}
```
And then annotate `commands` as `@Optional`, of course.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,34 @@
...
+open class External @Inject constructor(objects: ObjectFactory) {
+
+    @get:Input
+    val containerName: Property<String> = objects.property(String::class.java).convention("external-service")
+
+    @get:Input
+    val servicePorts: ListProperty<Int> = objects.listProperty(Int::class.java).convention(listOf(8080))
+
+    @get:Input
+    val containerImage: Property<String> = objects.property(String::class.java).convention("")
```
chrisr3, line 19: Remove convention and make `@Optional`.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    @Suppress("unchecked-cast")
+    @get:Input
+    val volumes: ListProperty<Map<String, String>> = objects.listProperty(Map::class.java) as ListProperty<Map<String, String>>
+
+    @get:Input
+    val environment: MapProperty<String, Any> = objects.mapProperty(String::class.java, Any::class.java)
+
+    @get:Input
+    val commands: Property<String> = objects.property(String::class.java).convention("")
```
chrisr3, line 29: Remove convention and make `@Optional`.

chrisr3 (changes requested):
Almost there, just a few small things :grinning:.

cordformation/src/main/kotlin/net/corda/plugins/External.kt
```diff
@@ -0,0 +1,34 @@
+package net.corda.plugins;
```
chrisr3, line 1: Kotlin doesn't need this semi-colon.

chrisr3 (changes requested):
Just being _really_ picky... :wink:

J-Gerakis:
@chrisr3 I think I'm getting somewhere here. The reason I was reluctant to use Optional for the containerImage is that it seems counter-intuitive at first: it's the only vital part without which this whole External thing is pointless. So I added a comment in the code to explain why we use optional here.

chrisr3 (approved):

chrisr3:
:tada: LGTM :tada: The Gradle is fine, I'm hoping the Docker API aspect is OK too, @palade ?

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -157,4 +157,87 @@ class DockerformTest : BaseformTest() {
...
+
+        val volumes = external["volumes"] as List<String>
+        assertThat(volumes).contains("C:\\Projects\\gs-rest-service\\target\\CordaDevTestAPI-0.0.1-SNAPSHOT.jar:/home/CordaDevTestAPI.jar")
+
+        val environment = external["environment"] as List<String>
+        assertThat(environment).contains("rpcUsername=user1")
+        assertThat(environment).contains("rpcPassword=pass2")
+
+        val ports = external["ports"] as List<String>
+        assertThat(ports).contains("8080:8080")
```
palade, line 196: This line shows twice `assertThat(ports).contains("8080:8080")`

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
```diff
@@ -157,4 +157,87 @@ class DockerformTest : BaseformTest() {
...
+        assertThat(ports).contains("8000:8000")
+
+        val exposedPort = external["expose"] as List<Int>
+        assertThat(exposedPort).contains(8080)
```
palade, line 200: This line shows twice `assertThat(exposedPort).contains(8080)`

cordformation/src/test/resources/net/corda/plugins/DeployTwoNodeCordappWithExternalService.gradle
```diff
@@ -0,0 +1,67 @@
...
+        sshdPort 22235
+    }
+
+    external {
+        containerName = "example-service"
+        containerImage = "docker.io/bitnami/java:latest"
+        servicePorts = [8080,8000]
+        volumes =
+        [
+                [
```
palade, line 52: I would change `volumes` to be an array of elements of type `volume`, instead of an array of arrays.

palade (changes requested):
I left a few comments, but other than that this LGTM!

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
J-Gerakis: Not quite! one is for 8080, the other for 8000.

cordformation/src/test/kotlin/net/corda/plugins/DockerformTest.kt
J-Gerakis: Same as above.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
```diff
@@ -201,6 +212,29 @@ open class Dockerform @Inject constructor(objects: ObjectFactory) : Baseform(obj
             services[it.containerName] = service
         }
 
+        if (external.containerImage.isPresent()) {
+
+            val extra = mutableMapOf(
+                    "container_name" to external.containerName.get(),
+                    "image" to external.containerImage.get(),
+                    "ports" to external.servicePorts.get().map {QuotedString("${it}:${it}") },
+                    "expose" to external.servicePorts.get()
```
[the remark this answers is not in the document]
J-Gerakis: Not always enough. This is a double guarantee.

cordformation/src/test/resources/net/corda/plugins/DeployTwoNodeCordappWithExternalService.gradle
J-Gerakis: Really? For a pair of Strings, it seems a bit like over-engineering to me.

cordformation/src/main/kotlin/net/corda/plugins/Dockerform.kt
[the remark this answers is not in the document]
palade: Can you give more details?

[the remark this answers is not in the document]
J-Gerakis: I'm trying to reproduce a situation where this is necessary (I think it happened with Tomcat?) If I can't, then I'll remove it.

J-Gerakis:
@palade I decided to remove 'expose'. I don't think this is really necessary in most cases.

palade (dismissed):
This LGTM!

palade:
> This LGTM!

Also will need to edit the `changelog.md` to add a summary of the change.

JamesHR3 (dismissed):
Good work!

changelog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 5
 
+### Version 5.0.14
+
+* `dockerform`: Parsing for adding external service to docker-compose
```
chrisr3, line 7: The current release is only 5.0.12, and so this change will become part of 5.0.13.

chrisr3 (changes requested):
Oops, wrong version.

chrisr3 (approved):
:tada: LGTM :tada: :grinning:

PR #355, opened 2020-10-17
CORDA-4060: Upgrade JarFilter to use ASM 8.0.1.

chrisr3:
ASM 8 includes support for JDK11's nest host and nest member byte-code elements.


palade (approved):
LGTM!

PR #357, opened 2020-10-24
CORDA-4066: Check that Gradle is compatible when each plugin is applied.

chrisr3:
Use Gradle's built-in version APIs to ensure that each plugin is compatible with the version of Gradle it is used with.

Also, do _not_ assume that Gradle properties can always be type-cast safely to `String` because properties could also be Groovy `GString`s. Use `toString()` to convert safely to `String` instead.


api-scanner/src/main/java/net/corda/plugins/apiscanner/ApiScanner.java
```diff
@@ -8,17 +9,19 @@
...
 import org.gradle.jvm.tasks.Jar;
+import org.gradle.util.GradleVersion;
 
 import javax.annotation.Nonnull;
 import java.util.Objects;
 
 import static org.gradle.api.plugins.JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME;
 
-@SuppressWarnings("unused")
+@SuppressWarnings({"unused", "UnstableApiUsage"})
```
palade, line 19: What is generating this warning?

palade (approved):
LGTM!

api-scanner/src/main/java/net/corda/plugins/apiscanner/ApiScanner.java
chrisr3: Gradle's `ObjectFactory` and "lazy" Gradle `Property`/`Provider` classes are all considered "experimental" in Gradle 5.6.

PR #361, opened 2020-12-18
Javac quasar plugin

woggioni:
This is a `javac` (the java compiler) plugin to ease detection of erroneous quasar usage, particularly calling a suspendable method 
from the body of a non suspendable method (which is a common cause of hard-to-troubleshoot runtime errors).

A Java method is deemed suspendable if 
- it is annotated with the `@Suspendable` annotation
- it is declared to throw `SuspendExecution`
- it is mentioned in any `META-INF/suspendables` file available in the compilation classpath
- it overrides another method mentioned in any `META-INF/suspendables` file available in the compilation classpath

## Usage

Simply apply the `quasar-utils` gradle plugin to your project and activate the java compiler plugin with

```groovy
quasar {
    enableJavacPlugin = true
}
```

From this moment, trying to compile a java source file like 
```java
import co.paralleluniverse.fibers.Suspendable;

interface Bar {
    @Suspendable
    void bar();
}

public class Foo {
    void foo(Bar bar) {
        bar.bar();
    }
}
```

will fail with

```
/home/user/code/sample-project/src/main/java/Foo.java:10: error: Invocation of suspendable method from non suspendable method
        bar.bar();
           ^
```

### Build

This plugin uses internal compiler API (it would be great to just rely on the public API, but the offered 
functionality is simply not enough) which is not guaranteed to be stable across Java version bumps.
there have been indeed small changes between version 8 and version 11 that are dealt with in 
`net.corda.plugins.javac.quasar.ASTUtils`. 
Despite this, this project builds (or at least it is supposed to) on both JDK 8 and JDK 11, 
the generated artifact only contains Java 8 bytecode and works on both JDK 8 and JDK 11 compiler 
(and hopefully even future Java compiler versions, but who knows).

### Supported compilers

This plugin has been successfully tested with

- Zulu 8
- Corretto 8
- OpenJdk 8 (OpenJ9)
- OpenJdk 8 (Hotspot)
- Corretto 11
- OpenJdk 11 (OpenJ9)
- OpenJdk 11 (Hotspot)


quasar-utils/javac-plugin/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+    mavenCentral()
+}
+
+dependencies {
+
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: junit_jupiter_version
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params', version: junit_jupiter_version
+    testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: junit_jupiter_version
+
+    if(JavaVersion.current() <= JavaVersion.VERSION_1_8) {
```
chrisr3, line 16:
```groovy
if (JavaVersion.current().java8Compatible) {
   ...
}
```

quasar-utils/javac-plugin/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+        compileOnly files("${System.properties['java.home']}/../lib/tools.jar")
+        testRuntimeOnly files("${System.properties['java.home']}/../lib/tools.jar")
```
chrisr3, line 18: How does this interact with Gradle's new [Java toolchains](https://docs.gradle.org/current/userguide/toolchains.html)?

quasar-utils/javac-plugin/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+test {
+    useJUnitPlatform()
+    def testCompilationClassPath = sourceSets["main"].output.classesDirs.files +
+            sourceSets["main"].runtimeClasspath.files +
+            sourceSets["test"].resources.srcDirs
+    systemProperty("test.compilation.classpath",
+            String.join(File.pathSeparator, testCompilationClassPath.collect {it.toString() }))
```
chrisr3, line 46: Maybe:
```
systemProperty 'test.compilation.classpath', testCompileClasspath.asPath
```

quasar-utils/javac-plugin/src/java11/java/net/corda/plugins/javac/quasar/ASTUtils.java
```diff
@@ -0,0 +1,11 @@
+package net.corda.plugins.javac.quasar;
+
+import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.util.Filter;
+import com.sun.tools.javac.util.Name;
```
chrisr3, line 5: According to [the documentation](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.compiler/com/sun/tools/javac/package-summary.html), the `com.sun.tools` package is "legacy":
> This package provides a legacy entry point for the javac tool. See the jdk.compiler module for details on replacement APIs.

Using this seems like a Bad Idea :tm:.

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/SuspendExecution.java
```diff
@@ -0,0 +1,3 @@
+package co.paralleluniverse.fibers;
+
+public class SuspendExecution extends Exception {}
```
chrisr3, line 3: This is a Quasar class is a Quasar package. Why is it here? ("Split packages" being Very Naughty these days, etc.)

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
```diff
@@ -0,0 +1,9 @@
+package co.paralleluniverse.fibers;
+
+import java.lang.annotation.*;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+public @interface Suspendable {
+}
```
chrisr3, line 9: This is another Quasar class. Are you also aware that we have recently modified Quasar so that Corda can declare its own `@Suspendable` annotation? CorDapp developers are not going to be using this annotation any more.

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/ASTUtils.java
```diff
@@ -0,0 +1,35 @@
...
+public class ASTUtils {
+    private static final Method getTypeMembers;
+    private static final Method filterClassMembers;
+
+    static  {
+        try {
+            if (Float.parseFloat(System.getProperty("java.specification.version")) > 8) {
+                filterClassMembers = Scope.class.getMethod("getSymbolsByName", Name.class, Filter.class);
+            } else {
+                filterClassMembers = Scope.class.getMethod("getElementsByName", Name.class, Filter.class);
```
chrisr3, line 20: This doesn't look brittle at all...

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/SuspendableChecker.java
```diff
@@ -0,0 +1,207 @@
...
+                    .anyMatch(s -> inSuspendablesFile(s) || superSuspendable(s));
+        }
+
+        private boolean inSuspendablesFile(Symbol sym) {
+            MethodSignature signature = MethodSignature.from(sym);
+            return suspendableDatabase.isSuspendable(signature.className, signature.methodName, signature.desc);
+        }
+
+        private boolean isSuspendable(Symbol sym) {
+            Annotation annotation = sym.getAnnotation(Suspendable.class);
```
chrisr3, line 182: This isn't going to work with a configurable `@Suspendable` annotation. (I'm also worried that it's actually referring to your "doppelganger" of Quasar's `@Suspendable` annotation, and not to Quasar's own one at all. What happens when Quasar is actually present on the compile classpath?

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/SuspendableDatabase.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Set;
+
+public class SuspendableDatabase {
```
chrisr3, line 13: This class has been derived from Quasar's `co.paralleluniverse.fibers.instrument.SimpleSuspendableClassifier`, but has not preserved Ron Pressler's copyright header. This looks like a licence violation to me.

quasar-utils/javac-plugin/src/test/resources/net/corda/plugins/javac/quasar/test/TestCase1.java
```diff
@@ -0,0 +1,13 @@
+package net.corda.plugins.javac.quasar.test;
+
+import java.util.function.Consumer;
+import co.paralleluniverse.fibers.Suspendable;
+import co.paralleluniverse.fibers.SuspendExecution;
+
+public class TestCase1 {
```
chrisr3, line 7: Not a very informative set of class names. What is each case testing?

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
```diff
@@ -25,6 +29,7 @@ class QuasarPlugin implements Plugin<Project> {
 
     private static final String QUASAR = "quasar"
     private static final String MINIMUM_GRADLE_VERSION = "5.1"
+    private static final String JAVAC_PLUGIN = "javac_plugin"
```
chrisr3, line 32: _Every_ Java Compiler plugin could be called `javac_plugin`. What is so special about this one?

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
```diff
@@ -85,6 +95,26 @@ class QuasarPlugin implements Plugin<Project> {
...
+                            "/META-INF/MANIFEST.MF"
+                    Manifest manifest = new Manifest(new URL(manifestPath).openStream())
+                    Attributes attr = manifest.getMainAttributes()
+                    String version = attr.getValue("version")
+                    dependencies.add(project.dependencies.create("net.corda.plugins:javac-plugin:$version") {
+                        it.transitive = false
+                    })
+                }
+            }
+        }
```
chrisr3, line 117: If we expect to download the compiler plugin as a regular project dependency then why does it also need to be an `implementation` dependency of `quasar-utils`?

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
```diff
@@ -100,6 +130,14 @@ class QuasarPlugin implements Plugin<Project> {
...
             }
         }
+        project.tasks.withType(JavaCompile).configureEach {
+            doFirst {
+                if(extension.enableJavacPlugin.get()) {
+                    options.annotationProcessorPath += project.configurations[JAVAC_PLUGIN]
+                    options.compilerArgs += "-Xplugin:net.corda.plugins.javac.quasar.SuspendableChecker"
+                }
+            }
+        }
```
chrisr3, line 140: How does this work for Kotlin classes?

chrisr3 (commented):
Interesting, but this doesn't look compatible with Corda 5's plan for CorDapps to use a custom `@Suspendable` annotation instead of `co.paralleluniverse.fibers.Suspendable`. I am also concerned that it uses a "legacy" Java compiler plugin API, and could prevent us from using Gradle's new "toolchains".

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: unfortunately this only works for Java code (but suspendable methods defined in Kotlin should be detected correctly, it would be good to add a test for it though)

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: actually my idea was to simply bundle it with `quasar-utils` for simplicity

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: I'll change that

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/SuspendableDatabase.java
woggioni: that's correct, I will add it

quasar-utils/build.gradle
```diff
@@ -33,7 +33,14 @@ pluginBundle {
...
 dependencies {
+    implementation project(":quasar-utils:javac-plugin")
     testImplementation("org.junit.jupiter:junit-jupiter-api:$junit_jupiter_version")
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:$junit_jupiter_version")
     testImplementation "org.assertj:assertj-core:$assertj_version"
 }
+
+jar {
+    manifest {
+        attributes "version" : archiveVersion.get()
```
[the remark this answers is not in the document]
woggioni: because my idea was to ship it together with `quasar-utils`, as a separate artifact but as a direct dependency of it, so that people wouldn't have to declare the dependency explicitly

quasar-utils/build.gradle
```diff
@@ -33,7 +33,14 @@ pluginBundle {
 }
 
 dependencies {
+    implementation project(":quasar-utils:javac-plugin")
```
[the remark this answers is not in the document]
woggioni: do we currently publish the Corda Gradle plugins to the Gradle Plugins Portal? My understanding was that we were publishing them on r3's artifactory.. Also why does it need to be published to the same repo? doesn't Gradle loop through all declared repositories for every dependency until it finds one that contains it?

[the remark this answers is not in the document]
chrisr3: > do we currently publish the Corda Gradle plugins to the Gradle Plugins Portal? My understanding was that we were publishing them on r3's artifactory.. Also why does it need to be published to the same repo? doesn't Gradle loop through all declared repositories for every dependency until it finds one that contains it?

We have been publishing the plugins to [both places](https://plugins.gradle.org/plugin/net.corda.plugins.quasar-utils) for a while now. And Gradle uses a _different_ set of repositories to resolve its plugins than it does to resolve the project's dependencies. See [pluginManagement](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_management). Having your compiler plugin as a runtime dependency in a separate repository would therefore require _both_ repositories to be added to the `pluginManagement` block, regardless of whether the CorDapp developers wished to use your compiler plugin or not.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
chrisr3: I don't think that "only works for Java code" is good enough when both Corda and our CorDapp API are written in Kotlin. I suspect that it would be better not to use any compiler plugins at all, but to scan all the compiled `.class` files as a separate task that finalizes `classes`.

This task would need to take account of the configurable `@Suspendable` annotation too.

woggioni: I will add a way to configure the name of the `@Suspendable` annotation. Regarding kotlin, the vast majority of our customers use Java, so this plugin will be useful for them, there is apparently [a way](https://github.com/JetBrains/kotlin/blob/master/plugins/allopen/allopen-cli/src/AllOpenPlugin.kt) to do the same thing for `kotlinc`. I believe a compiler plugin gives an overall better user experience:

- class files only contain line number information, this may be less clear with code that contains very long lines
- class files don't know where the file exactly is on the filesystem, only contain
- a separate verification check won't be automatically invoked by the IDE when building the project and is likely to only be invoked on CI pipeline, as it currently happens with `detekt`

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: I actually had forgotten of this bit of code which looks really horrible.. I think I will simply include the compiler plugin in quasar-utils jar

quasar-utils/javac-plugin/src/test/resources/net/corda/plugins/javac/quasar/test/TestCase1.java
woggioni: there are comments inside each test case's file and I don't believe that a single word (like a Java type identifier) is good enough to describe any of those tests

quasar-utils/javac-plugin/build.gradle
woggioni: this looks to me like a less readable way to express the same thing...

quasar-utils/javac-plugin/build.gradle
woggioni: I'll investigate this

quasar-utils/javac-plugin/build.gradle
woggioni: will fix

quasar-utils/javac-plugin/src/java11/java/net/corda/plugins/javac/quasar/ASTUtils.java
woggioni: I agree, unfortunately the jdk.compiler module is not available for java 8. We might decide to only make this work on Corda 5+ but it would be better to have it for all Corda versions

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/SuspendExecution.java
woggioni: I put it here to avoid having the plugin depending on quasar, but I actually only need the name of this class, which means I can easily remove it

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
woggioni: I wasn't aware of that (and to be honest I don't even understand what's the advantage of not using `@Suspendable` since corda and cordapp code will still depends on quasar either way). I will address it

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/ASTUtils.java
woggioni: this is meant to address the small differences between the jdk 8 and jdk 11 APIs

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/SuspendableChecker.java
woggioni: I actually haven't tried but I expect it to work fine since either of the two will be used by both (if they are in the compiler classpath, both quasar and this plugin will be loaded by the same classloader, so it won't be possible to have duplicate classes). Either way, I agree it is not particularly elegant and I will remove the quasar class duplicates.

quasar-utils/javac-plugin/build.gradle
chrisr3: Of course it does the same thing. Anything else would be a bug :expressionless:.

quasar-utils/javac-plugin/src/java11/java/net/corda/plugins/javac/quasar/ASTUtils.java
chrisr3: You cannot play the "inclusiveness" card while simultaneously _excluding_ all of our CorDapp developers who use Kotlin. But I was actually thinking that this was another reason why using a compiler plugin was the wrong approach in the first place.

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
chrisr3: The point is that _the API we expose to CorDapps_ will no longer contain either Quasar or its army of transitive dependencies. Obviously Corda will continue to use Quasar internally.

More specifically, `quasar-core-0.8.x` has been modified so that you can add the class names of _additional_ `@Suspendable` annotations. Quasar's built-in `@Suspendable` annotation is still supported regardless. `qusar-utils-6.0.0-SNAPSHOT` now has a new
```groovy
Property<String> suspendableAnnotation
```
on its `quasar` extension to configure Quasar with Corda's own `@Suspendable` annotation.

Corda 5.x will use `quasar-core-0.8.x`. Earlier versions of Corda use `quasar-core-0.7.x`, which still only supports Quasar's built-in `@Suspendable` annotation.

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/ASTUtils.java
chrisr3: You mean the _currently existing_ small differences, as far as JDK11.

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
woggioni: imho, it would have been simpler to just extract `@Suspendable` and `SuspendExecution` to a separate `quasar-annotations` module, with `quasar-core` that depends on `quasar-annotations` but thanks for the clarification anyway

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/SuspendableChecker.java
chrisr3: > I actually haven't tried but ...

I haven't had much luck finding official documentation for Java compiler plugins, and so naturally I haven't found anything to confirm that a plugin's classes will _always_ be hosted in the application classloader either.

quasar-utils/javac-plugin/src/test/resources/net/corda/plugins/javac/quasar/test/TestCase1.java
chrisr3: > there are comments inside each test case's file and I don't believe that a single word (like a Java type identifier) is good enough to describe any of those tests

Then _don't_ use just a single word?

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
woggioni: I guess you will now have to agree with the node, when you install a new cordapp, about which annotation is meant to mark a suspendable method

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
chrisr3: Unfortunately, Kotlin is a "first class citizen" and so I don't believe that we can support a feature like this for Java only, but I can obviously raise this with @mnesbit and @fenryka in the New Year to see what they think.

As for your claim about "the vast majority of our customers", I can only point out that the vast majority of the issues that have been escalated to me from Support have used Kotlin rather than Java :expressionless:.

If your goal here is to provide developer feedback via the IDE when why not implement this as an IntelliJ plugin instead?

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
chrisr3: > I guess you will now have to agree with the node, when you install a new cordapp, about which annotation is meant to mark a suspendable method in a CorDapp

Like I said, Corda will provide an _extra_ `@Suspendable` annotation, and Corda 5 CorDapps will be expected to use it. Quasar will continue to support its built-in `@Suspendable` annotation too, and so there shouldn't be any scope for disagreement between CorDapps and the node.

To be honest, I initially thought that you had asked @lankydan to review this PR because you were already aware of the changes currently being made to Quasar.

lankydan: Everything @chrisr3 said is correct, we now have our own `@Suspendable` annotation, I _think_ its in `core` but either way its in our code base.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
chrisr3: I am aware of Kotlin compiler plugins too, but these seem to be strongly coupled to the particular version of Kotlin being used. And I have fought _tooth and nail_ over the years to prevent `quasar-utils` becoming responsible for configuring _all_ versions of _all_ compilers for _every_ JVM language that someone could conceivably write a CorDapp in.

I don't know if you've ever watched "The Twilight Zone", but there was an episode called ["The Curious Case of Edgar Witherspoon"](https://en.wikipedia.org/wiki/The_Curious_Case_of_Edgar_Witherspoon) that is _eerily close_ as an analogy for how the Corda Gradle plugins will function in Corda some day, unless we're careful.

quasar-utils/javac-plugin/src/main/java/co/paralleluniverse/fibers/Suspendable.java
chrisr3: > Everything @chrisr3 said is correct, we now have our own `@Suspendable` annotation, I _think_ its in `core` but either way its in our code base.

It's actually in `:base`, but I'll let you off :wink:.

quasar-utils/javac-plugin/src/main/java/net/corda/plugins/javac/quasar/SuspendableChecker.java
woggioni: The plugins need to be specified from the command line simply with `-Xplugin:PluginFullQualifiedClassName` and they are loaded by `javac` assuming them to be in the compilation classpath. 

This is actually quite annoying because it means that a compiler plugin cannot have any third party library dependency (or at least it should shade all of its own dependencies) since otherwise, if your compiler plugin uses Guava 21, that is shared with the code it is supposed to compile, meaning that either your code is tied to the Guava version of your compiler plugin or the other way around.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: removed it completely

quasar-utils/build.gradle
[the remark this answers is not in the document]
woggioni: I embedded the javac plugin inside `quasar-utils` Gradle plugin jar

quasar-utils/javac-plugin/src/test/resources/net/corda/plugins/javac/quasar/test/TestCase1.java
woggioni: you mean something like `CheckPluginCanDetectSuspendableMethodsMarkedUsingCusomAnnotations` ?

woggioni:
I created a new `tar` artifact that contains the javac plugin, the idea is to publish the jar as a standalone artifact somewhere for people that wish to use it with a build system other than Gradle (it will be up to them to integrate it in their build) and to bundle the tar inside quasar-util's jar.

The reason for a tar archive instead of a jar is that a tar is not compressed and will only be compressed once inside the jar as a zip file entry, while a jar is already compressed and will be compressed a second time if included in another jar by Gradle (there is no way with Gradle API to use [STORED](https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html#STORED)).

My initial idea was to simply embed the class files in the Gradle plugin and selectively extract those classes at (Gradle) runtime, but I discovered that when Gradle runs it first copies all of its jars in a temporary folder, instrumenting all the class files and annotating all instrumented methods with `org.gradle.internal.classpath.Instrumented`, including the compiler plugin classes since they were in that way indistinguishable from the quasar-utils plugin classes. The issue is that the `org.gradle.internal.classpath.Instrumented` annotation is not available in the compilation classpath that `javac` uses, causing the compilation to fail because the javac plugin is using classes that are impossible to load.

quasar-utils/javac-plugin/build.gradle
woggioni: [as I understand](https://blog.gradle.org/java-toolchains), toolchains are a feature only available since gradle 6.7, this sets of plugins currently builds on `5.6.4` (that set of dependencies is only required at plugin build time), this means I cannot do anything but ignore Gradle toolchains for now

chrisr3: The Corda Gradle plugins for Corda 5.0 are currently targeting Gradle 6.6+, but there would be no problem raising this to Gradle 6.7+ if there were a good reason to use Toolchains. Unfortunately, you have created this unscheduled PR against a branch of the plugins that is still intended for use with Corda 4.x. So not only are you still using Gradle 5.6.4 but you are also missing all of the changes to `quasar-utils` to support the configurable `@Suspendable` annotation. (For reference, we did not make `co.paralleluniverse.fibers.SuspendExecution` throwables configurable in Quasar 0.8.x.)

I strongly recommend not continuing to work on this PR. As I understand it [CORE-187](https://r3-cev.atlassian.net/browse/CORE-187) was scheduled for "Milestone-2" anyway.

woggioni: I personally consider a mistake making `@Suspendable` configurable in the first place, but since we now have that, I thought it would be consistent to allow for a configurable `SuspendExecution` as well...

chrisr3: Advertising functionality _which Quasar does not support_ is a waste of time and effort.

Please just stop this. You seem only to be implementing what you believe the requirements "should" be, and not what the requirements actually _are_. This is not productive.

chrisr3:
Closed as abandoned.

PR #366, opened 2021-02-12
CORE-853: Do not log passwords when signing CorDapp jars.

### LINKED ISSUE #303: Option to turn off plain text passwords in stdout during signing
It appears that the current release outputs the signing `storepass` and `keypass` passwords as plain text to files. This is a problem for open source cordapps that expose the build output to contributors. Can we please have an option to not print out passwords? Ideally this should be environment driven so that it cannot be turned off by contributors. 

Thanks.

chrisr3:
Replace the key and store password with `****` when logging the signing options.

This fixes #303.


woggioni (approved):
LGTM

PR #367, opened 2021-02-15
CORE-855: Initial update for 6.0.0-SNAPSHOT.

chrisr3:
Move the development of the 6.0.0-SNAPSHOT plugins to the master branch.

This large and probably unsightly code-drop is just shifting the development of the `6.0.0-SNAPSHOT` plugins to the `master` branch. It is not yet considered to be "complete". However, it is still probably worth looking at by one or more someones other than me  :grinning:.


chrisr3:
Sorry, this PR is a bad idea. I'll break it up properly so that the `master` branch remains viable :facepalm:.

PR #372, opened 2021-02-16
CORE-855: Upgrade to Gradle 6.7.1.

chrisr3:
Upgrade to Gradle 6.7.1, which is the same version of Gradle that Corda 5 is currently using.

Notable features:
- Uses Kotlin 1.3.72 internally.
- Supports multiple JVM "toolchains", so that we can compile with JDK8 but test using JDK11 (for selected plugins).
- Stricter validation checks when writing plugins.


build.gradle
```diff
@@ -205,11 +202,11 @@ configure(publishProjects) { Project subproject ->
...
 artifactory {
     publish {
-        contextUrl = 'https://software.r3.com/artifactory'
+        contextUrl = artifactory_contextUrl
         repository {
             repoKey = 'corda-dev'
-            username = System.getenv('CORDA_ARTIFACTORY_USERNAME') ?: System.getProperty('corda.artifactory.username')
-            password = System.getenv('CORDA_ARTIFACTORY_PASSWORD') ?: System.getProperty('corda.artifactory.password')
+            username = project.findProperty('cordaArtifactoryUsername') ?: System.getenv('CORDA_ARTIFACTORY_USERNAME') ?: System.getProperty('corda.artifactory.username')
+            password = project.findProperty('cordaArtifactoryPassword') ?: System.getenv('CORDA_ARTIFACTORY_PASSWORD') ?: System.getProperty('corda.artifactory.password')
```
palade, line 209: I'm wondering maybe we can avoid having this check. My understanding of the `findProperty` procedure is that it will look up for some property set either in this file or in the `settings` file. This probably that the username and password may be hardcoded in either of those two files, which could be rarely the case.

palade (approved):
The proposed changes look good to me. I left one small comment.

build.gradle
chrisr3: This check allows people (i.e. "me" :wink:) to store my Artifactory API key in my `$HOME/.gradle/gradle.properties` file, and hence publish `SNAPSHOT` builds without needing either to set environment variables or Java system properties.

Putting my credentials into `settings.gradle` or `build.gradle` would be Bad Idea :tm:.

woggioni (approved):
LGTM except that I have no clue about how this shiny new thing of toolchains works but I trust @chrisr3 with using it sensibly

chrisr3:
> LGTM except that I have no clue about how this shiny new thing of toolchains works ... I guess I'll find my answers in Gradle documentation)

Gradle toolchains are documented [here](https://docs.gradle.org/current/userguide/toolchains.html), and you can indeed also select a JDK by vendor, as well as providing "auto-detection" and "custom provisioning". We'll probably revise how we use it as time goes on, but it has successfully cleared the first "build on Jenkins" hurdle,

PR #374, opened 2021-02-16
CORE-855: Update quasar-utils plugin to use OSGi Quasar 0.8.x.

chrisr3:
Update `quasar-utils` to use the OSGi-compatible Quasar 0.8.x artifacts. The tests only verify the Quasar `-javaagent:` parameter is correct, and do not actually instrument any classes. This means that they do not need to be executed on a Java 11+ JVM.

### DIFF
--- a/quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarExtension.groovy
+++ b/quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarExtension.groovy
@@ -21,6 +21,11 @@ class QuasarExtension {
     final Property<String> version
 
     /**
+     * Class name for Quasar's @Suspendable annotation.
+     */
+    final Property<String> suspendableAnnotation
+
+    /**
      * Dependency notation for the Quasar bundle to use.
      */
     final Provider<Map<String, String>> dependency
@@ -53,6 +58,7 @@ class QuasarExtension {
         ObjectFactory objects,
         String defaultGroup,
         String defaultVersion,
+        String defaultSuspendable,
         Iterable<? extends String> initialPackageExclusions,
         Iterable<? extends String> initialClassLoaderExclusions
     ) {
@@ -68,6 +74,7 @@ class QuasarExtension {
 
         debug = objects.property(Boolean).convention(false)
         verbose = objects.property(Boolean).convention(false)
+        suspendableAnnotation = objects.property(String).convention(defaultSuspendable)
         excludePackages = objects.listProperty(String)
         excludePackages.set(initialPackageExclusions)
         excludeClassLoaders = objects.listProperty(String)
@@ -75,21 +82,27 @@ class QuasarExtension {
         options = excludePackages.flatMap { packages ->
             excludeClassLoaders.flatMap { classLoaders ->
                 debug.flatMap { isDebug ->
-                    verbose.map { isVerbose ->
-                        def builder = new StringBuilder('=')
-                        if (isDebug) {
-                            builder.append('d')
+                    verbose.flatMap { isVerbose ->
+                        suspendableAnnotation.orElse('').map { ann ->
+                            def builder = new StringBuilder('=')
+                            if (isDebug) {
+                                builder.append('d')
+                            }
+                            if (isVerbose) {
+                                builder.append('v')
+                            }
+                            if (!packages.isEmpty()) {
+                                builder.append('x(').append(packages.join(';')).append(')')
+                            }
+                            if (!classLoaders.isEmpty()) {
+                                builder.append('l(').append(classLoaders.join(';')).append(')')
+                            }
+                            final String annotation = ann.trim()
+                            if (!annotation.isEmpty()) {
+                                builder.append('a(SUSPENDABLE=').append(annotation).append(')')
+                            }
+                            builder.length() == 1 ? '' : builder.toString()
                         }
-                        if (isVerbose) {
-                            builder.append('v')
-                        }
-                        if (!packages.isEmpty()) {
-                            builder.append('x(').append(packages.join(';')).append(')')
-                        }
-                        if (!classLoaders.isEmpty()) {
-                            builder.append('l(').append(classLoaders.join(';')).append(')')
-                        }
-                        builder.length() == 1 ? '' : builder.toString()
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
```diff
@@ -42,24 +47,20 @@ class QuasarPlugin implements Plugin<Project> {
...
-        project.pluginManager.apply(JavaPlugin)
-
         def rootProject = project.rootProject
-        def quasarGroup = rootProject.hasProperty('quasar_group') ? rootProject.property('quasar_group') : defaultGroup
-        def quasarVersion = rootProject.hasProperty('quasar_version') ? rootProject.property('quasar_version') : defaultVersion
-        def quasarClassifier = rootProject.hasProperty('quasar_classifier') ? rootProject.property('quasar_classifier') : defaultClassifier
-        def quasarPackageExclusions = rootProject.hasProperty("quasar_exclusions") ? rootProject.property('quasar_exclusions') : Collections.emptyList()
+        def quasarGroup = rootProject.findProperty('quasar_group')?.toString() ?: defaultGroup
+        def quasarVersion = rootProject.findProperty('quasar_version')?.toString() ?: defaultVersion
+        def quasarSuspendable = rootProject.findProperty('quasar_suspendable_annotation')?.toString()?.trim()
```
woggioni, line 53: shouldn't we have a default value for this?

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarExtension.groovy
```diff
@@ -53,45 +58,51 @@ class QuasarExtension {
...
-            version.flatMap { ver ->
-                classifier.map { cls ->
-                    [ group: grp, name: 'quasar-core', version: ver, classifier: cls, ext: 'jar' ]
-                }
-            }
+        dependency = group.zip(version) { grp, ver ->
+            [ group: grp, name: 'quasar-core-osgi', version: ver, ext: 'jar' ]
+        }
+        agent = dependency.map {
+            it['classifier'] = 'agent'
```
woggioni, line 71: doesn't this modify the above map
```groovy
[ group: grp, name: 'quasar-core-osgi', version: ver, ext: 'jar' ]
```
in place?

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
```diff
@@ -68,49 +69,92 @@ class QuasarPlugin implements Plugin<Project> {
...
+        }
+
+        @Override
+        void execute(DependencySet dependencies) {
+            if (isEmpty(extension.suspendableAnnotation)) {
+                def quasarDependency = handler.create(extension.dependency.get()) { dep ->
+                    dep.transitive = isTransitive
+                }
+                dependencies.add(quasarDependency)
+            }
```
woggioni, line 153: shouldn't we add the quasar dependency when `extension.suspendableAnnotation` is **NOT** empty?

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
chrisr3: No, because the default value is Quasar's own `co.paralleluniverse.fibers.Suspendable` annotation, which Quasar uses anyway.

woggioni: 👍🏿

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
chrisr3: The idea is that the CorDapp only needs to depend on Quasar for the sake of its `@Suspendable` annotation. So we don't need `quasar-core-osgi` on our compile classpath if we're using something else.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarExtension.groovy
chrisr3: `dependency` is a "lazy" Gradle `Provider`, and so is re-evaluated whenever you execute it. (Unless it has been finalized, which it isn't in this case.) Basically it's just a sequence of operations to generate a value.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: ok but that means that by default users will stick to quasar's `@Suspendable` unless they configure manually this plugin to use the `@Suspendable` from `corda-base`, right? My concern is simply they would never switch because:
- default is quasar's `@Suspendable`
- simply switching to `@Suspendable` from `corda-base` would cause their tests to fail unless they remember to also configure the gradle plugin

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarExtension.groovy
woggioni: oh, ok... I thought the returned value was cached after the first execution and the closure never run again

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
chrisr3: CorDapp developers start from a template Gradle project, which will need updating for Corda 5 anyway. We have never asked them to write their own Gradle projects from scratch. Few people in R3 can even do that.... :wink:.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarExtension.groovy
chrisr3: You _can_ make providers behave like that, if you want - see [finalizeValue](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/HasConfigurableValue.html#finalizeValue--) and [finalizeValueOnRead](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/HasConfigurableValue.html#finalizeValueOnRead--). _However_, Gradle may want to check these values several times and you wouldn't want to finalize them too early by accident.

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
woggioni: ok

woggioni (approved):

palade (approved):

PR #378, opened 2021-02-17
CORE-855: Update Cordformation plugin for Gradle 6.7.1.

chrisr3:
- Replace uses of deprecated Gradle configurations.
- Update definition of `cordapp` configuration not to use deprecated `compile`.
- Use new `cordaRuntimeOnly` configuration instead of `cordaRuntime`.
- _Finally_ remove the deprecated `rpcPort` element from the DSL.
- Resolve compiler warnings.

### DIFF
--- a/cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordapp.gradle
+++ b/cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordapp.gradle
@@ -5,11 +5,11 @@ plugins {
 apply from: 'repositories.gradle'
 
 dependencies {
-    cordaRuntime "$corda_group:corda:$corda_release_version"
-    cordaRuntime "$corda_group:corda-node-api:$corda_release_version"
+    cordaRuntimeOnly "$corda_group:corda:$corda_release_version"
+    cordaRuntimeOnly "$corda_group:corda-node-api:$corda_release_version"
     cordapp "$corda_group:corda-finance-contracts:$corda_release_version"
     cordapp "$corda_group:corda-finance-workflows:$corda_release_version"
-    cordaRuntime "org.slf4j:slf4j-simple:$slf4j_version"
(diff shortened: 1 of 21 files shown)


cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordapp.gradle
```diff
@@ -20,8 +20,8 @@ task deployNodes(type: net.corda.plugins.Cordform) {
         name 'O=Notary Service,L=Zurich,C=CH'
         notary = [validating : true]
         p2pPort 10002
-        rpcPort 10003
         rpcSettings {
+            port 10003
```
palade, line 24: Has this option to specify the RPC port changed to from `rpcPort` to `rpcSettings.port` or are both valid? Wondering if we may need to update the documentation

palade (approved):
The proposed changes look good to me. I left one small comment in case we may need to update the docs.

cordformation/src/test/resources/net/corda/plugins/DeploySingleNodeWithCordapp.gradle
chrisr3: The plugin has been screaming at people for ages that the old `rpcPort` is deprecated and scheduled to be removed in Corda 5:
```kotlin
@Deprecated("Use {@link CordformNode#rpcSettings(RpcSettings)} instead. Will be removed by Corda V5.0.")
```
People are _supposed_ to have migrated by now :hourglass_flowing_sand:. Although to be fair, I'm not even sure where this documentation you mention actually _is_... :eyes:.

palade: Actually the documentation is fine seems up to date: https://docs.corda.net/docs/corda-os/4.7/generating-a-node.html

woggioni (approved):

PR #379, opened 2021-02-17
CORE-855: Adjust definition of cordaProvided configuration.

chrisr3:
A `cordaProvided` dependency contributes to `compileClasspath` but not to `runtimeClasspath`, because we expect Corda to provide it at runtime. However, we would also like to include it on the `testRuntimeClasspath`, `integrationTestRuntimeClasspath` etc to make developers' testing easier.

We will use the contents of the `runtimeClasspath` to determine the dependencies to include in the CPK, and so need to keep it free of Corda artifacts and their transitive dependencies.

Also update `publish-utils` for Gradle 6.6 lazy artifact support, c.f. "task configuration avoidance".

### DIFF
--- a/quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
+++ b/quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
@@ -86,7 +86,7 @@ class QuasarPlugin implements Plugin<Project> {
         // If we're building a JAR then also add the Quasar bundle to the appropriate configurations.
         project.pluginManager.withPlugin('java') {
             // Add Quasar bundle to the compile classpath WITHOUT any of its transitive dependencies.
-            def cordaProvided = createCompileOnlyConfiguration(CORDA_PROVIDED_CONFIGURATION_NAME, project.configurations)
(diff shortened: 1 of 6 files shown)


palade (approved):

woggioni (approved):

quasar-utils/src/main/groovy/net/corda/plugins/quasar/QuasarPlugin.groovy
```diff
@@ -120,12 +120,16 @@ class QuasarPlugin implements Plugin<Project> {
         return configuration
     }
 
-    private static Configuration createCompileOnlyConfiguration(String name, ConfigurationContainer configurations) {
+    private static Configuration createCompileConfiguration(String name, ConfigurationContainer configurations) {
+        return createCompileConfiguration(name, "Implementation", configurations)
```
woggioni, line 124: you might want to use `org.gradle.api.plugins.JavaPlugin#IMPLEMENTATION_CONFIGURATION_NAME`

chrisr3: I suspect that value is "implementation" rather than "Implementation". I am leveraging Gradle's naming convention to modify all `xxxxImplementation` configurations here, where `xxxx` could be `test`, `integrationTest`, `smokeTest`, `bongoTest` etc.

PR #386, opened 2021-02-26
CORE-727: Initialize new ZipEntry objects more thoroughly.

chrisr3:
A directory entry contains zero bytes and should be `STORED`. It is also unsafe to add a `ZipEntry` from a `ZipInputStream` directly into a `ZipOutputStream` as a `ZipEntry` field has several internal fields that its stream will update.

Also create a temporary subdirectory for the heartbeat agent when adding it to the flask.

### DIFF
--- a/flask/src/main/java/net/corda/gradle/flask/FlaskJarTask.java
+++ b/flask/src/main/java/net/corda/gradle/flask/FlaskJarTask.java
@@ -39,6 +39,8 @@ import java.util.jar.Manifest;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
+import static java.util.zip.Deflater.BEST_COMPRESSION;
+import static java.util.zip.Deflater.NO_COMPRESSION;
 
 public class FlaskJarTask extends AbstractArchiveTask {
 
@@ -147,6 +149,10 @@ public class FlaskJarTask extends AbstractArchiveTask {
                 zipEntry.setTime(fileCopyDetails.getLastModified());
             }
             if (fileCopyDetails.isDirectory()) {
+                zipEntry.setMethod(ZipEntry.STORED);
+                zipEntry.setCompressedSize(0);
+                zipEntry.setSize(0);
+                zipEntry.setCrc(0);
                 zoos.putNextEntry(zipEntry);
             } else {
                 boolean compressed = Flask.splitExtension(fileCopyDetails.getSourceName())
@@ -160,8 +166,8 @@ public class FlaskJarTask extends AbstractArchiveTask {
                     }
                     zipEntry.setMethod(ZipEntry.STORED);
                 }
+                zoos.putNextEntry(zipEntry);
                 try (InputStream is = Flask.read(fileCopyDetails.getFile(), false)) {
-                    zoos.putNextEntry(zipEntry);
                     Flask.write2Stream(is, zoos, buffer);
                 }
             }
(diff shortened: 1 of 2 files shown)


flask/src/main/java/net/corda/gradle/flask/FlaskJarTask.java
```diff
@@ -218,12 +225,14 @@ public WorkResult execute(@Nonnull CopyActionProcessingStream copyActionProcessi
                  */
                 File temporaryJar = File.createTempFile("premature", ".zip", getTemporaryDir());
                 try (ZipOutputStream zipOutputStream = new ZipOutputStream(Flask.write(temporaryJar, true))) {
+                    zipOutputStream.setLevel(BEST_COMPRESSION);
```
woggioni, line 228: we surely don't want to use `BEST_COMPRESSION` (that takes a lot of cpu resources) for a temporary jar we are going to immediately discard

flask/src/main/java/net/corda/gradle/flask/FlaskJarTask.java
```diff
@@ -102,7 +104,8 @@ public FlaskJarTask(@Nonnull ObjectFactory objects) {
         from(getProject().tarTree(LauncherResource.instance), copySpec -> exclude(JarFile.MANIFEST_NAME));
 
         Provider<File> heartbeatJarProvider = getProject().provider(() -> {
-            File dest = new File(getTemporaryDir(), HeartbeatAgentResource.instance.getDisplayName());
+            File tempSubDir = Files.createTempDirectory(getTemporaryDir().toPath(), "heartbeat").toFile();
```
woggioni, line 107: this will create a temporary archive for every task execution.. It also came to my mind that I had thought about task concurrency when writing this and actually using an hardcoded filename was not an issue as the temporary directory is
`build/tmp/taskName` and you can't have two `FlaskJarTask` instances with the same name in the same project

woggioni: I also assume you won't run the same `flaskJar` task from 2 different Gradle processes concurrently in the same workspace, since that would also cause trouble when creating the final artifact

woggioni (commented):
It also came to my mind that I had thought about task concurrency when writing this and actually using an hardcoded filename was not an issue as the temporary directory is
`build/tmp/taskName` and you can't have two `FlaskJarTask` instances with the same name in the same project

flask/src/main/java/net/corda/gradle/flask/FlaskJarTask.java
chrisr3: True; I can switch that to `NO_COMPRESSION`.

flask/src/main/java/net/corda/gradle/flask/FlaskJarTask.java
chrisr3: If we don't need it, then we don't need it.

woggioni:
could you also revert the [other commit from this morning](https://github.com/corda/corda-gradle-plugins/commit/87aa7f75fba1b53bc451f9bd7cc5a4758ae0d0e4) for the same reason outlined above (it will just bloat the build directory with a new archive for each task execution with no actual benefit)?

chrisr3:
> could you also revert the [other commit from this morning](https://github.com/corda/corda-gradle-plugins/commit/87aa7f75fba1b53bc451f9bd7cc5a4758ae0d0e4) for the same reason outlined above (it will just bloat the build directory with a new archive for each task execution with no actual benefit)?

I don't understand what you mean by "bloat the build directory with a new archive", because it only generates a unique filename for a file that it must create anyway. The point here is that the parent directory is already unique, and so it doesn't matter.

woggioni (approved):
LGTM

woggioni:
> > could you also revert the [other commit from this morning](https://github.com/corda/corda-gradle-plugins/commit/87aa7f75fba1b53bc451f9bd7cc5a4758ae0d0e4) for the same reason outlined above (it will just bloat the build directory with a new archive for each task execution with no actual benefit)?
> 
> I don't understand what you mean by "bloat the build directory with a new archive", because it only generates a unique filename for a file that it must create anyway. The point here is that the parent directory is already unique, and so it doesn't matter.
> 
> But the same logic applies: If we don't need it, we don't need it.

I meant that it will create a new (different) premature jar every time you execute the task and those jars will accumulate in the build directory until you call clean

chrisr3:
> I meant that it will create a new (different) premature jar every time you execute the task and those jars will accumulate in the build directory until you execute `clean`

Gradle isn't supposed to re-execute the task unless the task's inputs change. Gradle also [gives no guarantee that the temporary directory will outlive the task's execution](https://docs.gradle.org/current/dsl/org.gradle.api.Task.html#org.gradle.api.Task:temporaryDir) either. But anyway, it's gone...

PR #387, opened 2021-02-26
CORE-727: Create directory ZipEntry objects using the factory.

chrisr3:
We create zip directory entries in two places, so move their creation into `ZipEntryFactory` and then add a test to ensure we have caught them all.

Move handling of the "last modified" timestamp into `ZipEntryFactory` too.

Note: When using `ZipInputStream`, each `ZipEntry`'s `time`, `comment` and `extra` values are available immediately. But a `DEFLATED` entry's entire `InputStream` needs to be consumed before its `size`, `compressedSize` and `crc` values are set.

(:rotating_light: And this commit can be _**SQUASHED**_ too! :rotating_light:)

### DIFF
--- a/flask/build.gradle
+++ b/flask/build.gradle
@@ -51,6 +51,7 @@ dependencies {
         add(conf, [group: "org.projectlombok", name: "lombok", version: lombok_version])
     }
     testImplementation project("flask-common")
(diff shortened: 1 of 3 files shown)


woggioni (approved):
LGTM

PR #394, opened 2021-03-11
NOTICK Fixed race condition on lock file creation

woggioni:
According to [RandomAccessFile documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/RandomAccessFile.html#%3Cinit%3E(java.io.File,java.lang.String)) there is no need to create the file before invoking the constructor since with 'rw' mode it will be created if it doesn't exist yet.
This also resolves a race condition where multiple Flask processes are started, the lock file file doesn't exist so one process try to create it but between the `File.exists` call and the `Files.createFile` call another process creates it causing `Files.createFile` to fail with `java.nio.file.FileAlreadyExistsException`

### DIFF
--- a/flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
+++ b/flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
@@ -17,7 +17,6 @@ public class LockFile implements Closeable {
 
     private static RandomAccessFile createLockFile(Path path) throws IOException {
         Files.createDirectories(path.getParent());
-        if (!Files.exists(path)) Files.createFile(path);
         return new RandomAccessFile(path.toFile(), "rw");
     }
 


flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -17,7 +17,6 @@
 
     private static RandomAccessFile createLockFile(Path path) throws IOException {
         Files.createDirectories(path.getParent());
-        if (!Files.exists(path)) Files.createFile(path);
         return new RandomAccessFile(path.toFile(), "rw");
```
blsemo, line 20: How did you verify that this is actually safe?

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -17,7 +17,6 @@
 
     private static RandomAccessFile createLockFile(Path path) throws IOException {
         Files.createDirectories(path.getParent());
```
chrisr3, line 19: I think [`Files.createDirectories`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectories-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-) will also throw `FileAlreadyExistsException` if two threads race to create the same directories.

woggioni: in theory, [it shouldn't](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectories-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-)
```
Unlike the createDirectory method, an exception is not thrown if the directory could not be created because it already exists
```

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
woggioni: what do you mean with "safe"? safe that the following line will create the file when it doesn't exist or safe that a race condition won't crash the code?

woggioni: at the moment we have unit test creating a flask jar and running it in a fresh environment, but we do not have a test for race conditions (as they are quite tricky to reproduce even in testing environment)

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
chrisr3: Yes, I saw that. Hence why I updated my comment.

chrisr3:
This branch needs rebasing before it can be merged.

woggioni:
I actually realized this is not needed anymore since [another PR](https://github.com/corda/corda-gradle-plugins/pull/395) will address this isssue

PR #395, opened 2021-03-11
CORE-924 fixed RandomAccessFile leak

woggioni:
`RandomAccessFile` is a `Closeable` and needs to be closed, otherwise it will leak an open file descriptor that could prevent the file from being deleted on some OSes (namely Windows)

### DIFF
--- a/flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
+++ b/flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
@@ -5,40 +5,39 @@ import lombok.SneakyThrows;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.io.RandomAccessFile;
+import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.EnumSet;
 
 @RequiredArgsConstructor
 public class LockFile implements Closeable {
 
     private final FileLock lock;
 
-    private static RandomAccessFile createLockFile(Path path) throws IOException {
+    private static FileChannel openFileChannel(Path path) throws IOException {
         Files.createDirectories(path.getParent());
-        if (!Files.exists(path)) Files.createFile(path);
-        return new RandomAccessFile(path.toFile(), "rw");
+        return FileChannel.open(path, EnumSet.of(StandardOpenOption.CREATE, StandardOpenOption.READ, StandardOpenOption.WRITE));
     }
 
     @SneakyThrows
     public static LockFile acquire(Path path, boolean shared) {
-        RandomAccessFile raf = createLockFile(path);
-        return new LockFile(raf.getChannel().lock(0L, Long.MAX_VALUE, shared));
+        FileChannel channel = openFileChannel(path);
+        return new LockFile(channel.lock(0L, Long.MAX_VALUE, shared));
     }
 
     @SneakyThrows
     public static LockFile tryAcquire(Path path, boolean shared) {
-        RandomAccessFile raf = createLockFile(path);
-        FileLock lock = raf.getChannel().tryLock(0L, Long.MAX_VALUE, shared);
-        if(lock != null)
-            return new LockFile(lock);
-        else return null;
+        FileChannel channel = openFileChannel(path);
+        FileLock lock = channel.tryLock(0L, Long.MAX_VALUE, shared);
+        return (lock != null) ? new LockFile(lock) : null;
     }
 
     @Override
     @SneakyThrows
     public void close() {
-        lock.release();
+        lock.channel().close();
     }
 }
--- a/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
+++ b/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
@@ -95,7 +95,7 @@ public class Launcher {
             }
             ZipEntry javaAgentsEntry = jar.getEntry(Flask.Constants.JAVA_AGENTS_FILE);
             if(javaAgentsEntry != null) {
-                try (InputStream inputStream = jar.getInputStream(jar.getEntry(Flask.Constants.JAVA_AGENTS_FILE))) {
+                try (InputStream inputStream = jar.getInputStream(javaAgentsEntry)) {
                     javaAgents = listOfStringFromPropertyFile(inputStream);
                 }
             }
@@ -160,8 +160,9 @@ public class Launcher {
                 int returnCode = builder.exec();
                 afterChildJvmExit(returnCode);
                 cache.touchLibraries();
-                Files.delete(cache.getPidFile());
                 return returnCode;
+            } finally {
+                Files.delete(cache.getPidFile());
             }
         }
     }


flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -24,21 +25,22 @@ private static RandomAccessFile createLockFile(Path path) throws IOException {
     @SneakyThrows
     public static LockFile acquire(Path path, boolean shared) {
         RandomAccessFile raf = createLockFile(path);
-        return new LockFile(raf.getChannel().lock(0L, Long.MAX_VALUE, shared));
+        return new LockFile(raf, raf.getChannel().lock(0L, Long.MAX_VALUE, shared));
```
chrisr3, line 28: While `RandomAccessFile` is indeed `Closeable`, it is implemented around an instance of `FileDescriptor` which is passed to the `FileChannel` and likely also used by the `FileLock` class. (On UNIX, anyway.)

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -24,21 +25,22 @@ private static RandomAccessFile createLockFile(Path path) throws IOException {
...
-            return new LockFile(lock);
+            return new LockFile(raf, lock);
         else return null;
     }
 
     @Override
     @SneakyThrows
     public void close() {
         lock.release();
+        raf.close();
```
chrisr3, line 44: Maybe just this is enough?
```java
lock.close()
```
Closing the file descriptor should automatically release all file locks against it. However, file-locking in Java is a wrapper around the platform's native primitives and so you will need to test how this behaves on Windows too.

chrisr3 (commented):
I suspect this may be overkill, but will also need explicit testing on Windows.

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
chrisr3: Actually, this looks more likely:
```java
lock.channel().close();
```
Or to be _really_ pedantic:
```java
try {
    lock.close();
} finally {
    lock.channel().close();
}
```

LWogan:
I've tested Walters change and it works on Windows

LWogan (approved):

chrisr3:
> I've tested Walters change and it works on Windows

Could you test whether reimplementing `LockFile.close()` as
```java
@Override
public void close() {
    lock.channel().close();
}
```
also works on Windows please? (The `FileChannel` here is documented as being associated with the `RandomAccessFile`.)

LWogan:
`@Override
public void close() {
    lock.channel().close();
}`

Tested this too and it also works on windows.

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -13,6 +13,7 @@
 @RequiredArgsConstructor
 public class LockFile implements Closeable {
 
+    private final RandomAccessFile raf;
```
chrisr3, line 16: We don't need this field. Closing its associated `FileChannel` is enough.

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -24,21 +25,22 @@ private static RandomAccessFile createLockFile(Path path) throws IOException {
...
     }
 
     @SneakyThrows
     public static LockFile tryAcquire(Path path, boolean shared) {
         RandomAccessFile raf = createLockFile(path);
         FileLock lock = raf.getChannel().tryLock(0L, Long.MAX_VALUE, shared);
         if(lock != null)
-            return new LockFile(lock);
+            return new LockFile(raf, lock);
         else return null;
```
chrisr3, line 37: I'm sure this can be rewritten as:
```java
return (lock != null) ? new LockFile(lock) : null;
```
This is neater than having an `if-else` block without any `{}`.

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
chrisr3, line 44: @LWogan has confirmed that
```java
@Override
public void close() {
    lock.channel().close();
}
```
is enough here.

chrisr3 (changes requested):
I think we can simplify this.

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
woggioni: so, this was created because of [this](https://r3-cev.atlassian.net/browse/CORE-924)
Thinking about it, I realized we don't really need to use `RandomAccessFile` since we never write to the lockfile, hence I replaced ti with a simple `FileChannel` (and also addressed https://github.com/corda/corda-gradle-plugins/pull/394)

woggioni: yeah, I tested myself on a Windows VM and that works

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
```diff
@@ -4,41 +4,36 @@
...
-        if (!Files.exists(path)) Files.createFile(path);
-        return new RandomAccessFile(path.toFile(), "rw");
-    }
-
     @SneakyThrows
     public static LockFile acquire(Path path, boolean shared) {
-        RandomAccessFile raf = createLockFile(path);
-        return new LockFile(raf.getChannel().lock(0L, Long.MAX_VALUE, shared));
+        Files.createDirectories(path.getParent());
+        FileChannel channel = FileChannel.open(path, EnumSet.of(StandardOpenOption.CREATE, StandardOpenOption.READ, StandardOpenOption.WRITE));
```
chrisr3, line 22: These lines are common to both `acquire()` and `tryAcquire()` and so could be refactored into
```java
private static FileChannel createChannel(Path path)
```
I see that there are also `DELETE_ON_CLOSE` and `TRUNCATE` options :grinning:.

chrisr3 (changes requested):
Just one thing...

flask/flask-common/src/main/java/net/corda/flask/common/LockFile.java
woggioni: I've noticed them as well... they could be useful but we do not want to always delete the lockfile on close, we currently use 2 lock files:
- one to lock the jar cache to prevent a process from removing from the cache jar files that another process is using (cache cleanup is performed at process startup)
- one as a parent-process heartbeat so that the child can notice that its creator has exited

For the first case we surely do not want the lockfile to be deleted when a process terminates as the same lockfile is shared between all the Flask processes (I assume the deletion would fail on Windows); for the second case we want to delete the lock file when the last between the parent and the child process exits and setting those flags means the child (that normally exits before the parent) tries to delete the lockfile while the parent process still holds it

chrisr3: Fair enough, I thought you were deleting the lock files manually afterwards.

woggioni: I'm going to verify this anyway

chrisr3: > we are in [the first case](https://github.com/corda/corda-gradle-plugins/blob/woggioni/flask-bugfix/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java#L165) and I'm wondering whether this is correct (I'm thinking about the case were you forcefully terminate the parent process sending SIGINT, the child is only killed by a shutdown hook at the end so it is still alive waiting to acquire the lock when then finally block runs, but that should be ok because I think that if you send SIGINT the finally block doesn't run at all)

I've just looked at that file more closely, and am now wondering why these lines:
```java
ZipEntry javaAgentsEntry = jar.getEntry(Flask.Constants.JAVA_AGENTS_FILE);
if(javaAgentsEntry != null) {
    try (InputStream inputStream = jar.getInputStream(jar.getEntry(Flask.Constants.JAVA_AGENTS_FILE))) {
```
aren't written as:
```java
ZipEntry javaAgentsEntry = jar.getEntry(Flask.Constants.JAVA_AGENTS_FILE);
if(javaAgentsEntry != null) {
    try (InputStream inputStream = jar.getInputStream(javaAgentsEntry)) {
```

woggioni: yeah, that's just silly... I'll address it

woggioni: > we are in the first case and I'm wondering whether this is correct (I'm thinking about the case were you forcefully 
> terminate the parent process sending SIGINT, the child is only killed by a shutdown hook at the end so it is still alive 
> waiting to acquire the lock when then finally block runs, but that should be ok because I think that if you send SIGINT the 
> finally block doesn't run at all)

I confirm the `finally` block isn't executed when the parent process is forcefully terminated

chrisr3: Sending `SIGINT` to a JVM sounds like UNIX-specific behaviour to me. I'm not aware that Windows has an equivalent. My dim memories of ancient Windows programming suggest that [Control-C handling is only for console programs](https://docs.microsoft.com/en-us/windows/console/ctrl-c-and-ctrl-break-signals).

woggioni: SIGINT is unix-specific but Windows has something similar for console applications (like corda) when you use CTRL+C, I think they are called "events"

chrisr3: > SIGINT is unix-specific but Windows has something similar for console applications (like corda) when you use CTRL+C, I think they are called "events"

I think that's what I already said. But AFAIK the Windows JVM is _not_ a console application. If it were, wouldn't it open a console when run?

chrisr3 (approved):
LGTM :+1:

PR #402, opened 2021-03-19
Fixed race condition on jar extraction

woggioni:
Every flask process extract its own jar dependencies in a shared cache; in the cache folder the path to each jar is `lib/${SHA256_OF_THE_JAR}/${JAR_FILE_BASENAME}.jar`; before starting the extraction the process checks that the path doesn't already exist to avoid harming other processes that may be already running and using that jar (rewriting the file with one with identical content could break java processes reading from that file). The problem is that the jar may already exist in the cache because another process created it, but that process might still be writing the jar to the disk.

The result is that a Java process starts with the assumption that all of its dependencies are already in place while they might actually be incomplete (because extracting a big jar to the disk takes some time).

To address this, I added an extraction lock that must be exclusively held by every process during the extraction (and while evaluating which jars need to be extracted), so that only a single process can add files to the cache at any given time.

### DIFF
--- a/flask/flask-launcher/src/main/java/net/corda/flask/launcher/JarCache.java
+++ b/flask/flask-launcher/src/main/java/net/corda/flask/launcher/JarCache.java
@@ -7,7 +7,6 @@ import net.corda.flask.common.LockFile;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.BufferedOutputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -96,6 +95,8 @@ class JarCache {
     @Getter
     private final Path lockFile;
 
+    private final Path extractionLock;
+
     private final Map<String, Path> extractedLibraries;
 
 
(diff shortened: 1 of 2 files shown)


flask/flask-launcher/src/main/java/net/corda/flask/launcher/JarCache.java
```diff
@@ -114,28 +117,30 @@ public JarCache(String appName) {
...
+                    extractedLibraries.put(hash, destination);
+                    if (!Files.exists(destination)) {
+                        Files.createDirectories(destination.getParent());
+                        InputStream is = cl.getResourceAsStream(jarEntryName);
+                        log.debug("Extracting '{}' to '{}'", jarEntryName, destination);
+                        if (is != null) {
+                            try {
+                                try (OutputStream os = new BufferedOutputStream(Files.newOutputStream(destination))) {
+                                    Flask.write2Stream(is, os, buffer);
+                                }
```
chrisr3, line 137: Do we need to use a `BufferedOutputStream` here? We already have a `buffer[]`.

woggioni: correct, we don't need it!

chrisr3 (approved):
LGTM :+1:

PR #405, opened 2021-04-03
CORE-1134: Replace cordformation DSL Closure elements with Action.

### LINKED ISSUE #403: Cordformation Plugin - Kotlin DSL
I managed to create a `deployNodes` task in Kotlin DSL following the examples from the corda examples repo. See the following example:

```
tasks.register<Cordform>("deployNodes") {
    dependsOn.add("jar")

    val notaryConfig = mapOf("validating" to false)

    val rpcUsersConfig = listOf(
        mapOf(
            "user" to "user1",
            "password" to "test",
            "permissions" to listOf("ALL")
        )
    )

    // nodeDefaults not working because of protected setter
    nodeDefaults = closureOf<net.corda.plugins.Node> {
        projectCordapp(closureOf<net.corda.plugins.Cordapp> {
            deploy = false
        })

        runSchemaMigration = true
    }

    node {
        name("O=Notary,L=London,C=GB")
        p2pPort(60000)

        notary = notaryConfig

        rpcSettings(closureOf<RpcSettings> {
            address("localhost:60001")
            adminAddress("localhost:60002")
        })
        true.also { runSchemaMigration = it }
    }

    node {
        name("O=party1,L=London,C=GB")
        p2pPort(10000)

        rpcSettings(closureOf<RpcSettings> {
            address("localhost:10001")
            adminAddress("localhost:10002")
        })

        rpcUsersConfig.also { rpcUsers = it }
        true.also { runSchemaMigration = it }
    }
}
```

This works fine. The only issue is that I can not use the `nodeDefaults` field because it has a protected setter.

See: [Baseform.kt](https://github.com/corda/corda-gradle-plugins/blob/401a43398ca83175818d87261fa62c48b3352c2f/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt#L83-L85)

If there is no reason why this setter should stay protected, I can provide a pull request which makes it public.

chrisr3:
Replace `Closure` in `cordformation`'s DSL with modern `Action`. This makes `cordformation` compatible with Gradle's Kotlin DSL as well as with Groovy.

This should fix #403.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@
 
 ### Version 5.0.14
 
+* `cordformation`: Replace `Closure` with `Action` in the DSL.
+
 ### Version 5.0.13
 
 * `cordformation`: Parsing for adding external service to docker-compose
--- a/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Baseform.kt
@@ -14,8 +14,8 @@ import org.gradle.api.plugins.JavaPluginConvention
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Nested
-import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.SourceSet.MAIN_SOURCE_SET_NAME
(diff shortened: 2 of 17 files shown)


palade (approved):

PR #406, opened 2021-04-06
CORE-1134: Replace cordformation DSL Closure elements with Action. (#405)

chrisr3:
* Replace cordformation DSL Closure elements with Action.
* Remove unused imports.

Forward ported from #405.


palade (approved):

PR #407, opened 2021-04-07
CORE-145: Remove long-deprecated DSL feature from Cordformation.

chrisr3:
The `node.cordapps = []` elements was supposed to have been deleted from Cordformation already, but has somehow survived. But no longer! It will not trouble the Corda Gradle plugins' 6.x series.


palade (approved):
The proposed changes look good to me.

PR #408, opened 2021-04-08
NOTICK: Annotate Baseform.signing field as "Nested".

chrisr3:
The `Baseform.signing` field is not an `@Input` property, but its contents are. Annotate `signing` as `@Nested` to instruct Gradle to examine `signing` for task properties.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 ### Version 5.0.14
 
 * `cordformation`: Replace `Closure` with `Action` in the DSL.
+* `cordformation`; Annotate `Baseform.signing` as `@Nested` instead of `@Input`.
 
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -5,6 +5,7 @@
 ### Version 5.0.14
 
 * `cordformation`: Replace `Closure` with `Action` in the DSL.
+* `cordformation`; Annotate `Baseform.signing` as `@Nested` instead of `@Input`.
```
palade, line 8:
```suggestion
* `cordformation`: Annotate `Baseform.signing` as `@Nested` instead of `@Input`.
```

palade (approved):
LGTM! Left one small comment :grin:

PR #409, opened 2021-04-08
NOTICK Fix shutdown hook execution on flask jars

woggioni:
The current version of flask calls `Process.destroyForcibly` on the child process immediately after calling `Process.destroy` and without waiting for a graceful shutdown of the child process.
I've also added the `net.corda.flask.kill.timeout.millis` JVM property to set the timeout before forcibly killing the child process if it doesn't gracefully terminate.

### DIFF
--- a/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
+++ b/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
@@ -14,6 +14,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
+import java.util.concurrent.TimeUnit;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 import java.util.stream.Collectors;
@@ -152,18 +153,40 @@ public class Launcher {
             }
             builder.getCliArgs().addAll(args);
             beforeChildJvmStart(builder);
-            try(LockFile processLock = LockFile.acquire(cache.getPidFile(), false)) {
-                builder.getProperties().put(Flask.JvmProperties.PID_FILE, cache.getPidFile());
-                Path heartbeatAgentPath = extractedLibraries.get(
-                        manifest.getMainAttributes().getValue(Flask.ManifestAttributes.HEARTBEAT_AGENT_HASH));
-                builder.getJvmArgs().add("-javaagent:" + heartbeatAgentPath);
-                int returnCode = builder.exec();
-                afterChildJvmExit(returnCode);
+            builder.getProperties().put(Flask.JvmProperties.PID_FILE, cache.getPidFile());
+            Path heartbeatAgentPath = extractedLibraries.get(
+                    manifest.getMainAttributes().getValue(Flask.ManifestAttributes.HEARTBEAT_AGENT_HASH));
+            builder.getJvmArgs().add("-javaagent:" + heartbeatAgentPath);
+            LockFile processLock = LockFile.acquire(cache.getPidFile(), false);
+            Process process = builder.build().inheritIO().start();
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                if(process.isAlive()) {
+                    process.destroy();
+                    try {
+                        long timeout = Long.parseLong(
+                                System.getProperty(
+                                        Flask.JvmProperties.KILL_TIMEOUT_MILLIS,
+                                        Flask.Constants.DEFAULT_KILL_TIMEOUT_MILLIS));
+                        process.waitFor(timeout, TimeUnit.MILLISECONDS);
+                    } catch (InterruptedException ie) {
+                        ie.printStackTrace();
+                    } finally {
+                        if (process.isAlive()) {
+                            process.destroyForcibly();
+                        }
+                        processLock.close();
+                    }
+                }
+                try {
+                    Files.delete(cache.getPidFile());
+                } catch (Exception ex) {
+                    ex.printStackTrace();
+                }
                 cache.touchLibraries();
-                return returnCode;
-            } finally {
-                Files.delete(cache.getPidFile());
-            }
+            }));
+            int returnCode = process.waitFor();
+            afterChildJvmExit(returnCode);
+            return returnCode;
         }
     }
 
--- a/flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
+++ b/flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
@@ -1,16 +1,20 @@
 package net.corda.gradle.flask
 
+
 import groovy.transform.CompileStatic
 import net.corda.flask.common.Flask
 import org.gradle.testkit.runner.GradleRunner
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.Timeout
 import org.junit.jupiter.api.io.TempDir
 
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption
 import java.security.MessageDigest
+import java.util.concurrent.TimeUnit
 import java.util.zip.ZipEntry
 import java.util.zip.ZipInputStream
 
@@ -19,6 +23,8 @@ import static net.corda.flask.common.Flask.Constants.ZIP_ENTRIES_DEFAULT_TIMESTA
 import static org.assertj.core.api.Assertions.assertThat
 import static org.junit.jupiter.api.Assertions.assertArrayEquals
 import static org.junit.jupiter.api.Assertions.assertEquals
+import static org.junit.jupiter.api.Assertions.assertNotEquals
+import static org.junit.jupiter.api.Assertions.assertTrue
 
 @CompileStatic
 class FlaskPluginTest {
@@ -70,6 +76,7 @@ class FlaskPluginTest {
         installResource("testProject", "settings.gradle", testProjectDir)
         installResource("testProject","src/main/java/net/corda/gradle/flask/test/Main.java", testProjectDir)
         installResource("testProject","src/main/java/net/corda/gradle/flask/test/AlternativeMain.java", testProjectDir)
+        installResource("testProject","src/main/java/net/corda/gradle/flask/test/HangingMain.java", testProjectDir)
         installResource("testProject","src/flask/java/net/corda/gradle/flask/test/TestLauncher.java", testProjectDir)
         installResource("testProject", "testAgent/build.gradle", testProjectDir)
         installResource("testProject", "testAgent/src/main/java/net/corda/gradle/flask/test/agent/JavaAgent.java", testProjectDir)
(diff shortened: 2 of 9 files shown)


flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
```diff
@@ -152,18 +154,39 @@ final int launch(Manifest manifest, List<String> jvmArgs, List<String> javaAgent
...
+                                System.getProperty(
+                                        Flask.JvmProperties.KILL_TIMEOUT_MILLIS,
+                                        Flask.Constants.DEFAULT_KILL_TIMEOUT_MILLIS));
+                        process.waitFor(timeout, TimeUnit.MILLISECONDS);
+                    } catch (InterruptedException ie) {
+                        //Ignore
+                    }
+                    if (process.isAlive()) {
+                        process.destroyForcibly();
+                    }
```
chrisr3, line 177: Maybe:
```java
} finally {
    if (process.isAlive()) {
        process.destroyForcibly();
    }
}
```
Just to be sure.

flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
```diff
@@ -167,4 +174,32 @@ class FlaskPluginTest {
         }
         assertEquals("net.corda.gradle.flask.test.AlternativeMain", prop.mainClassName)
     }
+
+    @Test
+    @DisplayName("Check that shutdown hooks are executed on flask jars")
+    void shutdownHookTest() {
```
chrisr3, line 180: Should this test have its own `@org.junit.jupiter.api.Timeout`, just in case something goes wrong?

flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
```diff
@@ -167,4 +174,32 @@ class FlaskPluginTest {
...
+            totalWait += loopWait
+            if(totalWait > timeoutMillis) {
+                throw new AssertionError("Failed to create test file '$testFile'")
+            }
+            Thread.sleep(loopWait)
+        }
+        process.destroy()
+        int returnCode = process.waitFor()
+        assertNotEquals(0, returnCode)
+        assertTrue(!Files.exists(testProjectDir.resolve("shutdown-hook-executed")))
```
chrisr3, line 203: AssertJ has an `assertThat()` that can provide a better assertion for a `Path` than `assertTrue()`.

chrisr3 (commented):
Simple things, just from inspection.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
```diff
@@ -152,18 +154,40 @@ final int launch(Manifest manifest, List<String> jvmArgs, List<String> javaAgent
...
+                if(process.isAlive()) {
+                    process.destroy();
+                    try {
+                        long timeout = Long.parseLong(
+                                System.getProperty(
+                                        Flask.JvmProperties.KILL_TIMEOUT_MILLIS,
+                                        Flask.Constants.DEFAULT_KILL_TIMEOUT_MILLIS));
+                        process.waitFor(timeout, TimeUnit.MILLISECONDS);
+                    } catch (InterruptedException ie) {
+                        log.error(ie.getMessage(), ie);
```
chrisr3, line 173: This is a shutdown hook, and so the SLF4J `Logger` is probably shutting down too. It's probably better just to write to `stderr` here.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
```diff
@@ -152,18 +154,40 @@ final int launch(Manifest manifest, List<String> jvmArgs, List<String> javaAgent
...
+                        if (process.isAlive()) {
+                            process.destroyForcibly();
+                        }
+                        processLock.close();
+                    }
+                }
+                try {
+                    Files.delete(cache.getPidFile());
+                } catch (IOException ex) {
+                    log.warn(ex.getMessage(), ex);
```
chrisr3, line 184: Again, we cannot rely on `log` still to be running at this point.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
```diff
@@ -152,18 +154,40 @@ final int launch(Manifest manifest, List<String> jvmArgs, List<String> javaAgent
...
+                    } finally {
+                        if (process.isAlive()) {
+                            process.destroyForcibly();
+                        }
+                        processLock.close();
+                    }
+                }
+                try {
+                    Files.delete(cache.getPidFile());
+                } catch (IOException ex) {
```
chrisr3, line 183: Does this also need to be inside the `try-finally` block? An `InterruptedException` will currently skip these lines. Alternatively, you could provide the thread with an `UncaughtExceptionHandler`.,

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
```diff
@@ -152,18 +154,40 @@ final int launch(Manifest manifest, List<String> jvmArgs, List<String> javaAgent
...
+                    log.warn(ex.getMessage(), ex);
+                }
+            }));
+            int returnCode = process.waitFor();
+            afterChildJvmExit(returnCode);
+            cache.touchLibraries();
```
chrisr3, line 189: Do we want
```java
int returnCode;
try {
    returnCode = process.waitFor();
    afterChildJvmExit(returnCode);
} finally {
    cache.touchLibraries();
}
```
here? Because `waitFor()` can throw `InterruptedException`, which would skip the `cache.touchLibraries()` step.

Does Flask also need an `afterChildJvmError(Exception e)` handler? (We cannot expect to handle `Error` or `Throwable`.)

chrisr3 (changes requested):
Some extra thoughts about a clean shutdown.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
woggioni: it isn't super-important to execute `cache.touchLibraries();` (the worst thing that can happen is that those libraries will be evicted from the cache and will need to be extracted again the next time the program runs). I can simply move that step at the end of the shutdown hook. 
I don't think we can create an `afterChildJvmError(Exception e)` handler since we cannot access exceptions thrown by the child process (it isn't the case for flask but a child process might not even be a Java process at all)

chrisr3: I'm not suggesting handling exceptions thrown _by_ the child process. I am concerned about the case where `process.waitFor()` throws an exception and so skips the `afterChildJvmExit()` hook.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
woggioni: I do not think that `Files.delete` can ever throw `InterruptedException`, the goal was to not block the execution of the code if deletion of the pidfile fails (which doesn't make much sense atm since that statement is already the last statement in its thread). I will add `cache.touchLibraries();` after the deletion of the pidfile and change the exception type in the catch clause to `Exception`

chrisr3: My bad - I was thinking that any `InterruptedException` thrown by `process.waitFor()` would propagate out of the handler without reaching `Files.delete` :facepalm:.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
```diff
@@ -6,6 +6,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
```
chrisr3, line 9: This import is unused.

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
woggioni: ok, but the only exception that can be thrown by `process.waitFor` is `InterruptedException` when the thread is interrupted by another thread, but there is no other thread in Flask that will ever attempt such a thing...

flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
```diff
@@ -2,15 +2,20 @@ package net.corda.gradle.flask
 
 import groovy.transform.CompileStatic
 import net.corda.flask.common.Flask
+import org.gradle.internal.jvm.Jvm
```
chrisr3, line 5: Unused import

flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
```diff
@@ -2,15 +2,20 @@ package net.corda.gradle.flask
...
 import org.gradle.testkit.runner.GradleRunner
 import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.Timeout
 import org.junit.jupiter.api.io.TempDir
 
 import java.nio.file.Files
 import java.nio.file.Path
+import java.nio.file.Paths
```
chrisr3, line 15: Another unused import

flask/src/test/groovy/net/corda/gradle/flask/FlaskPluginTest.groovy
```diff
@@ -167,4 +175,33 @@ class FlaskPluginTest {
...
+            .start()
+        Path testFile = testProjectDir.resolve("shutdown-hook-executed")
+        int totalWait = 0
+        final int timeoutMillis = 5000
+        while(!Files.exists(testFile)) {
+            int loopWait = 100
+            totalWait += loopWait
+            if(totalWait > timeoutMillis) {
+                throw new AssertionError("Failed to create test file '$testFile'")
+            }
```
chrisr3, line 199: Maybe:
```groovy
assertTrue(totalWait < timeoutMillis, "Failed to create test file '$testFile'")
```

chrisr3 (changes requested):
Apart from more unused imports, I have noticed that `FlaskPlugin` is using the deprecated `JavaApplication.mainClassName` property. We should probably use its replacement [JavaApplication.mainClass](https://docs.gradle.org/6.7.1/dsl/org.gradle.api.plugins.JavaApplication.html#org.gradle.api.plugins.JavaApplication:mainClass).

flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
woggioni: this why I hate checked exceptions... they only reason I wrote that 
```
} catch (InterruptedException ie) {
```
line is because the compiler forced me to do so, I do not expect that exception to ever happen

woggioni:
well it really is a property of `net.corda.gradle.flask.FlaskJarTask` it doesn't share anything (except for the name) with `JavaApplication.mainClassName`

chrisr3:
> well it really is a property of `net.corda.gradle.flask.FlaskJarTask` it doesn't share anything (except for the name) with `JavaApplication.mainClassName`

I am referring to this:
```groovy
project.pluginManager.withPlugin('application') {
    String result = project.extensions.findByType(JavaApplication.class)?.mainClassName
    if(!result) throw new GradleException("mainClassName property from \"${JavaApplication.class.name}\" extension is not set")
    mainClassName = result
}
```
which tries to read the value of the **deprecated** `JavaApplication::mainClassName` property. This will break once Gradle finally deletes the property entirely.

woggioni:
oh, thanks! sorry I had completely forgotten that piece of code... tbh I'm more of the idea of removing it completely. I originally wrote it so that that you could seamlessly add flask to a Gradle project that already uses the `application` plugin and see it working without having to configure anything. 

The documentation clearly states you should explicitly configure the main class name of the `flaskJar` task and the current "clever" behavior isn't even documented, apart from that I don't think are there many use cases (if any) where you want to apply both the flask plugin and the application plugin to the same project

chrisr3:
> tbh I'm more of the idea of removing it completely.

And its now-unused imports...? :grin:

chrisr3 (approved):

chrisr3:
LGTM :+1:

PR #412, opened 2021-04-12
NOTICK: Fix Gradle deprecations that will become errors later.

chrisr3:
Fix errors and warnings that are highlighted by Gradle 7.0. (Note that we are still using Gradle 6.7.1, but should be upgrading to Gradle 6.8.3 soon.)


woggioni (approved):
LGTM

PR #413, opened 2021-04-12
CORE-268: Create cordapp-cpk Gradle plugin.

chrisr3:
Provide a new `cordapp-cpk` Gradle plugin to create CPK CorDapps, and update `cordformation` to deploy CPKs for Corda 5.

**NOTE: You will need Artifactory credentials to run the `cordformation` plugin's tests.**

Main PR features are:
- The `cordapp-cpk` plugin integrates with the `java-library` plugin and Bnd's `builder` plugin to create the CPK's "main" jar as an OSGi bundle.
- `cordapp-cpk` creates new Gradle configurations `cordaProvided`, `cordaPrivateProvided`, `cordapp`, `cordaRuntimeOnly` and `cordaEmbedded` to control how CPK dependencies are managed. (Some of these configurations are shared with the `cordformation` plugin.)
- The existing (deprecated) `cordaCompile` and `cordaRuntime` configurations have been removed.
- The "main" jar is also packaged into a "jar-like" `cpk` container, along with its non-Corda runtime dependencies (excluding other CPKs). The CPK artifact is assigned a classifier of `cordapp`.
- The non-CPK runtime dependencies are listed inside an XML document `META-INF/DependencyConstraints` inside the "main" jar.
- The "main" jar's CPK dependencies are listed inside another XML document `META-INF/CPKDependencies`.
- Bnd generates OSGi `Export-Package` metadata for all of the "main" jar's packages by default. However, this behaviour can be customised via a new `osgi {}` DSL extension on the `jar` task itself.
- Both "main" jar and CPK are signed using `jarsigner`. The signing key is configurable via the `cordapp {}` Gradle DSL extension (as per the existing `cordapp` plugin).
- The "main" jar's OSGi metadata is verified using Bnd and a new `VerifyBundle` task.
- On publication, the `cordapp-cpk` plugin also publishes a "companion" POM that contains the CPK's `cordapp` and `cordaProvided` dependencies. The plugin can use these "companion" POMs to support transitive dependencies on other CPKs. Publishing these to Artifactory is also supported.
- Requires Gradle >= 6.6, and supports "lazy" Gradle properties and task configuration avoidance.

- The `cordformation` plugin has been updated always to deploy the `type="cpk"`, `classifier="cordapp"` artifact for a given `"group:name:version"` Maven coordinate.

And there are lots and lots of test cases too :grinning:.


thschroeter (approved):
I haven't reviewed in depth but I believe this change should be merged.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
```diff
@@ -0,0 +1,268 @@
...
+            } else {
+                "$first-$second"
+            }
+        }
+
+        fun parsePackages(value: String): Set<String> {
+            return value.split(",").mapTo(LinkedHashSet(), String::trim)
+        }
+
+        fun dynamic(value: String): String = "$value;resolution:=dynamic;version=!"
```
palade, line 86: What does `version=!`  do?

palade (commented):
The proposed changes look to do me. Also, I noticed that this is post merge, I left a small question below.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
chrisr3: It deletes the version attribute, should there be one. I didn't want a "dynamic import" to include a version.

PR #414, opened 2021-04-13
CORE-1161: Upgrade to Gradle 6.8.3.

chrisr3:
Plugins written in Kotlin are still compiled to use the Kotlin 1.3 API, despite Gradle 6.8.3 using the Kotlin 1.4 runtime libraries. This should allow Gradle versions < 6.8 to continue using them.

JarFilter uses the `kotlinx-metadata-jvm` library, the latest version of which now _requires_ Kotlin 1.4. So JarFilter 6 now requires Gradle 6.8+. (This shouldn't be an issue for Corda.)

Also upgrade the following:
- JUnit 5.7.0 -> 5.7.1
- Artifactory plugin 4.20.0 -> 4.21.0
- Gradle Publish plugin 0.12.0 -> 0.14.0
- Gradle Enterprise Plugin 3.6 -> 3.6.1

### DIFF
--- a/jar-filter/build.gradle
+++ b/jar-filter/build.gradle
@@ -13,8 +13,9 @@ repositories {
 }
 
 ext {
-    kotlin_metadata_version = '0.1.0'
-    test_kotlin_api_version = '1.3'
+    kotlin_metadata_version = '0.2.0'
+    test_kotlin_api_version = '1.4'
+    test_kotlin_version = '1.4.32'
 }
 
 gradlePlugin {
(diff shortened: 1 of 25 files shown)


jar-filter/build.gradle
```diff
@@ -13,8 +13,9 @@ repositories {
 }
 
 ext {
-    kotlin_metadata_version = '0.1.0'
-    test_kotlin_api_version = '1.3'
+    kotlin_metadata_version = '0.2.0'
+    test_kotlin_api_version = '1.4'
+    test_kotlin_version = '1.4.32'
```
palade, line 18: Why there is a need for `test_kotlin_version` and why is different from `kotlin_version` that is defined in `gradle.properties`?

palade (approved):
The proposed changes look good to me. I left a question related to one particular change.

jar-filter/build.gradle
chrisr3: The `kotlin_version` property is the version of Kotlin that the plugin _itself_ is built with. This must be compatible with the versions of Gradle that the plugin is expected to be used by. For example, Gradle 6.8.3 uses the Kotlin 1.4.20 runtime library and its Kotlin DSL uses the Kotlin 1.3 API.

In contrast, `test_kotlin_version` is the version of Kotlin that I compile all of the test classes with. These are the classes that the `jar-filter` plugin will transform, and are expected to be Kotlin 1.4. In fact, the latest version of the `kotlinx-metadata-jvm` library only supports Kotlin 1.4+ metadata, which means that Kotlin < 1.4 will need to use the plugin from the `release/5.0` branch.

PR #422, opened 2021-05-06
CORE-1300: Copy CPKDependencies into the CPK file.

chrisr3:
Modify the `cordapp-cpk` plugin's `PackagingTask` to extract the `CPKDependencies` file from the "main" jar and include it inside the `.cpk` file too.

### DIFF
--- a/cordapp-cpk/src/test/kotlin/net/corda/plugins/cpk/GradleProject.kt
+++ b/cordapp-cpk/src/test/kotlin/net/corda/plugins/cpk/GradleProject.kt
@@ -21,6 +21,7 @@ import org.gradle.util.GradleVersion
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.TestReporter
 import java.io.IOException
+import java.io.InputStream
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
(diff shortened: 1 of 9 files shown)


cordapp-cpk/src/test/kotlin/net/corda/plugins/cpk/GradleProject.kt
```diff
@@ -33,7 +34,7 @@ import kotlin.test.fail
 const val expectedCordappContractVersion = 2
 const val expectedCordappWorkflowVersion = 3
 const val expectedCordappServiceVersion = 4
-const val cordaReleaseVersion = "4.6"
+const val cordaReleaseVersion = "4.8"
```
palade, line 37: Probably very obvious answer, but why is this attribute "4.*" ?

palade (approved):
I left one small but other than that LGTM 👍

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpk/GradleProject.kt
chrisr3: Because no "stable" artifacts exist yet for Corda 5.x :grinning:. All I need here is for some `net.corda` artifacts from a real Maven repository to test that the `cordapp-cpk` plugin _refuses_ to use them.

It also helps that the Corda 4.x repository is public...

PR #425, opened 2021-05-11
CORE-1335: Validate the CPK XML against a schema XSD.

chrisr3:
- Create a schema XSD describing the CorDapp `CPKDependencies` and `DependencyConstraints` XML documents, and update tests to validate them both against it.
- Prefix XML namespace with `urn:` to make a valid URI (as required).
- Apply "best practices" for JAXP.
- Rewrite JAXP factories in Java to work around a problem with the Kotlin compiler.

### DIFF
--- a/cordapp-cpk/build.gradle
+++ b/cordapp-cpk/build.gradle
@@ -96,6 +96,10 @@ processTestResources {
     }
 }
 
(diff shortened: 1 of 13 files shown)


palade (dismissed):

palade (approved):

PR #426, opened 2021-05-20
CORE-1390: Compute the highest platform version used by the CorDapp.

chrisr3:
- Compute `Cordapp-Built-Platform-Version` manifest tag for each "main" jar.
- Copy `Cordapp-Built-Platform-Version` into CPK manifest as `Corda-CPK-Built-Platform-Version`.
- Remove any transitive dependencies belonging to dependent CPKs from our compile classpath. This
allows us to simplify how we create the Gradle configurations.

Also extend the test coverage to examine the compile classpaths when building CorDapps.

### DIFF
--- a/cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/DependencyCalculator.kt
+++ b/cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/DependencyCalculator.kt
@@ -20,6 +20,7 @@ import org.gradle.api.tasks.TaskAction
 import org.gradle.api.tasks.TaskDependency
 import java.io.File
 import java.util.Collections.unmodifiableList
+import java.util.Collections.unmodifiableSet
 import javax.inject.Inject
 
 @Suppress("UnstableApiUsage")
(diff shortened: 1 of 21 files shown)


cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/DependencyCalculator.kt
```diff
@@ -28,6 +29,17 @@ open class DependencyCalculator @Inject constructor(objects: ObjectFactory) : De
         private const val NON_CORDA = false
         private const val CORDA = true
 
+        private val HARDCODED_EXCLUDES: Set<Pair<String, String>> = unmodifiableSet(setOf(
```
palade, line 32: why does it need to be wrapped in an `unmodifiableSet` ?

palade (approved):
The proposed changes look good to me.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/DependencyCalculator.kt
chrisr3: Although a Kotlin `Set` cannot be modified, it is still a modifiable Java set underneath. I just want to make it _absolutely clear_ that this set is a constant.

PR #427, opened 2021-05-22
CORE-1434: Warn if `@Suspendable` not configured for quasar-utils.

chrisr3:
- Log a warning if the `quasar-utils` plugin has not been configured with an explicit `@Suspendable` annotation - either via the `quasar_suspendable_annotation` property or via the `quasar` DSL extension. This warning will only be displayed once per project.
- Ensure that the `cordaProvided` and `cordaRuntimeOnly` configurations defined by `quasar-utils` exactly match the same configurations created by the `cordapp-cpk` plugin. The point is that it shouldn't matter which plugin creates these configurations in practice.


palade (approved):
LGTM!

PR #429, opened 2021-05-28
CORE-268: Improve message for CPK validation errors.

chrisr3:
Bnd generates the validation error messages, but we can add some general advice afterwards about the probable cause of the underlying problems.


joeldudleyr3 (approved):
LGTM

PR #431, opened 2021-06-03
CORE-1589: Verify CPK OSGi imports match classpath exports.

chrisr3:
Add an extra validation step to the CPK `verifyBundle` task to check that the package imports align with the package exports. Also resolve the (`cordapp`, `cordaProvided`) configurations separately from `compileClasspath` for this, so that their transitive dependencies are not affected by the contents of the CPK. (We need this to make the `verifyBundle` task meaningful.)


palade (approved):

PR #436, opened 2021-06-16
NOTICK patched Cordformation

woggioni:
- added preliminary support for `httpRcpSettings` in `node.conf`
- assembled a CPB archive out of all CPKs in the `cordapp` configuration

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/HttpRpcSettings.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/HttpRpcSettings.kt
@@ -1,0 +1,44 @@
+package net.corda.plugins
+
+import com.typesafe.config.Config
+import com.typesafe.config.ConfigFactory
+import com.typesafe.config.ConfigValueFactory
+import org.gradle.api.tasks.Input
+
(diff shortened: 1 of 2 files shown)


chrisr3:
AFAIK the node's "network bootstrapper" has just been changed in a way that might break Cordformation. The `BaseformTest` class currently contains this SNAPSHOT version of Corda:
```kotlin
private const val cordaVersion = "1623057204977-rc"
```
Can you update this to the latest `-rc` SNAPSHOT version, and then re-run the tests please? There is no point making any further changes to Cordformation for the sake of an obsolete version of Corda 5..

woggioni:
the problem is that I have to amend unit tests as well.. some of them expect a cpk file to be present in the cordapps directory (which isn't the case anymore)

chrisr3:
> the problem is that I have to amend unit tests as well.. some of them expect a cpk file to be present in the cordapps directory (which isn't the case anymore)

In which case, it's probably not worth trying to fix Cordformation.

chrisr3:
The `cordformation` plugin is incompatible with Corda 5, and has been retired.

PR #439, opened 2021-06-28
Optimize NetworkBootstrapper performance when generating member infos

Simon-Dobson:

The goal of this change is to optimize the Network Bootstrapper performance. Profiling node startup initiated by bootstrapper to run migration scripts and then to generate member infos indicates around 40% of the time is spent in class instrumentation. When running without the quasar agent we see a significant improvement in bootstrapper performance. It appears that quasar instrumentation is not necessary when nodes are started briefly to run specific subcommands - such as to migrate the db and/or generate member infos. 

The proposal is to allow callers to disable the quasar agent by setting System property net.corda.flask.launcher.disableQuasarAgent to true (default is false) this will 

# PR Checklist:

- [ ] Have you run the unit, integration and smoke tests as described here? https://docs.corda.net/head/testing.html
- [ ] If you added/changed public APIs, did you write/update the JavaDocs?
- [ ] If the changes are of interest to application developers, have you added them to the changelog, and potentially release notes?
- [ ] If you are contributing for the first time, please read the agreement in CONTRIBUTING.md now and add to this Pull Request that you agree to it.

Thanks for your code, it's appreciated! :)

### DIFF
--- a/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
+++ b/flask/flask-launcher/src/main/java/net/corda/flask/launcher/Launcher.java
@@ -123,9 +123,18 @@ public class Launcher {
             JavaProcessBuilder builder = new JavaProcessBuilder();
             builder.setMainClassName(Optional.ofNullable(System.getProperty(Flask.JvmProperties.MAIN_CLASS))
                     .orElse(manifest.getMainAttributes().getValue(Flask.ManifestAttributes.APPLICATION_CLASS)));
+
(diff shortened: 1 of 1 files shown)


woggioni:
well.. I don't think this is the right place for this change as this plugin is completely unaware of both quasar and the network bootstrapper. Otoh it could be possible to simply disable Quasar entirely in the network bootstrapper (and possibly remove it from the jar as well) as I don't think we need fibers to bootstrap a Corda network

Simon-Dobson:
Fair comment. I'll try to make bootstrapper use a corda.jar that doesn't have the Quasar javaAgent

PR #444, opened 2021-07-11
CORE-1921: Ensure we build the jars required by CPKDependencies XML.

chrisr3:
Gradle's dependency graph for the `cordaCPKDependencies` task needs to include the `jar` tasks for any transitive CorDapp  projects. Replace the build dependencies from the `runtimeClasspath` and `compileClasspath` configurations with the `cordapp-cpk` plugin's `cordappPackaging` and `cordappExternal` configurations respectively, which are their "OSGi-aware" replacements.

This has also exposed a more subtle problem where Gradle sometimes resolves dependent CorDapp projects to their `build/classes/*/main` directories instead of their "main" jars. This is _**toxic**_ to OSGi because it deprives Bnd of those CorDapps' OSGi metadata, which in turn means that Bnd generates broken OSGi metadata for the new "main" jar. I have _hopefully_ resolved this by setting the `LibraryElements` variant attribute to `JAR` on each CorDapp dependency.


palade:

palade (approved):

woggioni (approved):

PR #446, opened 2021-07-15
CORE-1998: Ensure that cordapp-cpk plugin always compiles against jars.

chrisr3:
Bnd _must_ compile against jars, and not against directories of classes. Set the `LibraryElements.JAR` variant attribute on the correct resolvable configurations to ensure that Gradle does this.

Bnd sets Gradle's `LibraryElements.JAR` attribute on the `compileClasspath` configuration for us already, but there's no harm in us setting it too. We also need to set this attribute on the `cordappExternal` configuration in order to be consistent with `compileClasspath`, and to configure any "detached configuration" used when tracing transitive CorDapp dependencies.

### DIFF
--- a/cordapp-cpk/build.gradle
+++ b/cordapp-cpk/build.gradle
@@ -10,6 +10,7 @@ plugins {
 description 'Configures this project to create a CPK CorDapp'
 
 ext {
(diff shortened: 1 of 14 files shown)


woggioni (approved):
LGTM

PR #447, opened 2021-07-16
[CORE-1942] Added cpb plugin

woggioni:
I have something kind of working.. the only issue is that I do not understand why it works (I still have to implement exclusion of `corda-api` CPKs)


cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
```diff
@@ -0,0 +1,52 @@
...
+                            it.name = dependency.name
+                            it.classifier = "cordapp"
+                            it.type = "cpk"
+                        }
+                    }
+                    is ProjectDependency -> {
+                        dependency.targetConfiguration = CORDA_CPK_CONFIGURATION_NAME
+                    }
+                }
+            }
```
woggioni, line 48: If I move this bit to line 31, outside of the `register` closure, the assembled CPB contains jar files instead of CPKs.
This makes no sense at all to me, as far as I understand, the content of the `register` closure (except for eventual `doLast`, `doFirst` callbacks) is also executed in the configuration phase and immediately after line 30

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
```diff
@@ -0,0 +1,52 @@
...
+    companion object{
+        private const val CPB_TASK_NAME = "cpb"
+        private const val CPB_CONFIGURATION_NAME = CPB_TASK_NAME
+        private const val CPB_ARTIFACT_CLASSIFIER = "package"
+        private const val CPB_FILE_EXTENSION = CPB_TASK_NAME
+    }
+
+    override fun apply(project: Project) {
+        project.plugins.findPlugin(CordappPlugin::class.java) ?:
+            throw GradleException("The 'net.corda.plugins.cpb' plugin needs the 'net.corda.plugins.cordapp-cpk' plugin to be applied first")
```
chrisr3, line 26: Just use
```kotlin
project.pluginManager.apply('net.corda.plugins.cordapp-cpk')
```
This will either apply the `cordapp-cpk` plugin, or do nothing if it has already been applied.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        val cpbConfiguration : Configuration = project.configurations.create(CPB_CONFIGURATION_NAME)
+        val cordappConfiguration = project.configurations.named(CORDAPP_CONFIGURATION_NAME).get()
```
chrisr3, line 29: Easier to use `getByName(CORDAPP_CONFIGURATION_NAME)`.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
chrisr3: Lines inside the closure will only be executed if/when the task is configured, whereas lines outside the closure will be executed as soon as the plugin is applied.

Don't forget that you need to use the `allCordapps` configuration instead of `cordapps` in order to collect all transitive CPKs, but this configuration will not be populated immediately. (See `Coniguration.withDependencies()`.)

You should also be aware that "transitive CPK resolution" is not guaranteed to work (unfortunately), which is another reason why we also need to be able to add CPKs to the `cpb` configuration manually.

chrisr3 (commented):
I have added some points to think about. We need Gradle to _defer_ resolving the `cpb` configuration until Gradle's configuration phase is complete, so you need to be careful not to invoke any method which will force Gradle to resolve it early.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/CordappPlugin.kt
```diff
@@ -125,10 +125,6 @@ class CordappPlugin @Inject constructor(private val layouts: ProjectLayout): Plu
                         // to our own compile classpath.
                         // WE ARE MUTATING THESE DEPENDENCIES FOR EVERY CONFIGURATION THEY APPEAR IN!
                         dep.isTransitive = false
-
-                        // We also need to GUARANTEE that Gradle uses the jar artifact here.
-                        // Only the jar contains the OSGi metadata we need.
-                        dep.attributes(attributor::forJar)
```
woggioni, line 131: if I don't do this I get
```
Caused by: org.gradle.api.InvalidUserCodeException: Cannot set artifact / configuration information on a dependency that has attributes or capabilities configured
```
:-(

chrisr3: Unfortunately we need that line.

chrisr3:
I've pushed a commit to help with the CPK dependency resolving. We _cannot_ remove the variant attribute from the CPKs because that is fixing CORE-1921, and so I have created brand new dependencies instead. (We would have needed to use `Dependency.copy()` before modifying them anyway.)

Another issue was that we need the `allCordapps` configuration's `withDependencies` handler to run before the CPB's `withDependencies` handler, as otherwise `allCordapps` will not contain any transitive CPKs. A `withDependency` handler is not invoked until its configuration is used for dependency resolution, and so I am forcing this to happen with a "throw-away" detached configuration. I'd like to think that there is a more natural way of doing this...

I've also set a lot of flags on the new configurations, e.g. `isVisible`, `isTransitive`, `isCanBeResolved` and `isCanBeConsumed` because Gradle expects such things these days.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/CordappPlugin.kt
chrisr3: I think I now understand what is happening here. We only need to add the variant attribute to a `ProjectDependency` because nothing downloaded from a Maven repository will have a `classes` directory. However, it also appears that we _must not_ add a variant attribute to external dependencies because this would force everyone to publish Gradle metadata into those Maven repositories too :scream:.

I have added some extra tweaks to #446 so that the `cordaCPK` configuration is also "variant aware". This should allow us to handle CPK dependencies more nicely in the `cordapp-cpb` plugin. However, I have also discovered that any `ProjectDependency` created in the `withDependencies()` handler isn't automatically linked to the `cpk` task that builds it! This is a problem that I am frantically investigating.

chrisr3:
> why do we need to extend the cpb configuration if we are adding the dependencies manually at line 55?

Not _quite_ sure what you're asking, but I am assume you are referring to these lines:
```kotlin
project.configurations.detachedConfiguration()
    .setTransitive(false)
    .setVisible(false)
    .extendsFrom(cpbConfiguration)
    .resolvedConfiguration
```
Gradle these days wants `Configuration` obejcts to have specific purposes, e.g. "for resolving dependencies", "for containing artifacts for other projects to use", or  "just to _contain_ dependencies (and do nothing else)".

We need to use the `cpb` configuration in "dependency resolution" so that any `withDependencies` handlers it has are executed. (The `cpb` configuration extends `allCordapps`, and we _definitely_ need the `withDependencies` handler _there_ to run because it finds all of the transitive CPKs!) However, the `cpb` configuration is flagged as `canBeResolved = false` which prevents us from resolving it directly. So we extend it, and resolve the extension instead. Once this has happened, we know that `cpb` will contain deoendencies for _all_ dependent CorDapps, which we can then transform into dependencies on their CPK artifacts.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
```diff
@@ -0,0 +1,120 @@
...
+            .setTransitive(false)
+            .setVisible(false)
+            .withDependencies { dependencies ->
+                // Force Gradle to discover any transitive CPKs
+                // and include them in cpbConfiguration.
+                project.configurations.detachedConfiguration()
+                    .setTransitive(false)
+                    .setVisible(false)
+                    .extendsFrom(cpbConfiguration)
+                    .resolvedConfiguration
```
woggioni, line 52: if I understand correctly, this creates a temporary configuration that is then immediately garbage collected...

woggioni: why don't we resolve `cpbConfiguration` if that is the point of this bit of code?

chrisr3: Because we are not interested in the resolved contents of the `cpb` configuration. The only reason to resolve anything at all is to force Gradle to execute the `withDependencies` handler on `allCordapps`. In theory, we could extend `allCordapps` instead of `cpb` and then resolve that instead. However, I suppose someone _could_ add a `withDependencies` handler to `cpb` at some point, in which case we'd need that to execute that handler too before iterating over `cpb.allDependencies`.

woggioni: but why is the `cpb` configuration not transitive as well? do we want to force developers to state all of their dependencies one by one as opposed to just declare their direct dependencies?

woggioni: oh, no sorry it's the `cpbPackaging` configuration that is not transitive, disregard my previous comment

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
```diff
@@ -0,0 +1,120 @@
...
+                }
+            }.apply {
+                isCanBeConsumed = false
+            }
+
+        project.afterEvaluate {
+            // We MUST resolve the CPB configurations before
+            // Gradle builds the Task Execution Graph!
+            cpbPackaging.resolve()
+        }
```
woggioni, line 83: why do we need to do this?

chrisr3: Because of [Gradle #17765](https://github.com/gradle/gradle/issues/17765) :scream:

chrisr3: BTW, are you going to rebase this branch? I have an extra tweak to apply now that #446 has been merged.

woggioni: done

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
chrisr3: AFAIK the "transitive" flag only means something if the `Configuration` can also be resolved. The "transitive CPK" mechanism _does_ try to honour this flag too, although it is actually implemented by the `CordappDependencyCollector`... :male_detective:.

woggioni: I've noticed that even if it isn't clear to me why we've had to do that, resolving the configuration eagerly

woggioni:
can't we use that `Attribute` black magic for both `ProjectDependency` and `ExternalDependency`?

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
chrisr3: We need Gradle to execute all `withDependencies` handlers for `cpb` and `allCordapps` _before_ it executes this `withDependencies` handler for `cpbPackaging`, as otherwise `cpb.allDependencies` will not contain everything.

I don't see an obvious way of controlling the order in which Gradle executes these handler, other than [doing what Gradle says is required to invoke it](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html#withDependencies-org.gradle.api.Action-).

woggioni:
ok, I've just read you comment about it from yesterday... adding avarian would force the whole repo to contain Gradle's metadata in addition to Maven's ones

chrisr3:
> can't we use that `Attribute` black magic for both `ProjectDependency` and `ExternalDependency`?
> As I understand it is basically [this stuff](https://docs.gradle.org/current/userguide/variant_model.html)

Using Gradle variant attributes is fine (:thinking:?) for `ProjectDependency` because these are handled by the `cordapp-cpk` plugin in a way that should be 100% invisible to CorDapp developers. However, an `ExternalDependency` must be downloaded from a Maven repository somewhere, which means that Gradle would also start requiring that CPKs in remote Maven repositories be published with a Gradle `LibraryElements` variant attribute of "cpk" :scream:.

Now Gradle _supports_ publishing variant attributes into remote Maven repositories, by "embracing and extending" the POM:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
```
_However_, this would also prevent people from ever publishing their CPKs as ordinary Maven artifacts. I prefer _not_ to be used as a Gradle Salesman, and so the `cordapp-cpk` plugin uses well-understood Maven coordinates instead.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbPlugin.kt
chrisr3: With any luck, we will only need this `afterEvaluate` handler for some versions of Gradle < 7.x (for unknown value of x...).

woggioni:
well, as long as they solely use our Gradle plugin to publish the CPKs that shouldn't be an issue (I thought it were the `.module` files to contain Gradle's proprietary metadata). I generally agree with the idea that it is better not to tie ourselves to Gradle too much if possible as someone might at some point come up with the idea of consuming stuff that has been published on a Maven repo with our Gradle plugin using Maven

chrisr3:
> well, as long as they solely use our Gradle plugin to publish the CPKs that shouldn't be an issue (I thought it were the `.module` files to contain Gradle's proprietary metadata).

The `.module` file _does_ contain the Gradle metadata, and it is the extra comment tag in the POM that tells Gradle to use that `.module` file.

The `cordapp-cpk` plugin isn't "magic". If someone wants to create a Maven equivalent or even a _hand-crafted_ CPK then that would be their business.

chrisr3:
So assuming that the `.cpb` file has the correct contents now, it probably still needs to be both reproducible and signed.

woggioni:
the content looks good to me, I'll add the functionality to sign it and write a unit test to check it actually works with Maven as well (I assume I can use `MavenLocal` for unit testing). Will also take care of making it reproducible

chrisr3:
> the content looks good to me, I'll add the functionality to sign it and write a unit test to check it actually works with Maven as well (I assume I can use `MavenLocal` for unit testing). Will also take care of making it reproducible

Maybe not to `MavenLocal`? You can define your own file-based Maven repository for publishing to like this:
```gradle
publishing {
    repositories {
        maven {
            name = 'Bongo'
            url = layout.buildDirectory.dir('repos/bongo')
        }
    }
}
```
and then publish CPKs to it using the `publishAllPublicationsToBongoRepository` task.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbTask.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    companion object {
+        private const val CPB_ARTIFACT_CLASSIFIER = "package"
+        const val CPB_FILE_EXTENSION = "cpb"
+        private const val CPK_FILE_SUFFIX = ".$CPK_FILE_EXTENSION"
+        private const val JAR_FILE_SUFFIX = ".jar"
+        private val EXCLUDED_CPK_TYPES = setOf("corda-api")
+    }
+
+    val signing : Signing = project.objects.newInstance(Signing::class.java)
```
chrisr3, line 25: I am assuming that the CPB will want to use the same signing information as both the "main" jar and the CPK, because the CorDapp developers are not going to want to configure it twice. You should be able to do what the CPK task does, i,e, add a `doLast` handler when the plugin configures the new `cpbTask`.

And because the `cordapp.signing` object is now part of the `cpbTask`'s input properties, you also need to declare it as such:
```kotlin
val cpbTaskProvider = project.tasks.register(CPB_TASK_NAME, CpbTask::class.java) { cpbTask ->
    cpbTask.inputs.nested("signing", cordapp.signing)
    cpbTask.from(allCPKs)
    ...
```
This guarantees that Gradle knows to rebuild the CPB if the signing block changes.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbTask.kt
```diff
@@ -0,0 +1,76 @@
...
+        isCaseSensitive = true
+        isPreserveFileTimestamps = false
+        isReproducibleFileOrder = true
+        isZip64 = true
+
+        doLast {
+            if(signing.enabled.get()) {
+                sign(signing, archiveFile.get().asFile)
+            }
+        }
```
chrisr3, line 46: It's probably better to do this from the plugin, because the `signing` block you want to use belongs to the `cordapp` extension and not this task.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpb/CpbTask.kt
woggioni: mmm... I think it is better to provide the flexibility to sign the cpb with a different key but I could use the `Signing` object of the cordapp extension as the default. The idea is that you should be able to use this task to create a cpb with whatever signature you want and whatever cpk (even one that comes from a different project or  an external Maven repo)

chrisr3: People are creating their _own_ CPB here. Other people's CPK will be downloaded from remote repositories and will already have been signed. Please just use the existing mechanism for the sake of consistency.

chrisr3 (commented):
This looks good overall. How difficult would it be to make the `external-cordapp` project more ambitious, e.g. now that we've worked out how it _can_ use the `:corda-api` project without being inconsistent with other test projects?

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpk/CordappWithOwnGuavaVersionTest.kt
```diff
@@ -1,5 +1,6 @@
 package net.corda.plugins.cpk
 
+import net.corda.plugins.cpb.CpbTest
```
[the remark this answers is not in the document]
woggioni: sorry, this was a mistake

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+publishing {
+    repositories {
+        maven {
+            name = 'Test'
+            url = getProperty("maven.repository.dir")
+        }
+    }
+    publications {
+        maven(MavenPublication) {
+            artifact cpk
```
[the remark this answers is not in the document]
woggioni: what do you mean with it being "incomplete"? what is the issue with only publishing the cpk (which is all I need here)?

cordapp-cpk/src/test/resources/cordapp-cpb/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+allprojects {
+    group = 'com.example'
+    version = cordapp_version
+    repositories {
+        maven {
+            name = 'test-repository'
+            url = getProperty("maven.repository.dir")
+        }
+    }
+}
```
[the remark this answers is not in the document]
woggioni: In theory I do not even need to apply `repositories.gradle` though

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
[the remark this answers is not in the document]
chrisr3: > what do you mean with it being "incomplete"? what is the issue with only publishing the cpk (which is all I need here)?

The CPK has a `cordapp` classifier, and so your published Maven artifact has no "default" artifact, i.e. no "main" jar. This in turn means that the POM's `<packaging>` type will be `pom` and not `jar`, and _that_ means that the `cordapp-cpk` plugin will not also publish the "companion" POM artifact that makes transitive CPK dependencies work...

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,139 @@
...
+        private val HEX_ARRAY = "0123456789ABCDEF".toCharArray()
+        private fun ByteArray.toHex(): String {
+            val hexChars = CharArray(size * 2)
+            for (j in indices) {
+                val v: Int = this[j].toInt() and 0xFF
+                hexChars[j * 2] = HEX_ARRAY[v ushr 4]
+                hexChars[j * 2 + 1] = HEX_ARRAY[v and 0x0F]
+            }
+            return String(hexChars)
+        }
```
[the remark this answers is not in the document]
woggioni: this is a lot slower, almost as slow as copying all the test resource files in the junit temporary directory XP

cordapp-cpk/src/test/resources/cordapp-cpb/build.gradle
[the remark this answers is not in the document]
chrisr3: > well... I wasn't expecting my test artifact to be found on Maven central, but naver say never...

Hopefully not :wink:, but there's no need to search all of those Maven repositories in vain either.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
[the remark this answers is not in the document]
chrisr3: I haven't benchmarked either, but this version is considerably _shorter_. And if Kotlin provided a built-in `ByteArray.toHex()` function then I would ask you to use that, regardless of how it were implemented.

By the same metric, not writing a lot of unnecessary code to copy resource files around is "shorter" too.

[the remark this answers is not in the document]
woggioni: I don't need to benchmark it to tell that allocating a new string for every byte (and parsing the format string) is waaay slower than the above code, it also absolutely irrelevant given it will be executed a dozen of times at most on 32 bytes arrays, I'll amend it (it's a shame this is not in kotlin stdlib)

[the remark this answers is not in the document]
woggioni: I'll probably just use Base64 in the future, which is part of Java stdlib (and is also more compact)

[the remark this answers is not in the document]
chrisr3: > I'll probably just use Base64 in the future, which is part of Java stdlib (and is also more compact)

Base64 is fine too. As you point out, any difference in performance between base64 and hexadecimal is irrelevant because all that matters here is that you are creating two `Map<String, String>` for `AbstractMap.equals()` to compare.

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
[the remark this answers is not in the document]
woggioni: I see... it probably works simply because external-cordapps doesn't have any other CPK dependency itself, may I should actually add one to test this as well?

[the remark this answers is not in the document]
chrisr3: > I see... it probably works simply because `external-cordapps` doesn't have any other CPK dependency itself, maybe I should actually add one to test this as well?

Yes, something like that. Actually even as little as adding
```gradle
dependencies {
    cordaProvided project(':corda-api')
}
```
to `external-cordapp` would do because the "companion" POM contains both `cordaProvided` and `cordapp` dependencies. You'd just need to publish the `corda-api` jar to your "Test" repository too.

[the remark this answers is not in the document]
woggioni: but if I declare `corda-api` as cordaProvided it won't be included in the CPB, how would that test be meaningful?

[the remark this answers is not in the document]
chrisr3: > but if I declare `corda-api` as `cordaProvided` it won't be included in the CPB, how would that test be meaningful?

Hmm, OK. I was thinking that transitively resolving the `corda-api` artifact would allow your `cordapp-cpb` project to depend on `corda-api`, and so allow it to include a stub contract class etc. _However_, that's probably testing the `cordapp-cpk` plugin rather than the `cordapp-cpb` plugin :grin:.

If you want to resolve transitive CPKs 3 levels deep then you'll need something that looks more like `transitive-cordapps`. Have you also tested that "system CPKs" are _not_ packaged?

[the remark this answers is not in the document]
woggioni: > If you want to resolve transitive CPKs 3 levels deep then you'll need something that looks more like transitive-cordapps. Have you also tested that "system CPKs" are not packaged?

I've tested it manually but it would be good to have a unit test instead

cordapp-cpk/src/test/resources/external-cordapp/corda-platform-cordapp/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+    contract {
+        name = 'Corda platform CorDapp'
+        versionId = cordapp_contract_version.toInteger()
+        licence = 'Test-Licence'
+        vendor = 'R3'
+    }
+}
+
+dependencies {
+//    cordaProvided project(":corda-api")
```
woggioni, line 19: for some reason, if I uncomment this the `:external-cordapp-transitive-dependnecy` cpk is not included in the assembled CPB

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+}
+
+evaluationDependsOnChildren()
+
+allprojects { p ->
+    println(p.path)
+
+    if (p.path in [rootProject.path, ":corda-platform-cordapp", ":external-cordapp-transitive-dependency"]) {
+        apply plugin: 'java-library'
+        apply plugin: 'maven-publish'
```
chrisr3, line 12: You need to publish the `corda-api` project too, if you expect
```gradle
dependencies {
   cordaProvided project(':corda-api')
}
```
to work. Although best _not_ to modify `group`, `version` or the `jar` task for the `corda-api` module.

cordapp-cpk/src/test/resources/external-cordapp/corda-platform-cordapp/build.gradle
chrisr3: Perhaps because you haven't published `corda-api` to your "Test" repository?

chrisr3 (commented):
Just a few things...

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,146 @@
...
+                return FileVisitResult.CONTINUE
+            }
+
+            override fun visitFileFailed(file: Path, exc: IOException) = FileVisitResult.CONTINUE
+
+            override fun postVisitDirectory(dir: Path, exc: IOException?): FileVisitResult = FileVisitResult.CONTINUE
+        })
+
+        val expectedCpks = (externalCpkFiles.asSequence().filter { it != cordaPlatformCordappCpk} + sequenceOf(*internalCpKFiles.toTypedArray()))
+            .associateByTo(TreeMap(), { it.fileName.toString() }, {Files.newInputStream(it).use(::sha256)})
```
chrisr3, line 110: We can ask Gradle to help us here by copying all the CPKs into a subdirectory of `testProject.buildDir`, This would allow us to simplify the walk:
```kotlin
val dependentCpkFiles = mutableListOf<Path>()
Files.walkFileTree(testProject.buildDir, object : FileVisitor<Path> {
    ..
}
```
followed by:
```kotlin
val cordaPlatformCordappCpk = dependentCpkFiles.singleOrNull {
    it.fileName.toString().startsWith("corda-platform-cordapp")
} ?: Assertions.fail("Corda platform CorDapp CPK not found")

val expectedCpks = dependentCpkFiles.filter { it != cordaPlatformCordappCpk }
    .associateByTo(TreeMap(), { it.fileName.toString() }, {Files.newInputStream(it).use(::sha256)})
```

cordapp-cpk/src/test/resources/cordapp-cpb/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+        licence = 'Test-Licence'
+        vendor = 'R3'
+    }
+}
+
+dependencies {
+    cordapp group: group, name: "external-cordapp", version: cordapp_version
+    cordapp project('project-dependency')
+}
+
```
chrisr3, line 38: And ask Gradle to copy all dependent CPKs somewhere more convenient. The `copyCPKs` task will be executed as a dependency of the `assemble` task:
```gradle
configurations {
    cpks
}

dependencies {
    cpks project(path: 'project-dependency', configuration: 'cordaCPK')
    cpks "com.example:corda-platform-cordapp:$cordapp_version:cordapp@cpk"
    cpks "com.example:external-cordapp:$cordapp_version:cordapp@cpk"
    cpks "com.example:external-cordapp-transitive-dependency:$cordapp_version:cordapp@cpk"
}

def cpkDir = file("$buildDir/cpks")
def copyCPKs = tasks.register('copyCPKs', Copy) {
    from configurations.cpks
    into cpkDir
}

artifacts {
    archives(file: cpkDir) {
        builtBy copyCPKs
    }
}
```

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,146 @@
...
+                .withTaskName("publishAllPublicationsToTestRepository")
+                .build("-Pmaven.repository.dir=$mavenRepoDir",
+                    "-Pcorda_api_version=$cordaApiVersion",
+                    "-Phost_version=$hostVersion",
+                    "-Pcordapp_version=$cordappVersion",
+                    "-Pcordapp_contract_version=$expectedCordappContractVersion")
+            testProject = GradleProject(cpbProjectDir, reporter)
+                .withTestName("cordapp-cpb")
+                .withSubResource("project-dependency/build.gradle")
+                .withTaskName("cpb")
```
chrisr3, line 55: We can allow the `GradleProject` to execute its default task, i.e. `assemble`.

chrisr3 (commented):
No need to search for all CPKs when we can ask Gradle to copy them all into `testProject.buildDir` for us instead.

cordapp-cpk/src/test/resources/external-cordapp/corda-platform-cordapp/build.gradle
woggioni: I haven't but I don't understand why that would be an issue, that is a project dependency, not an external dependency

chrisr3: > I haven't but I don't understand why that would be an issue, that is a project dependency, not an external dependency

Because once published into a Maven repository, the transitive dependency on `corda-api` becomes:
```xml
<dependency>
    <groupId>net.corda</groupId>
    <artifactId>corda-api</artifactId>
    <version>5.0.0</version>
</dependency>
```
which Gradle no longer recognises as `project(':corda-api')`.

woggioni: shouldn't it then throw an exception saying that such a dependency cannot be found in any Maven repo? 

it silently "works" instead and it excludes the CPK of `:external-cordapp-transitive-dependency` which I don't understand how is related to `project(':corda-api')`

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.junit.jupiter.api.*
+import org.junit.jupiter.api.io.TempDir
+import java.io.IOException
+import java.io.InputStream
+import java.nio.file.*
+import java.nio.file.attribute.BasicFileAttributes
+import java.security.DigestInputStream
+import java.security.MessageDigest
+import java.security.cert.X509Certificate
+import java.util.*
```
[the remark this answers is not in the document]
woggioni: yeah, I don't like `*` imports in general but I do not understand how to persuade intellij to stop using them (I have configured it in that way in theory, but it keeps happening with the `java.util` package)

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
woggioni: I do not understand why, especially considering it is only needed at compile time of this project (I was expecting the `cordaProvided` configuration to not translate to consumer projects but maybe I'm wrong here)

cordapp-cpk/src/test/resources/external-cordapp/corda-platform-cordapp/build.gradle
chrisr3: Have you checked the output of the test Gradle projects for any warnings? I believe the "transitive CPK" mechanism _does_ warn if it cannot find a CPK's companion POM, which also contains the CPKs `cordaProvided` dependencies. However, Gradle cannot _fail_ if there's a problem with the companion POM because it might belong to a third-party CPK and so not be fixable by the current CorDapp developer.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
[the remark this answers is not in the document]
chrisr3: > yeah, I don't like `*` imports in general but I do not understand how to persuade intellij to stop using them (I have configured it in that way in theory, but it keeps happening with the `java.util` package)

Have you checked both `Settings/Editor/Code Style/Kotlin` and `Settings/Editor/Code Style/Java`? You may have o scroll down to the bottom.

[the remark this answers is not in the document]
woggioni: solved it, thanks

cordapp-cpk/src/test/resources/external-cordapp/corda-platform-cordapp/build.gradle
woggioni: as far as I know, if a transitive dependency cannot be found anywhere, the dependency resolution fails even if you, as a developer, cannot do anything about it (because it's not a direct dependency). I don't think we should have a different behavior...

For the warning.. I don't see any warning apart from a few 

```
[ant:signjar] 
[ant:signjar] Warning: 
[ant:signjar] The signer's certificate is self-signed.
```

which I assume to be related to the fact that I'm using the dev cert for signing

woggioni: I mean, if I publish a library that depends on `corda-base` and I consume it from a project that is missing the R3 artifactory repo from the `repositories` list, Gradle fails complaining that `corda-base` cannot be found anywhere, I don't think there's anything wrong with it

woggioni: I found the warning though
```
Resource missing. [HTTP GET: https://repo.maven.apache.org/maven2/net/corda/corda-api/5.0.0/corda-api-5.0.0.pom]
CorDapp has unresolved dependencies:
- net.corda:corda-api:5.0.0
Cannot resolve CPK companion artifact 'com.example.external-cordapp:external-cordapp.corda.cpk:1.2.1' - SKIPPED
```

woggioni: I do not understand the last line though.. I'm investigating it

chrisr3: I'm fairly sure there are warnings like these somewhere:
```
Configuring cordapp-cpb as a cordapp
Resource missing. [HTTP GET: https://repo.maven.apache.org/maven2/net/corda/corda-api/5.0.0/corda-api-5.0.0.pom]
CorDapp has unresolved dependencies:
- net.corda:corda-api:5.0.0
Cannot resolve CPK companion artifact 'com.example.external-cordapp:external-cordapp.corda.cpk:1.2.1' - SKIPPED
```

chrisr3: The point is that the `cordapp-cpk` plugin found the companion POM for
```
com.example.external-cordapp:external-cordapp.corda.cpk:1.2.1
```
but couldn't resolve it because
```
net.corda:corda-api:5.0.0
```
had not been published to your "Test" repository.

woggioni: ok, but why is it excluding the dependencies that could be resolved though? `:external-cordapp-transitive-dependency` has been published in the test repo, even if it's `cordaProvided` dependency, `corda-api` was not... I do not think it is a good idea to skip `corda-api` because if it cannot not be found but I especially do not understand why `external-cordapp-transitive-dependency` is also skipped

chrisr3: > as far as I know, if a transitive dependency cannot be found anywhere, the dependency resolution fails even if you, as a developer, cannot do anything about it (because it's not a direct dependency).

This transitive resolution is being performed by the `CordappDependencyCollector` as it tries to resolve the "companion" POMs. The "companion" POM is an extra "helper" artfifact that I make a _best effort_ to publish using _Dark Gradle Magic_, and so I cannot _demand_ that every published CPK will have one.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
woggioni: you mean check it is present in the `META-INF/CPKDependencies` file of the CPK file assembled by `cordapp-cpb` ?

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
woggioni: for now I simply removed the dependency on `corda-api` as publishing that would have required to change all the test cases that make use of it (due to the missing `maven.repository.dir` project property), tests are green but I don't think the current behavior of the `cordapp-cpk` plugin is correct

cordapp-cpk/src/test/resources/external-cordapp/corda-platform-cordapp/build.gradle
woggioni: imho you should require that if it is required for transitive dependencies resolution, otherwise you would end up assebling CPB or installing CPKs that cannot be started

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
chrisr3: > you mean check it is present in the `META-INF/CPKDependencies` file of the CPK file assembled by `cordapp-cpb` ?

It is a transitive CPK and so can be expected to be found in `testProject.cpkDependencies`, i.e. `cordapp-cpb-cordapp.cpk`.

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
chrisr3: > for now I simply removed the dependency on `corda-api` as publishing that would have required to change all the test cases that make use of it (due to the missing `maven.repository.dir` project property), tests are green but I don't think the current behavior of the `cordapp-cpk` plugin is correct

You don't have to modify the `corda-api` project _in any way_ in order to publish it. The only reason you weren't publishing it before was because of this filter in your `allprojects {}` block:
```gradle
allprojects { p ->
    println(p.path)

    if (p.path in [rootProject.path, ":corda-platform-cordapp", ":external-cordapp-transitive-dependency"]) {
        // DELIBERATELY ChOOSING NOT TO PUBLISH ':corda-api'
    }
}
```
Whereas something like this in `external-cordapp/build.gradle` would have published `:corda-api` just fine:
```gradle
allprojects {
    apply plugin: 'maven-publish'

    publishing {
        repositories {
            maven {
                name = 'Test'
                url = getProperty("maven.repository.dir")
            }
        }
        publications {
            maven(MavenPublication) {
                from components.java
                pluginManager.withPlugin('net.corda.plugins.cordapp-cpk') {
                    artifact tasks.named('cpk')
                }
            }
        }
    }
}
```

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,131 @@
...
+                .withTestName("cordapp-cpb")
+                .withSubResource("project-dependency/build.gradle")
+                .build("-Pmaven.repository.dir=$mavenRepoDir",
+                    "-Pcordapp_version=$cordappVersion",
+                    "-Phost_version=$hostVersion",
+                    "-Pcordapp_contract_version=$expectedCordappContractVersion")
+        }
+
+        @AfterAll
+        fun tearDown() = externalCordappProjectDir.deleteAll()
```
chrisr3, line 66:
```kotlin
externalProject.delete()
```

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        private val cordaDevCertPrincipal = X500Principal("CN=Corda Dev Code Signer, OU=R3, O=Corda, L=London, C=GB")
+
+
+        private lateinit var externalProject: GradleProject
+        private lateinit var testProject: GradleProject
+        private lateinit var testDir: Path
+        private lateinit var mavenRepoDir : Path
+        private lateinit var externalCordappProjectDir : Path
+        private lateinit var cpbProjectDir : Path
+
```
chrisr3, line 37: I think all of these `Path` properties can become local variables now. In fact, `@TempDir tmpDir` could probably be replaced by `@TempDir testDir`.

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+    contract {
+        name = 'External CorDapp'
+        versionId = cordapp_contract_version.toInteger()
+        licence = 'Test-Licence'
+        vendor = 'R3'
+    }
+}
+
+dependencies {
+//    cordaProvided project(":corda-api")
```
chrisr3, line 56: I do believe you can add this dependency back in, if you tweak the `allprojects {}` block above it. Perhaps just split the publishing out into a separate `allprojects {}` block?

chrisr3 (commented):
The `external-cordapp` project can publish `corda-api` very easily, and so I think it's worth doing.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
woggioni: but I use them in multiple methods.. how can they become local variable? I assign them in `@BeforeAll` and use them in `@Test`

chrisr3: > but I use them in multiple methods.. how can they become local variable? I assign them in `@BeforeAll` and use them in `@Test`

But that's the point - you _don't_ use them in multiple methods any more. At least, not once the `@AfterAll` method becomes
```kotlin
@AfterAll
fun tearDown() = externalProject.delete()
```
BTW, does JUnit execute that `@AfterAll` method if it hasn't also been annotated with `@JvmStatic`?

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
chrisr3: In fact, if you put the publishing into an `allproject {}` block that also applies the `java-library` and `maven-publish` plugins then you can probably get rid of `evaluationDependOnChildren()` too.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            testProject = GradleProject(cpbProjectDir, reporter)
+                .withTestName("cordapp-cpb")
+                .withSubResource("project-dependency/build.gradle")
+                .build("-Pmaven.repository.dir=$mavenRepoDir",
+                    "-Pcordapp_version=$cordappVersion",
+                    "-Phost_version=$hostVersion",
+                    "-Pcordapp_contract_version=$expectedCordappContractVersion")
+        }
+
+        @AfterAll
```
chrisr3, line 65: JUnit won't execute this unless you also annotate as `@JvmStatic`.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
```diff
@@ -0,0 +1,138 @@
...
+        private lateinit var cpbProjectDir : Path
+
+        @Suppress("unused")
+        @BeforeAll
+        @JvmStatic
+        fun setup(@TempDir tmpDir: Path, reporter: TestReporter) {
+            testDir = tmpDir
+            mavenRepoDir = Files.createDirectory(testDir.resolve("maven"))
+            externalCordappProjectDir = Files.createTempDirectory(testDir.parent, "externalCordapp")
+            cpbProjectDir = Files.createDirectory(testDir.resolve("cpb"))
```
chrisr3, line 45:
```kotlin
fun setup(@TempDir testDir: Path, reporter: TestReporter) {
    val mavenRepoDir = Files.createDirectory(testDir.resolve("maven"))
    val externalCordappProjectDir = Files.createTempDirectory(testDir.parent, "externalCordapp")
    val cpbProjectDir = Files.createDirectory(testDir.resolve("cpb"))
```
and your `@Test` will still compile because it no longer uses these as `Path` properties.

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
```diff
@@ -0,0 +1,61 @@
+plugins {
+    id 'net.corda.plugins.cordapp-cpk'
+}
+
+evaluationDependsOnChildren()
```
chrisr3, line 5: We don't need `evaluationDependsOnChildren()` if we always apply `java-library` in `allprojects {}`. (Because `evaluationDependsOnChildren()` is one of those things that you never want in a Gradle project if you can help it, just like `afterEvaluate {}`.)

cordapp-cpk/src/test/resources/external-cordapp/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+
+allprojects { p ->
+    apply plugin: 'maven-publish'
```
chrisr3, line 8: Just add
```gradle
apply plugin: 'java-library`
```
under here to enable `from components.java`.

chrisr3 (commented):
A few little things left.

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
chrisr3: It is _trivial_ to demonstrate that none of `mavenRepoDir`, `externalCordappProjectDir` nor `cpbProjectDir` is used by the `@Test` function any more, but whatever. This code will change when JUnit 5.8 is released anyway.

chrisr3 (approved):
LGTM

cordapp-cpk/src/test/kotlin/net/corda/plugins/cpb/CpbTest.kt
woggioni: fixed, sorry... I didn't understand what you meant in the first place

cordapp-cpk/src/test/resources/corda-api/build.gradle
```diff
@@ -34,3 +34,5 @@ Bundle-SymbolicName: net.corda.api
 Bundle-Name: Fake Corda APIs
 """
 }
+
+
```
chrisr3, line 38: No idea what's happening here... :thinking: :wink:

chrisr3 (approved):
LGTM

chrisr3 (approved):

PR #448, opened 2021-07-21
CORE-2057: Test with CPKs from remote Maven repositories.

chrisr3:
Test the "transitive CPK" mechanism when the dependent CPKs are artifacts in remote Maven repositories.

This is slightly more work than it should be because `@TempDir` in JUnit 5.7.x only creates a single temporary directory. This is due to change in JUnit 5.8, so replicate this behaviour ourselves for now.


cordapp-cpk/src/test/resources/transitive-remote-cordapps/src/main/kotlin/com/example/transitives/ExampleContract.kt
```diff
@@ -0,0 +1,14 @@
...
+import net.corda.v5.ledger.contracts.Contract
+import net.corda.v5.ledger.transactions.LedgerTransaction
+import org.apache.commons.io.IOUtils
+import java.io.ByteArrayInputStream
+
+class ExampleContract : Contract {
+    override fun verify(ltx: LedgerTransaction) {
+        ByteArrayInputStream("Hello Corda!".toByteArray()).use { input ->
+            IOUtils.copy(input, System.out)
+        }
```
woggioni, line 12: I assume the point of this is to simply use `commons-io:commons-io:$commons_io_version` somehow so that it appears in OSGi metadata

chrisr3: It's a "something" to appear in both the OSGi metadata _and_ the CPK's `lib/` folder, yes.

woggioni (approved):
LGTM

PR #449, opened 2021-07-22
CORE-2057: Update how the CPK tests publish CPKs.

chrisr3:
The most reliable way to check whether a project has a CPK to publish is to use the plugin manager.


woggioni (approved):
I was going to suggest this in the previous PR but I thought it wasn't too important

chrisr3:
> I was going to suggest this in the previous PR but I thought it wasn't too important

It turns out that the `MavenPublication` fails to publish the `cpk` artifact without this change because the `cpk` task hasn't been registered yet. (Most likely, the `MavenPublication` is being created before the `cordapp-cpk` plugin has been applied... :facepalm:)

chrisr3:
> > I was going to suggest this in the previous PR but I thought it wasn't too important
> 
> It turns out that the `MavenPublication` fails to publish the `cpk` artifact without this change because the `cpk` task hasn't been registered yet. (Most likely, the `MavenPublication` is being created before the `cordapp-cpk` plugin has been applied... facepalm)

The `:corda-api` project doesn't have a CPK artifact to publish in the first place, of course.

PR #451, opened 2021-07-26
CORE-2087: Allow cordapp-cpk to apply OSGi "consumer policy" to Corda packages.

chrisr3:
Corda's APIs currently all live under this package mask:
```
net.corda.v5.*
```
We expect to add other masks over time. Add a new `Import-Policy-Packages` property to `cordapp-configuration.properties` containing these package masks as a comma-separated list. The `cordapp-cpk` plugin will uses these masks to create extra `Import-Package` instructions for Bnd:
```
net.corda.v5.*;version='${range;[=,+);${@}}'
```
The `${range}` macro here will ensure that OSGi will only check the Corda API major version number when loading CorDapps, e.g.:
```
5 <= version < 6
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,28 @@ subprojects {
                 languageVersion = of(8)
             }
         }
+
+        if (parent == rootProject) {
+            // Add a META-INF/pluginVersion.properties resource into each plugin jar.
+            def generateResources = tasks.register('generateResources') {
+                def versionFile = layout.buildDirectory.dir('generated').map { dir ->
+                    dir.file('pluginVersion.properties')
+                }
+
+                outputs.file versionFile
+                doLast {
+                    versionFile.get().asFile.withWriter { out ->
+                        out.println "version=${version}"
+                    }
+                }
+            }
+
+            tasks.named("processResources", ProcessResources) {
+                from(files(generateResources)) {
+                    into 'META-INF'
+                }
+            }
+        }
     }
 
     tasks.withType(JavaCompile).configureEach {
--- a/cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
+++ b/cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
@@ -34,7 +34,7 @@ fun TaskInputs.nested(nestName: String, osgi: OsgiExtension) {
 }
 
 @Suppress("UnstableApiUsage", "MemberVisibilityCanBePrivate", "unused")
-open class OsgiExtension(objects: ObjectFactory, jar: Jar) {
+open class OsgiExtension(objects: ObjectFactory, useImportPolicy: Provider<Boolean>, jar: Jar) {
     private companion object {
         private const val CORDAPP_CONFIG_PLUGIN_ID = "net.corda.cordapp.cordapp-configuration"
         private const val CORDAPP_CONFIG_FILENAME = "cordapp-configuration.properties"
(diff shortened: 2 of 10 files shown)


build.gradle
```diff
@@ -24,6 +24,28 @@ subprojects {
                 languageVersion = of(8)
             }
         }
+
+        if (parent == rootProject) {
+            // Add a META-INF/pluginVersion.properties resource into each plugin jar.
+            def generateResources = tasks.register('generateResources') {
+                def versionFile = layout.buildDirectory.dir('generated').map { dir ->
```
woggioni, line 31: wouldn't it be better to use 
```
project.convention.getPlugin(JavaPluginConvention.class)
    .sourceSets
    .getByName(SourceSet.MAIN_SOURCE_SET_NAME)
    .resources
    .outputDir
```
?

chrisr3: Better how? :thinking:

I've actually based this solution on [this advice](https://discuss.gradle.org/t/how-can-a-custom-gradle-plugin-determine-its-own-version/36761/3) from the Gradle Forums.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
```diff
@@ -61,6 +61,11 @@ open class OsgiExtension(objects: ObjectFactory, jar: Jar) {
             "org.hibernate.proxy"
         ))
 
+        private val Array<String>.packageRange: IntRange get() {
+            val firstIdx = if (size > 2 && this[0] == "META-INF" && this[1] == "versions") { 3 } else { 0 }
+            return firstIdx..size - 2
+        }
```
woggioni, line 67: it doesn't make much difference, but this is not a generic method that can apply to any array of Strings, it would be easier to understand as a private method

build.gradle
woggioni: I had 2 concerns:
1 - Gradle changes the location of the generated source and the `generated` directory doesn't exist anymore at some point (and I think code would break in that case)
2 - The project compiles with an annotation processors that, for some reason, creates an `pluginVersion.properties` that you would be overwriting here; this is not possible though because the actual destination of generated code is not the `generated` folder directly but some more nested path inside it, so this is fine

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
chrisr3: > it doesn't make much difference, but this is not a generic method that can apply to any array of Strings, it would be easier to understand as a private method

You're right - it doesn't make much difference :wink:. However, I've changed it anyway.

cordapp-cpk/src/main/kotlin/net/corda/plugins/cpk/OsgiExtension.kt
```diff
@@ -61,6 +61,15 @@ open class OsgiExtension(objects: ObjectFactory, jar: Jar) {
             "org.hibernate.proxy"
         ))
 
+        private fun getPackageRange(packages: Array<String>): IntRange {
```
palade, line 64: A comment with an example may be useful

palade (approved):

build.gradle
chrisr3: I think that's the point - I create _my own_ output directory which has its own dedicated task to generate its contents. Gradle will create the directory automatically, and will also complain if the final jar tries to contain two files called `META-INF/pluginVersion.properties`.

[Gradle 7 is becoming more picky about such things.](https://discuss.gradle.org/t/gradle-7-0-seems-to-take-an-overzealous-approach-to-inter-task-dependencies/39656/2)

PR #454, opened 2021-07-30
CORE-2136: Support platform dependencies when building CPKs.

chrisr3:
Update `cordapp-cpk` plugin to support Gradle `platform` dependencies. Specifically, the transitive CPK mechanism now supports Corda publishing a POM containing the Corda API versions as a `<dependencyManagement/>` element.

Also add Dokka support for generating the plugin's `javadoc` artifact.


mnesbit (approved):
OK

PR #463, opened 2021-08-31
CORE-2495: Copy name, scm, licenses and developers to CPK companion POM.

chrisr3:
Maven Central requires POMs to contain the following extra fields:
- `name`
- `scm`
- `licenses`
- `developers`

And include `organization` too, for good measure.


palade (approved):

PR #464, opened 2021-09-06
CORE-2561: Allow Quasar instrumentation to be turned off for Test and JavaExec.

chrisr3:
Quasar instrumentation is expensive, so allow us to turn it off for all `Test` and `JavaExec` tasks via each project's `quasar` extension.

The assumption here is that we either want instrumentation or not, and will not need to instrument only _some_ `Test` or `JavaExec` tasks.

We still instrument by default, for the sake of backwards compatibility with earlier versions of the `quasar-utils` plugin.

Use case:
```groovy
quasar {
    instrumentTests = false
    instrumentJavaExec = false
}
```


quasar-utils/build.gradle
```diff
@@ -37,3 +39,18 @@ dependencies {
     testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:$junit_jupiter_version")
     testImplementation "org.assertj:assertj-core:$assertj_version"
 }
+
+tasks.withType(Test).configureEach {
+    javaLauncher = javaToolchains.launcherFor {
+        languageVersion = of(11)
+    }
```
woggioni, line 46: this will force the test execution to run under a different JVM (not the same JVM that is running Gradle) and I'm not sure users will be happy with that (I personally prefer configuring the gradle JVM myself instead of relying on Gradle toolchain black magic)

chrisr3: Yes, because Quasar 0.8.x only runs on Java 11. However, the plugins themselves are built using Java 8.

woggioni: oh, right.. I initially thought it was for the tests in user's projects

woggioni (approved):

PR #465, opened 2021-09-07
NOTICK: Use shared pipeline

benmcmahon100:


chrisr3 (commented):
So which repositories will this shared pipeline publish into? We currently expect:
- `-SNAPSHOT` -> `corda-dev`
- "releases" -> `corda-releases`

Note that the plugins also have `release/5.0` and `release/4.0` branches for Corda 4.x and earlier, and we _must_ be able to publish new releases of their plugins into the repositories that everyone already expects.

benmcmahon100:
@chrisr3 so the shared pipeline uses an entirely different flow and verbiage, but I'm going to revert and try the old pipeline and setup as I estimate there's more work in migrating than what I thought, as we don't have a docker that supports JDK 8 & 11, and TBH at that stage I'd almost prefer to just do them separately, but either way I think it'd be a slower turn around. Currently testing out if I can spot why it's upset from before

PR #466, opened 2021-09-07
CORE-2568: Rollback plugin version

benmcmahon100:
Corrected branch name and applied fix, and some pipeline whitespace formatting

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -50,7 +50,7 @@ pipeline {
     stages {
         stage('Build') {
             steps {
-                sh './gradlew --no-daemon -s --no-build-cache clean build'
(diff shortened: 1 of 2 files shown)


chrisr3 (commented):
So does this fix publishing to Artifactory? And if so, do you understand why?

BTW CORE-2568 seems to concern the "internal" Gradle plugins' repo rather than this one.

benmcmahon100:
Ah darn I made a typo with the issue title. I don't really understand beyond that it seems that the latest plugin is not compatible with the latest jenkins plugin. I checked and there's no new version of the jenkins plugin, but I think jenkins itself might be behind. I'll raise it with waldemar

chrisr3 (approved):
LGTM, although we still need to understand this plugin issue!

PR #467, opened 2021-09-08
CORE-2578: Exclude JetBrains' annotations jar from CPKs.

chrisr3:
CPKs do not need to contain `org.jetbrains:annotations`. This is just a transitive dependency of
```groovy
'net.corda.kotlin:kotlin-stdlib-jdk8-osgi'
```
that is not required by
```groovy
'org.jetbrains.kotlin:kotlin-osgi-bundle'
```


mnesbit (approved):
OK

PR #482, opened 2021-11-18
CORE-3049: Differentiate between remote and project CPK dependencies.

chrisr3:
Denote dependent CPKs which belong to the same Gradle build as the CPK that the `cordapp-cpk` plugin is building as all being signed with the same key. This allows us to avoid writing the SHA-256 of the "development" key into the CPK's `META-INF/CPKDependencies` file, and hence to sign all of these CPKs later with the actual "release" key.

CPKs which are downloaded from a Maven repository are assumed to have been signed with their correct "release" key.


ronanbrowne (approved):
LGTM

PR #485, opened 2021-12-31
ENT-6378 Include `corda-shell` in `cordformation` task

lankydan:
Due to moving the `corda-shell` out of the main Corda codebase, it is no longer run by default and instead relies on the `corda-shell` jar existing in a node's `drivers` directory to execute.

The `cordformation` task has been modified to include the `corda-shell` jar as long as the project calling `cordformation` contains a runtime dependency (e.g. `runtimeOnly`) on the `corda-shell` artifact.

For example:

OS:

```
runtimeOnly "net.corda:corda-shell:4.8-SNAPSHOT"
```

ENT:

```
runtimeOnly "com.r3.corda:corda-shell:4.8-SNAPSHOT"
```

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Node.kt
@@ -423,6 +423,7 @@ open class Node @Inject constructor(private val project: Project) {
             installWebserverJar()
         }
         installAgentJar()
+        installShell()
         installDrivers()
         installCordapps()
         installCordappConfigs()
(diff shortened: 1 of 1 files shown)


lankydan:
I'm all ears if there is a better way to do this @chrisr3, this is the only thing I could figure out on my own.

chrisr3 (commented):
The idea is presumably to install `corda-shell` into the Node's `drivers/` directory, along with everything that has already been added to the `cordaDriver` Gradle configuration. So my obvious question is: Can't we just add `corda-shell` to the `cordaDriver` configuration too?
```gradle
cordaDriver {
    "net.corda:corda-shell:$cordaReleaseVersion"
}
```
And if not, why not? (This approach would not require a code change.)

If the Node no longer includes `corda-shell` by default, why should this dependency exist on the runtime classpath? Conversely, if someone now needs to add an explicit dependency on `corda-shell` manually then again, why not add it manually to the `cordaDrivers` configuration?

I also need to ask wht `corda-shell`'s transitive dependencies are these days, as these would need to be installed too if not already present. Fortunately, the `cordaDriver` configuration is resolvable, and so would handle this too.

lankydan:
I'll have a look at the `cordaDriver` code.

The `corda-shell` will become a fat jar (at least thats what I've done in the shell repo), so it doesn't need to depend on anything else.

Thanks.

chrisr3:
> I'll have a look at the `cordaDriver` code.

It's not much: the resolved contents of the `cordaDriver` configuration are just copied to the node's `drivers` directory.

You shouldn't need to change `cordformation` because of `corda-shell`. You _will_ almost certainly need to update DemoBench though.

PR #486, opened 2021-12-31
CORE-3384: Remove Lombok from the flask plugin.

chrisr3:
Remove Lombok from the `flask` plugin project so that there is less to maintain.

There should be no functional change, although we need some new functional interfaces that allow lambdas to throw exceptions.

### DIFF
--- a/flask/build.gradle
+++ b/flask/build.gradle
@@ -82,9 +82,6 @@ repositories {
 }
 
 dependencies {
-    ['compileOnly', 'annotationProcessor', 'testCompileOnly', 'testAnnotationProcessor'].each { conf ->
-        add(conf, [group: "org.projectlombok", name: "lombok", version: lombok_version])
-    }
     testImplementation project("flask-common")
     testImplementation "org.assertj:assertj-core:$assertj_version"
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junit_jupiter_version"
--- a/flask/flask-common/build.gradle
+++ b/flask/flask-common/build.gradle
@@ -7,10 +7,6 @@ repositories {
 }
 
 dependencies {
-    ['compileOnly', 'annotationProcessor', 'testCompileOnly', 'testAnnotationProcessor'].each { conf ->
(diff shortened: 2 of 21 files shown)


ronanbrowne (approved):
LGTM

woggioni-contour:
> There should be no functional change, although we need some new functional interfaces that allow lambdas to throw exceptions.

There is no way you can sort that unfortunately, all the Java Stream API has been designed to accept functional interfaces that do not throw exception (see [Stream.map](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#map(java.util.function.Function)) as an example, which takes a `java.util.function.Function` as an argument, whose [apply](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/Function.html#apply(T)) method doesn't throw)...  Java checked exception are just a design mistake

chrisr3:
> There is no way you can sort that unfortunately, all the Java Stream API has been designed to accept functional interfaces that do not throw exception.

Actually we solved that problem :grin:.

woggioni-contour:
that's not bad as a workaround, but it still have the limitation of unnecessarily wrapping the original exception (which is a common side effect of checked exceptions)

chrisr3:
> that's not bad as a workaround, but it still have the limitation of unnecessarily wrapping the original exception (which is a common side effect of checked exceptions)

It's a solution, not a work-around. And only a checked exception would be wrapped because unchecked exceptions are just rethrown. Not that wrapping exceptions has ever been raised as a concern on the first place.

woggioni-contour:
it's a workaround as you have to create a subclass of the original java functional interface every time, still a good workaround though. Wrapping exceptions for no other reason than make the compiler happy has ALWAYS been a concern

chrisr3:
> it's a workaround as you have to create a subclass of the original java functional interface every time, still a good workaround though. Wrapping exceptions for no other reason than make the compiler happy has ALWAYS been a concern

It is called "standard Java programming".

PR #488, opened 2022-02-08
CORE-3809: Fix cordapp-cpb plugin only to resolve cpbPackaging configuration after entire configuration phase.

chrisr3:
Resolve the `cpbPackaging` configuration via a `projectsEvaluated` handler instead of the project's own `afterEvaluate` handler. We _must not_ resolve this configuration until every other project in the build has also had a chance to apply the `cordapp-cpk` plugin.

This is a subtle fix to an incomplete workaround to a Gradle bug.


ronanbrowne (approved):
Ship it

PR #510, opened 2022-06-29

josephzunigadaly:
Copy `cordapp-cpk` into `cordapp-cpk2`.

This is a copy of existing code into a new folder. It has not been modified and is not included in gradle settings.

Future PRs will build on this. This PR gives us:
- Trusted (already approved) code to use as a starting point
- Easy check of no modifications in new folder with diff
- Future PRs will only show new changes in diff, not this big list of files

We can verify that the copy has been made with no changes by running diff:
```shell
corda-gradle-plugins $ ./gradlew clean               # Removes build folder from cordapp-cpk
corda-gradle-plugins $ diff -r cordapp-cpk cordapp-cpk2
corda-gradle-plugins $ 
```


chrisr3 (commented):
So what's the plan on how to proceed? Both `cordapp-cpk` and `cordapp-cpk2` plugins (and presumably `cordapp-cpb` and `cordapp-cpb2` too) still share the same package names, which means that their classes cannot co-exist on the plugins classpath.

You probably also need to refactor packages
- `net.corda.plugins.cpk` -> `net.corda.plugins.cpk2`
- `net.corda.plugins.cpb` -> `net.corda.plugins.cpb2`

etc to make this work. But why copy at all, when you could `git mv ...`? I was expecting the `6.0.0-BETAxx` version to be built by the `release/6.0` branch from now on.

josephzunigadaly:
> But why copy at all, when you could git mv ...? I was expecting the 6.0.0-BETAxx version to be built by the release/6.0 branch from now on

I would be happy with a move instead. I'll swap this PR to do a move.

chrisr3:
>  It has not been modified and is not included in gradle settings.

I'm not following you. Gradle will ignore all of these files unless you include `cordapp-cpk2` in `settings.gradle`.

chrisr3 (approved):
LGTM - and probably more "exciting" than expected too :wink:

PR #512, opened 2022-07-06

josephzunigadaly:
Add Corda-CPB-Format = 2.0 to the CPB manifest.

### DIFF
--- a/cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpb2/CpbTest.kt
+++ b/cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpb2/CpbTest.kt
@@ -1,5 +1,7 @@
 package net.corda.plugins.cpb2
 
+import net.corda.plugins.cpb2.CpbTask.Companion.CPB_FORMAT_VERSION
+import net.corda.plugins.cpb2.CpbTask.Companion.CPB_VERSION_ATTRIBUTE
 import net.corda.plugins.cpk2.GradleProject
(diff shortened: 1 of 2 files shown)


cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpb2/CpbTest.kt
```diff
@@ -100,7 +102,8 @@ class CpbTest {
         val embeddedCpkFiles = TreeMap<String, String>()
         JarInputStream(Files.newInputStream(cpbFile), true).use { jarInputStream ->
             assertEquals("customName", jarInputStream.manifest.mainAttributes.getValue("Corda-CPB-Name"))
-            assertEquals("customVersion", jarInputStream.manifest.mainAttributes.getValue("Corda-CPB-Version"))
+            assertEquals("customVersion", jarInputStream.manifest.mainAttributes.getValue(CPB_VERSION_ATTRIBUTE))
+            assertEquals("2.0", jarInputStream.manifest.mainAttributes.getValue(CPB_FORMAT_VERSION))
```
chrisr3, line 106: Any reason why you're not using `CPB_CURRENT_FORMAT_VERSION` here? Or should it be `private const`?

josephzunigadaly: I've updated the test to use the constant as the expected value.

chrisr3 (approved):
LGTM

PR #515, opened 2022-07-11
CORE- 5660: add snyk scanning to the main CI build 

ronanbrowne:
- calls snyk cli tool under the hood to run a test and generate reports 
- Exclude PR builds as they will be handled by separate logic at a later stage
- Generate HTML report on Build pipeline also if this has ran


chrisr3 (approved):
You're the Jenkinsfile Guy :wink:...

PR #520, opened 2022-07-12

josephzunigadaly:
Remove META-INF/DependencyConstraints file


cordapp-cpk2/src/test/resources/with-dependent-cordapp/build.gradle
```diff
@@ -40,13 +40,3 @@ dependencies {
...
-
-// Copy the DependencyConstraints XML from ':cordapp' into our build directory.
-def copy = tasks.register('copy', Copy) {
-    into project.layout.buildDirectory
-    from project(':cordapp').tasks.named('cordappDependencyConstraints')
-}
-
-tasks.named('assemble') {
-    dependsOn copy
-}
```
chrisr3, line 52: This script meant that `WithDependentCordappTest` didn't need to rummage around in the test project directory structure to find something from the `:cordapp` sub-project. If you want the CPK now instead of the XML then you can do this:
```gradle
configurations {
    cordappDependency {
        canBeConsumed = false
    }
}

// Copy the CPK from ':cordapp' into our build directory.
def copy = tasks.register('copy', Copy) {
    into project.layout.buildDirectory
    from configurations.cordappDependency
}

tasks.named('assemble') {
    dependsOn copy
}
```
And also add this to the `dependencies` block, of course:
```gradle
dependencies {
    ...
    cordappDependency project(path: ':cordapp', configuration: 'cordaCPK')
}
```

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/WithDependentCordappTest.kt
```diff
@@ -114,16 +112,16 @@ class WithDependentCordappTest {
...
-            .noneMatch { it.fileName == "biz.aQute.bnd.annotation-${bndVersion}.jar" }
-            .noneMatch { it.fileName == "slf4j-api-${librarySlf4jVersion}.jar" }
-            .noneMatch { it.fileName == "slf4j-api-${cordaSlf4jVersion}.jar" }
-            .anyMatch { it.fileName == "guava-${libraryGuavaVersion}.jar" }
-            .anyMatch { it.fileName == "commons-io-$commonsIoVersion.jar" }
-            .anyMatch { it.fileName == "library.jar" }
+        val contractCpk = testProjectDir.resolve("cordapp")
+            .resolve("build")
+            .resolve("libs")
+            .resolve("cordapp-cordapp.cpk")
```
chrisr3, line 118: I think this should be (after tweaks to `with-dependent-cordapp` script):
```kotlin
val contractCpk = testProject.buildDir.resolve("cordapp-cordapp.cpk")
assertThat(contractCpk).isRegularFile
assertThat(listLibrariesForCpk(contractCpk))
   // etc...
```
I prefer the test projects to make their "interesting" artifacts readily available, rather than have the tests rummage through an arbitrary directory structure.

chrisr3 (changes requested):
A few Kotlinisms, but also some deeper issues with a couple of the tests.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/GradleProject.kt
```diff
@@ -87,6 +86,15 @@ fun Path.hashOfEntry(entryName: String): ByteArray {
...
 }
 
+fun listLibrariesForCpk(cpk: Path) = JarInputStream(cpk.toFile().inputStream(), false).use {
+    generateSequence { it.nextJarEntry }
+        .filter { !it.isDirectory }
+        .map { it.name }
+        .filter { it.startsWith("lib/") }
+        .map { it.removePrefix("lib/") }
+        .toList()
+}
```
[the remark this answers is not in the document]
josephzunigadaly: I've updated it to use the `listLibrariesForCpk` suggested here

cordapp-cpk2/src/test/resources/eager-dependencies/build.gradle
```diff
@@ -25,7 +25,3 @@ dependencies {
     cordaProvided project(':corda-api')
     implementation "commons-io:commons-io:$commons_io_version"
 }
-
-tasks.withType(net.corda.plugins.cpk2.DependencyConstraintsTask) {
-    println "Eagerly configuring DependencyConstraints task '$name'"
-}
```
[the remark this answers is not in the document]
josephzunigadaly: Done. I've used the suggested replacement.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/WithDependentCordappTest.kt
josephzunigadaly: I agree, that's much better. Thank you for your help.

cordapp-cpk2/src/test/resources/with-dependent-cordapp/build.gradle
josephzunigadaly: These changes have been applied.

josephzunigadaly:
Thank you for your advice Chris. I've applied all the suggested changes.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/CordappGradleConfigurationsTest.kt
```diff
@@ -74,40 +74,30 @@ class CordappGradleConfigurationsTest {
 
     @Test
     fun testApiIncluded() {
-        assertThat(testProject.output)
-            .contains("CorDapp library dependency: javax.annotation-api-1.3.2.jar")
```
chrisr3, line 78: This information is still valid for CPKv2, even if the particular CPKv1 task that used to log it no longer exists. Is there anywhere else you could log this now? (Perhaps not as part of _this_ PR, though.)

chrisr3 (approved):
This looks OK, but we have lost the "INFO" level logging of the CPK's libraries. It would be nice to get that back for the final PR.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/CordappGradleConfigurationsTest.kt
josephzunigadaly: Thank you Chris. I'll bring back this logging in a future PR.

PR #522, opened 2022-07-14
CORE-5071: Add snyk Gradle plugin to C5 projects

mcgovc:
Removing .ci/nightly/JenkinsfileSnykScan as this is not needed. The existing CI needed to be amended so the following CI job just needs to run the additional Snyk stage for the release/5.1 branch:
https://ci01.dev.r3.com/job/Corda-Gradle-Build-Plugins/job/corda-gradle-plugins/job/release%252F5.1/ 

Similar change to https://github.com/corda/corda-gradle-plugins/pull/515

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -39,6 +39,8 @@ pipeline {
         ARTIFACTORY_BUILD_NAME = "Corda Gradle Plugins / Publish To Artifactory"
             .replaceAll("/", " :: ")
         ARTIFACTORY_REPO = "${isRelease ? "corda-releases" : "corda-dev"}"
+        SNYK_TOKEN = "r3-gradle-plugins"
+        SNYK_COMMANDS = "--all-sub-projects --prune-repeated-subdependencies --target-reference='${env.BRANCH_NAME}' --project-tags=Branch='${env.BRANCH_NAME.replaceAll("[^0-9|a-z|A-Z]+","_")}'"
     }
     stages {
         stage('Build') {
@@ -95,10 +97,23 @@ pipeline {
     }
     post {
         always {
+            script {
+                if(!changeRequest()){
+                    snykSecurityScan.generateHtmlElements()
+                }
+            }
             junit testResults: '**/build/test-results/**/*.xml', keepLongStdio: true, allowEmptyResults: true
         }
         cleanup {
             deleteDir()
         }
     }
+}
+
+def isReleaseBranch() {
+    return (env.BRANCH_NAME =~ /^release\/.*$/)
+}
+
+def isMainBranch() {
+    return (env.BRANCH_NAME =~ /^master$/)
 }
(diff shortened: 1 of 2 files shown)


chrisr3 (commented):
I'm not a Jenkins expert, so will defer to @benmcmahon100.

.ci/Jenkinsfile
```diff
@@ -47,6 +49,16 @@ pipeline {
                 sh './gradlew --no-daemon -s --no-build-cache clean build'
             }
         }
+        stage('Snyk Security Scan') {
+            when {
+                not {
+                    changeRequest()
+                }
+            }
```
benmcmahon100, line 57: This will then run on any branch build, including manually triggered ones for misc branches. Is that correct?

.ci/Jenkinsfile
```diff
@@ -85,10 +97,23 @@ pipeline {
     }
     post {
         always {
+            script {
+                if(!env.CHANGE_ID) {
```
benmcmahon100, line 101: Should probably match up with the above condition exactly

.ci/Jenkinsfile
```diff
@@ -85,10 +97,23 @@ pipeline {
...
     }
 }
+
+def isReleaseBranch() {
+    return (env.BRANCH_NAME =~ /^release\/.*$/)
+}
+
+def isMainBranch() {
+    return (env.BRANCH_NAME =~ /^master$/)
+}
```
benmcmahon100, line 119: I thought these already existed?

.ci/Jenkinsfile
chrisr3: I would like to know that my PR is going to fix any reported security problems _before_ merging it, so running on branch builds sounds like a Good Idea... :thinking:.

mcgovc: I think Chris is correct here, and also this is keeping consistent with release/6.0 & release/5.0
https://github.com/corda/corda-gradle-plugins/blob/master/.ci/Jenkinsfile
https://github.com/corda/corda-gradle-plugins/blob/release/5.0/.ci/Jenkinsfile

.ci/Jenkinsfile
mcgovc: In the post stage 'changeRequest()' does not have the desired return value. Tested here and while there was no errors the HTML element for the Snyk Report was not generated. 
Run - https://ci01.dev.r3.com/job/Corda-Gradle-Build-Plugins/job/corda-gradle-plugins/job/release%252F5.1/3/

.ci/Jenkinsfile
mcgovc: The changes here were ported from the main (release/6.0) branch and I was trying to keep them as consistent as possible. As the logic is not being used within the pipeline it can be removed. 
Ref main (release/6.0): https://github.com/corda/corda-gradle-plugins/commit/18d04004d6d6405a790835d6f6ebd3b01e5c20e9

benmcmahon100: Fair enough! What's release/6.0 for?

benmcmahon100 (approved):

.ci/Jenkinsfile
mcgovc: Honestly, I'm not sure

PR #527, opened 2022-07-27
CORE-4355: Restore "CorDapp library dependency" log output

josephzunigadaly:
Restore the library log message and the tests covering the log messages.

### DIFF
--- a/cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/DependencyCalculator.kt
+++ b/cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/DependencyCalculator.kt
@@ -156,9 +156,7 @@ open class DependencyCalculator @Inject constructor(objects: ObjectFactory) : De
         _libraries.apply {
             setFrom(packageFiles - bundledFiles)
             disallowChanges()
-        }
-
-        for (library in _libraries) {
+        }.forEach { library ->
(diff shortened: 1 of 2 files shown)


cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/DependencyCalculator.kt
```diff
@@ -158,6 +158,10 @@ open class DependencyCalculator @Inject constructor(objects: ObjectFactory) : De
             disallowChanges()
         }
 
+        for (library in _libraries) {
+            logger.info("CorDapp library dependency: {}", library.name)
+        }
```
chrisr3, line 163: I think we can chain these staements together:
```kotlin
_libraries.apply {
    setFrom(packageFiles - bundledFiles)
    disallowChanges()
}.forEach { library ->
    logger.info("CorDapp library dependency: {}", library.name)
}
```

chrisr3 (changes requested):
Small Kotliny thing...

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/DependencyCalculator.kt
josephzunigadaly: Done

chrisr3 (approved):
LGTM

PR #528, opened 2022-07-28
CORE-4813: CPK Package names validation

josephzunigadaly:
Prevent CPKs from exporting `net.corda` package names.


cordapp-cpk2/src/test/resources/cpk-export-reserved-package/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+dependencies {
+    cordaProvided project(':corda-api')
+    implementation "commons-io:commons-io:$commons_io_version"
+}
+
+jar {
+    archiveBaseName = 'cpk-export-reserved-package'
```
chrisr3, line 29: But suppose the CPK writer does this instead?
```gradle
jar {
    archiveBaseName = 'cpk-export-reserved-package'
    osgi {
        autoExport = false
    }
}
```
And provides a `package-info.java` too, of course:
```java
@Export
package net.corda.contract;

import org.osgi.annotation.bundle.Export;
```
And not forgetting to include this in the test:
```kotlin
.withSubResource("src/main/java/net/corda/contract/package-info.java")
```

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/OsgiExtension.kt
```diff
@@ -188,6 +188,11 @@ open class OsgiExtension(objects: ObjectFactory, jar: Jar) {
 
     @get:Input
     val exports: Provider<String> = _exports.map { names ->
+
+        require(names.none { it.startsWith("net.corda.") }) {
+            "CorDapps must not export \"net.corda\" package"
+        }
+
```
chrisr3, line 195: This only catches the packages that either the plugin computes when
```gradle
osgi {
    autoExport = true
}
```
or that the CPK authors explicitly export themselves. However, if you use `autoExport = false` then the plugin delegates this calculation to Bnd and the `@Export` annotations etc. This allows CPK authors who understand OSGi to declare some of the packages in their "main" jar as "Private-Package", i.e. invisible from other "main" jars in the sandbox.

You will probably want to perform this kind of check in the `VerifyBundle` task instead, adding extra errors into `verifier`:
```kotlin
verifier.error("Export Package clause found for Corda package [%s]", packageName)
```
Any violations detected are reported together at the end.

chrisr3 (changes requested):
Putting the check here only covers the default case where the `cordapp-cpk2` plugin exports every package in the "main" jar.

cordapp-cpk2/src/test/resources/cpk-export-reserved-package/build.gradle
josephzunigadaly: I've added this as a test and applied the fix suggested in the second comment.

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/OsgiExtension.kt
josephzunigadaly: I've moved the check into VerifyBundle and used verifier.error for every net.corda package found.
Also expanded the existing test to check two net.corda packages are detected.

cordapp-cpk2/src/test/resources/cpk-export-reserved-package-no-auto-export/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    cordaProvided project(':corda-api')
+    implementation "commons-io:commons-io:$commons_io_version"
+}
+
+jar {
+    archiveBaseName = 'cpk-export-reserved-package-no-auto-export'
+    osgi {
+        autoExport = false
+    }
+}
```
chrisr3, line 33: These two test projects are essentially the same, except one uses `autoExport = true` and the other `autoExport = false`. So why not unify them?
```gradle
jar {
    archiveBaseName = 'cpk-export-reserved-package`
    osgi {
        autoExport = osgiAutoExport
    }
}
```
Each test can pass its own value for the `osgiAutoExport` property via `buildAndFail()`:
```koltin
"-PosgiAutoExport=false"
```

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
```diff
@@ -107,6 +108,11 @@ open class VerifyBundle @Inject constructor(objects: ObjectFactory) : DefaultTas
         }
     }
 
+    private fun verifyExportPackage(verifier: Verifier) = verifier.exportPackage.entries
+        .map { it.key }
```
chrisr3, line 112: Better:
```kotlin
verifier.exportPackage.keyList()
```

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
```diff
@@ -107,6 +108,11 @@ open class VerifyBundle @Inject constructor(objects: ObjectFactory) : DefaultTas
...
+        .filter { it.startsWith("net.corda.") }
```
chrisr3, line 113: What happens if someone puts something into the `net.corda` package itself?

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
```diff
@@ -88,6 +88,7 @@ open class VerifyBundle @Inject constructor(objects: ObjectFactory) : DefaultTas
             verifier.setProperty(STRICT, strict.get().toString())
             verifier.verify()
             verifyImportPackage(verifier)
+            verifyExportPackage(verifier)
```
chrisr3, line 92: The `autoExport = false` setting allows the CPK to have private packages too. And to be honest, _I don't know_ if we can handle CPKs having private `net.corda` packages - although I suspect the security policy will prevent it :thinking:.

chrisr3 (changes requested):
I think you can create a single test CPK project for this, and pass it the `autoExport` value to use via a Gradle property.

And having opened this can of worms, I think I can see some private packages wriggling around... :worm::worm:.

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
josephzunigadaly: It did fail. I've added a test to cover it and updated to code to detect `net.corda` using a regex.

I also added a test to check `net.cordapp` is allowed.

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
josephzunigadaly: Done

cordapp-cpk2/src/test/resources/cpk-export-reserved-package-no-auto-export/build.gradle
josephzunigadaly: That's much better. Thank you.

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
josephzunigadaly: Seems like detecting private packages is desired too.

Do you think we could achieve this by looking at the directory paths in the resulting jar? If any file contains the path `/net/corda/` then we output the error?

chrisr3: Bnd has already scanned the bundle's metadata. Check the `verifier` object for a `privatePackage` member.

josephzunigadaly: Thank you Chris. I'll use that.

josephzunigadaly: Done. I've added a test to cover private packages too.

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
```diff
@@ -107,6 +108,16 @@ open class VerifyBundle @Inject constructor(objects: ObjectFactory) : DefaultTas
         }
     }
 
+    private fun verifyReservedPackage(verifier: Verifier) {
+        val reservedPackageName = Regex("^net.corda(\\..+)?$")
+        verifier.exportPackage
+            .keyList()
+            .asSequence()
+            .plus(verifier.privatePackage.keyList())
```
chrisr3, line 116: Private packages are not the same as exported packages, which makes these distinct violations. These two should therefore generate different error messages. I was expecting something more like:
```
verifyImportPackage(verifier)
verifyExportPackage(verifier)
verifyPrivatePackage(verifier)
```
Obviously the "package name matching" logic can still be shared.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/CpkExportReservedPackageNoAutoExportTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun testLogExplainsFailure() {
+        assertThat(testProject.output).contains(
+            "Export Package clause found for Corda package [net.corda.contract]"
+        )
+    }
+
+    @Test
+    fun testPrivatePackageDetected() {
+        assertThat(testProject.output).contains(
+            "Export Package clause found for Corda package [net.corda.internal]"
```
chrisr3, line 52: Private packages are not "exported" (in an OSGi sense), by definition.

chrisr3 (changes requested):
This is over-generalised - `Export-Package` and `Private-Package` are different OSGi headers, which makes them distinct violations. Atm, a CPK developer would look at these errors and think: 
> "Why is my CPK's private package being exported?"

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
```diff
@@ -107,6 +108,16 @@ open class VerifyBundle @Inject constructor(objects: ObjectFactory) : DefaultTas
         }
     }
 
+    private fun verifyReservedPackage(verifier: Verifier) {
+        val reservedPackageName = Regex("^net.corda(\\..+)?$")
```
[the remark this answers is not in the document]
josephzunigadaly: I've moved it into a companion object and fixed the escaping of the first `.`. Let me know if you'd prefer a string comparison instead.

cordapp-cpk2/src/main/kotlin/net/corda/plugins/cpk2/VerifyBundle.kt
josephzunigadaly: I've split the method into two and shared the package matching logic

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/CpkExportReservedPackageNoAutoExportTest.kt
josephzunigadaly: The private package error message has been changed to:
> "Private package found for Corda package [$packageName]" 

Let me know if you would like different wording.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/CpkExportReservedPackageNoAutoExportTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    private lateinit var testProject: GradleProject
+
+    @BeforeAll
+    fun setup(@TempDir testProjectDir: Path, reporter: TestReporter) {
+        testProject = GradleProject(testProjectDir, reporter)
+            .withTestName("cpk-export-reserved-package")
+            .withSubResource("src/main/java/net/corda/contract/ExampleContract.java")
+            .withSubResource("src/main/java/net/corda/contract/package-info.java")
+            .withSubResource("src/main/java/net/corda/internal/Test.java")
```
chrisr3, line 27: You should be able to add the exact same Java files into _both_ tests. The `net.cordapp.Test` file should be allowed inside a private package as well as an exported one, i.e. there shouldn't be any error concerning `net.cordapp` in the output.

chrisr3 (commented):
This is OK. I just have one tiny extra point about the `autoExport=false` test.

cordapp-cpk2/src/test/kotlin/net/corda/plugins/cpk2/CpkExportReservedPackageNoAutoExportTest.kt
josephzunigadaly: I've added `net.cordapp.Test` to the private test too

chrisr3 (approved):
Negative tests are tricky beasts because they can also pass because of a typo or accidental extra whitespace in the expectation. But this will do.

PR #532, opened 2022-08-18
CORDA-4262: Fix task dependency for cordformation's project CorDapp.

chrisr3:
The `cordaCordapp` configuration is intended to be "consumed" rather than "resolved". Therefore create `BaseForm`'s task dependency using the configuration's `artifacts` instead of its `buildDependencies`.

And even better, I've worked out how to include our `net.corda.plugins.cordapp` plugin in a `net.corda.plugins.cordformation` test.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,12 @@
 
 ## Version 5.1
 
+### Version 5.1.1
+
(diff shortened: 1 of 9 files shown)


ronanbrowne (approved):
LGTM

PR #543, opened 2022-09-03
CORDA-4263: Allow cordformation to recognise jars with -jdk11 appendix.

chrisr3:
Allow `cordformation` to deploy Corda jars that have been published with archive appendix of `jdk11`.

### DIFF
--- a/cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
+++ b/cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
@@ -38,7 +38,7 @@ class Cordformation : Plugin<Project> {
         }
 
         fun createJarRegex(jarName: String, releaseVersion: String): Regex {
-            return "\\Q$jarName\\E(-enterprise)?(-jdk11)?-\\Q$releaseVersion\\E(-.+)?\\.jar\$".toRegex()
+            return "\\Q$jarName\\E(-enterprise)?(-jdk1[17])?-\\Q$releaseVersion\\E(-.+)?\\.jar\$".toRegex()
         }
 
         /**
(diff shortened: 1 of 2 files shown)


cordformation/src/main/kotlin/net/corda/plugins/Cordformation.kt
```diff
@@ -37,18 +37,23 @@ class Cordformation : Plugin<Project> {
             return outputFile
         }
 
+        fun createJarRegex(jarName: String, releaseVersion: String): Regex {
+            return "\\Q$jarName\\E(-enterprise)?(-jdk11)?-\\Q$releaseVersion\\E(-.+)?\\.jar\$".toRegex()
```
arshadm, line 41: Could we make it so that it's not jdk11 specific, maybe something like jdk[0-9]+ ?

chrisr3: Do you expect a Java 17 version of Corda 4?

arshadm: It was just hardwiring a jdk version that seemed a bit strange. I guess we can update the plug-in if we need to support anything other than jdk11

chrisr3: I didn't want to claim for something that we would never create, and as far as I know that's currently just Java 11.

The 5.1.x branch has replaced this regular expression entirely because it has created an explicit `corda` configuration to seach for `corda.jar` inside. (And possibly `corda-testserver` / `corda-webserver` too, but that's all!)

arshadm (approved):

PR #544, opened 2022-09-03
CORDA-4263: Allow cordformation to recognise jars with -jdk11 appendix.

chrisr3:
Allow `cordformation` to deploy Corda jars that have been published with archive appendix of `jdk11`.


ronanbrowne (approved):
LGTM

PR #546, opened 2022-09-20
CORE-6789: Upgrade default Quasar version to 0.8.8_r3.

chrisr3:
Update `quasar-utils` plugin to use Quasar 0.8.8_r3 by default.

Also update documentation and tests.


ronanbrowne (approved):
LGTM

PR #553, opened 2022-09-22
CORE-6614 CORE-5821: added waiver

nargas-ritu:
CORE-6614 CORE-5821: added waivers to the reported vulnerabilities


chrisr3 (commented):
This waiver completely misses the point that Snyk is only finding Kotlin in the `:jar-filter:unwanteds` sub-project, _which only exists to test the `jar-filter` plugin and is utterly irrelevant to Corda_, and so Snyk should not be scanning it in the first place.

What you actually want is a waiver for _anything_ concerning the `:jar-filter:unwanteds` sub-project,

chrisr3:
Alternatively, we can try #554.

PR #554, opened 2022-09-22
CORE-6614: Remove Kotlin from :jar-filter:unwanteds runtimeClasspath.

chrisr3:
The `:jar-filter:unwanteds` is a test artifact that is built using Kotlin. Snyk really has no business scanning it in the first place, but converting its Kotlin dependency from `api` to `compileOnly` should suffice for now since Snyk should only be scanning the `runtimeClasspath` configurations.


nargas-ritu (approved):
Makes sense!

PR #556, opened 2022-09-26
CORE-6551: Add applyImportPolicies to CPK plugin's task inputs.

chrisr3:
Add the `applyImportPolicies` property to the jar task's inputs when creating CPKs. This is largely for "completeness" because the policies are included into the `Import-Package` instruction anyway. However, it does allow Gradle to rebuild CPKs if this flag changes, which is what developers would expect to happen even if it makes no difference to Gradle.


ronanbrowne (approved):
LGTM

PR #557, opened 2022-09-26
NOTICK: Completely remove the default CPK metadata scan.

chrisr3:
Corda's API is just too volatile for the CPK plugin to have any kind of "basic" metadata scan, so strip it out. Especially since these values are _wrong!_


ronanbrowne (dismissed):
LGTM

PR #561, opened 2022-11-03
NOTICK: Update README.md for format version 2

josephzunigadaly:
Update the readme file with some changes made for format version 2.

### DIFF
--- a/cordapp-cpk2/README.md
+++ b/cordapp-cpk2/README.md
@@ -3,9 +3,9 @@
 
 ## Purpose.
 Applying this plugin to a project declares that the project should create a CPK-format CorDapp. The CPK-format
-CorDapp is a ZIP file with a `.jar` extension that contains the output of the `jar` task (the "main" jar), along
+CorDapp is a ZIP file with a `.jar` extension that is the output of the `jar` task, along
 with that jar's dependent jars. In practice, the plugin will not include any of Corda's own jars among these
-dependencies, nor any jars which should be provided by Corda, e.g. Kotlin or Quasar. The "main" jar  should also
+dependencies, nor any jars which should be provided by Corda, e.g. Kotlin or Quasar. The jar  should also
 contain sufficient OSGi metadata to be a valid OSGi bundle.
 
 ## Usage.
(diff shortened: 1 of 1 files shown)


cordapp-cpk2/README.md
```diff
@@ -3,7 +3,7 @@
 
 ## Purpose.
 Applying this plugin to a project declares that the project should create a CPK-format CorDapp. The CPK-format
-CorDapp is a ZIP file with a `.cpk` extension that contains the output of the `jar` task (the "main" jar), along
+CorDapp is a ZIP file with a `.jar` extension that contains the output of the `jar` task (the "main" jar), along
```
relyafi, line 6: Might be worth clarifying how you identify the "main" JAR vs the "CorDapp" JAR given they now both end with JAR. Generally I see that there are two JARs built - one with a `-sources` suffix and one without.

cordapp-cpk2/README.md
```diff
@@ -138,19 +138,9 @@ not become transitive `cordaProvided` dependencies of any CorDapps which depend
...
-        <version>$BUNDLE_VERSION</version>
-        <signers>
-            <signer algorithm="$HASH_ALGORITHM">$BASE64_HASH_OF_PUBLIC_KEY</signer>
-        </signers>
-    </cpkDependency>
-    ...
-</cpkDependencies>
+are listed as lines in this "main" jar's `META-INF/CPKDependencies.json` file:
+```json
+{"formatVersion":"2.0","dependencies":[{"name":"$BUNDLE_SYMBOLIC_NAME","version":"$BUNDLE_VERSION","verifySameSignerAsMe":true}]}
```
relyafi, line 143: It might be more readable to pretty-print this across multiple lines.

cordapp-cpk2/README.md
josephzunigadaly: These two terms refer to the same jar file in version 2. The `-sources` file is the output of `sourcesJar` which is outside the scope of this document.

relyafi: Then isn't the text confusing? I still read it as there were two separate JAR files.

josephzunigadaly: I'll remove talk of the main jar.

josephzunigadaly: Done

relyafi (dismissed):

cordapp-cpk2/README.md
```diff
@@ -306,7 +303,7 @@ to ensure that the bundle respects OSGi's [Service Loader Mediator Specification
 
 ## Corda Metadata
 
-The plugin will generate the following tags in the "main" jar's `MANIFEST.MF` by default:
+The plugin will generate the following tags in the jar's `MANIFEST.MF` by default:
 - `Corda-Contract-Classes`
 - `Corda-Flow-Classes`
 - `Corda-MappedSchema-Classes`
```
chrisr3, line 309: I think this section is no longer true, and should be removed. All Corda Metadata tags are now imported from the `cordapp-configuration` Gradle plugin.

chrisr3 (changes requested):
Small update needed.

cordapp-cpk2/README.md
josephzunigadaly: I've deleted the list of tags and changed the paragraph to:

> The plugin will generate `Corda-*-Classes` tags in the jar's `MANIFEST.MF`. The generated tags are controlled by the `cordapp-configuration` Gradle plugin.

Let me know if you wanted me to delete the whole "Corda Metadata" section.

chrisr3: No harm in keeping it, I guess. But in that case, we should call the plugin by its full Gradle plugin ID `net.corda.cordapp.cordapp-configuration`, and explain where to get it from.

josephzunigadaly: Done

chrisr3 (approved):
LGTM

PR #562, opened 2022-11-03
NOTICK: Removing --all-sub-projects arg from SNYK_COMMANDS

mcgovc:
Description:
Removing --all-sub-projects arg from SNYK_COMMANDS as recently we updated snykDeltaScan to use the --file arg and both being added causes multiple vulnerability reports to be generated by snyk test which means when the result is piped to snyk delta the execution fails.

Tested Against:
https://ci01.dev.r3.com/job/Corda-Gradle-Build-Plugins/job/corda-gradle-plugins/job/PR-561/5/


chrisr3 (approved):
LGTM. (What's the worst thing that could happen?)

ronanbrowne (approved):
LGTM

PR #563, opened 2022-11-08
NOTICK: Bump version due to issue with transitive dependencies no longer being available 

ronanbrowne:
the following transitive dependency from older versions of this plugin is no longer available,   `org.jfrog.buildinfo:build-info-extractor:2.23.0`

Bumping to latest plugin version resolves this  as this is no longer on dependency graph

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -18,7 +18,7 @@ osgi_version=8.1.0
 bnd_version=6.3.1
 jacksonVersion=2.13.4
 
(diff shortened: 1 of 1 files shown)


yarivus (approved):
LGTM

mcgovc (approved):
LGTM

chrisr3 (approved):
LGTM - so long as we can still publish to Artifactory.

PR #571, opened 2023-02-08
NOTICK: Upgrade to Gradle 7.6.

chrisr3:
Upgrade Gradle 7.4.2 -> 7.6.

Note that despite Gradle 7.6 containing the Kotlin 1.7.10 libraries, Kotlin plugins must still target the Kotlin 1.4 API to be compatible with Gradle. _However_, the Kotlin compiler is now warning us that:
```
w: Language version 1.4 is deprecated and its support will be removed in a future version of Kotlin
```

_**I strongly recommend AGAINST writing NEW Gradle plugins in Kotlin!**_


ronanbrowne (approved):
LGTM

PR #572, opened 2023-02-09
CORE-10230: Automatically include JetBrains annotations as compile-only dependency.

chrisr3:
Allow CPK developers access to the JetBrain annotations at compile time, exactly as if they were using `kotlin-stdlib` instead of `kotlin-osgi-bundle`.


ronanbrowne (approved):
LGTM

PR #576, opened 2023-03-14
CORE-11643: Bnd check for classes in wrong folder is now only a warning.

chrisr3:
Instruct Bnd that finding classes in folders other than their package folder is a warning rather than an error. This allows CorDapps to contain multi-release jars, which have classes in `META-INF/versions` folders.

Also translate the "dummy" Corda API from Kotlin to Java to make the tests more realistic, which in turn means that the `kotlin-osgi-bundle` _must_ now be a `cordaProvided` dependency.


ronanbrowne (approved):
lgtm

PR #580, opened 2023-04-19
NOTICK: Fix generated name for CPK marker POM.

chrisr3:
The `MavenPom.getName()` is a Gradle property, so map the parent name instead of assigning it.

Rename this POM as a "marker" rather than a "companion".


ronanbrowne (approved):
lgtm

PR #594, opened 2023-09-17
CORE-17149: Set Configuration flags correctly for Gradle 8+.

chrisr3:
Gradle 8+ requires each `Configuration`'s `canBeConsumed`, `canBeResolved` and `canBeDeclared` flags to be set consistently.

- `canBeConsumed`, `canBeResolved` must not both be true.
- `canBeConsumed`, `canBeDeclared` must not both be true.

`setCanBeDeclared()` was introduced in Gradle 8.2, so set this flag reflectively to preserve the plugin's compatibility with Gradle 7.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 ### Version 7.0.4
 
 * `cordapp-cpk2`: Upgrade to Bndlib 6.4.1.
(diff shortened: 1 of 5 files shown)


ronanbrowne (approved):
LGTM

PR #595, opened 2023-10-05
CORE-17493: Add cordappName field to contract and workflow fields of cpk2

malachyb:
In a cordapp CPK we have a field that sets a name for a cordapp that persists through versions. However, this is assigned automatically based on other values, so we are adding the ability for users to set this name for their own cordapp manually to allow for smoother upgrades where those other values may need to change, as well as to help make CPKs more identifiable to developers.

PR #596, opened 2023-10-25
CORE-17494: Add check to CPB creation to ensure that 2 CPKs don't share a name

malachyb:
As part of https://github.com/corda/corda-gradle-plugins/pull/595, we are adding the ability for users to set an identifiable name for their own CPKs. However, we do not want 2 CPKs in the same CPB to have the same name, so we need to add a check in the CPB plugin to ensure that names are not reused within it.

### DIFF
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
@@ -66,11 +66,11 @@ public class CpbTask extends Jar {
     @NotNull
     public AbstractCopyTask from(@NotNull Object... args) {
         return super.from(args, copySpec ->
-            copySpec.exclude(this::isCPK)
+            copySpec.exclude(this::isValidCPK)
         );
     }
 
-    private boolean isCPK(@NotNull FileTreeElement element) {
+    private boolean isValidCPK(@NotNull FileTreeElement element) {
         if (!element.getName().endsWith(CPK_FILE_SUFFIX)) {
             return false;
         }
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
@@ -14,7 +14,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Set;
 import java.util.jar.JarInputStream;
 
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
@@ -74,9 +74,9 @@ public class CpbTask extends Jar {
             return false;
         }
 
-        Path cpkPath = element.getFile().toPath();
+        final Path cpkPath = element.getFile().toPath();
         try (JarInputStream cpkStream = new JarInputStream(new BufferedInputStream(Files.newInputStream(cpkPath)))) {
-            String cpkType = cpkStream.getManifest().getMainAttributes().getValue(CORDA_CPK_TYPE);
+            final String cpkType = cpkStream.getManifest().getMainAttributes().getValue(CORDA_CPK_TYPE);
             return cpkType != null && EXCLUDED_CPK_TYPES.contains(cpkType.toLowerCase());
         } catch (IOException e) {
             throw new InvalidUserDataException(e.getMessage(), e);
@@ -85,7 +85,6 @@ public class CpbTask extends Jar {
 
     public void checkForDuplicates() {
         Set<String> cpkNames = new HashSet<>();
-        System.out.println("checking duplicates");
         FileCollection files = getInputs().getFiles();
         for (File file : files) {
             Path path = file.toPath();
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
@@ -84,7 +84,7 @@ public class CpbTask extends Jar {
     }
 
     public void checkForDuplicateCpkCordappNames() {
-        Set<String> cpkNames = new HashSet<>();
+        Set<String> cpkCordappNames = new HashSet<>();
         FileCollection files = getInputs().getFiles();
         for (File file : files) {
             Path path = file.toPath();
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
@@ -13,12 +13,15 @@ import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.jar.JarInputStream;
 
 import static java.util.Collections.singleton;
 import static net.corda.plugins.cpk2.CordappUtils.CORDAPP_TASK_GROUP;
 import static net.corda.plugins.cpk2.CordappUtils.CORDA_CPK_TYPE;
+import static net.corda.plugins.cpk2.CordappUtils.CPK_CORDAPP_NAME;
 import static net.corda.plugins.cpk2.CordappUtils.CPK_FILE_EXTENSION;
 
 @DisableCachingByDefault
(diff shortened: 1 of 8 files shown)


cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -55,6 +58,8 @@ public CpbTask() {
             m.getAttributes().put(CPB_NAME_ATTRIBUTE, getArchiveBaseName());
             m.getAttributes().put(CPB_VERSION_ATTRIBUTE, getArchiveVersion());
         });
+
+        cpkNames = new HashMap<>();
```
simon-johnson-r3, line 62: I'm not a gradle expert, but having a static you reset on each instantiation of this type strikes me as a bit odd. Could you explain it more?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -73,6 +78,20 @@ private boolean isCPK(@NotNull FileTreeElement element) {
         final Path cpkPath = element.getFile().toPath();
         try (JarInputStream cpkStream = new JarInputStream(new BufferedInputStream(Files.newInputStream(cpkPath)))) {
             final String cpkType = cpkStream.getManifest().getMainAttributes().getValue(CORDA_CPK_TYPE);
+            final String cpkCordappName = cpkStream.getManifest().getMainAttributes().getValue(CPK_CORDAPP_NAME);
+            System.out.println("Processing CPK " + cpkType + " " + cpkCordappName);
+            if (cpkCordappName != null) {
+                System.out.println(cpkCordappName);
+                if (cpkNames.containsKey(cpkCordappName)) {
+                    if (cpkNames.get(cpkCordappName) == 1) {
+                        cpkNames.put(cpkCordappName, 2);
```
simon-johnson-r3, line 87: So you're allowing two cpks of the same name? I assume this is one for workflow and one for contract? Is it possible to make this check a bit more foolproof?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
malachyb: I'm not entirely sure why but for some reason when I don't reset it then it triggers the error. I think it might be persisting through multiple calls of the same plugin within the same project

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
malachyb: The plugin seems to iterate through each CPK twice or for some other reason calls this function twice on each CPK, so one instance of the name appearing will be counted twice

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
kyriathar: I think what we are doing here (resetting it) suggests it should be made an instance property. 

However it's interesting. I would expect this process to be one-off? I.e. pack CPKs intended for a single CPB and then exits? What more CPKs are there being processed/ output-ed by the prints added?

malachyb: That fixed the issue, thanks

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
simon-johnson-r3: You should be able to call this function as many times as you like. Shouldn't `cpkNames` be local to this method?

Maybe it would help get to the bottom of this behaviour if you logged all the cpkNames? I don't think we can merge this without knowing what's going on.

simon-johnson-r3 (approved):

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -61,18 +65,29 @@ public CpbTask() {
     @NotNull
     public AbstractCopyTask from(@NotNull Object... args) {
         return super.from(args, copySpec ->
-            copySpec.exclude(this::isCPK)
```
kyriathar, line 64: (this is how it used to be so nothing to do with the PR) 

Not a Gradle expert but I think it looks to me a bit weird to exclude when `isCPK`.

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -61,18 +65,29 @@ public CpbTask() {
...
+            if (cpkCordappName != null) {
+                if (cpkNames.containsKey(cpkCordappName)) {
+                    if (!cpkNames.get(cpkCordappName).equals(manifest)) {
+                        throw new InvalidUserDataException("Two CPKs may not share a cordappCpkName. Error in " + cpkCordappName);
+                    }
+                } else {
+                    cpkNames.put(cpkCordappName, manifest);
+                }
+            }
             return cpkType != null && EXCLUDED_CPK_TYPES.contains(cpkType.toLowerCase());
```
kyriathar, line 91: (again before this PR)

the body of this function seems a bit misaligned with its name? The initial check (at line 73) does check if it's a CPK but then here we check if its `Corda-CPK-Type` (which I am not sure we currently use) is of of excluded cpk types?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -61,18 +65,29 @@ public CpbTask() {
...
+            String cpkCordappName = manifest.getMainAttributes().getValue(CPK_CORDAPP_NAME);
+            if (cpkCordappName != null) {
+                if (cpkNames.containsKey(cpkCordappName)) {
+                    if (!cpkNames.get(cpkCordappName).equals(manifest)) {
+                        throw new InvalidUserDataException("Two CPKs may not share a cordappCpkName. Error in " + cpkCordappName);
+                    }
+                } else {
+                    cpkNames.put(cpkCordappName, manifest);
+                }
+            }
```
kyriathar, line 90: Perhaps we could try and extract this logic into some function named `checkCordappNameUniqueness`? This function is about checking if it's CPK or even valid CPK. But this is about 2 CPKs sharing the same cordapp name.

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -31,6 +34,7 @@ public class CpbTask extends Jar {
     public static final String CPB_VERSION_ATTRIBUTE = "Corda-CPB-Version";
     public static final String CPB_FORMAT_VERSION = "Corda-CPB-Format";
     public static final String CPB_CURRENT_FORMAT_VERSION = "2.0";
+    private final HashMap<String, Manifest> cpkNames = new HashMap<>();
```
kyriathar, line 37:
```suggestion
    private final Map<String, Manifest> cpkNames = new HashMap<>();
```

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
malachyb: It might make more sense if I split it into 3 functions that all get called by `isValidCPK`. Or maybe just 2 if Corda-CPK-Type isn't used anywhere?

simon-johnson-r3: Yes, I'd not spotted this before but
```
copySpec.exclude(this::isCPK)
```
seems to be excluding based on a predicate.

That would make "hijacking" this method fairly inappropriate for this same check. I'd revert this method to `isCPK` and make this logic it's own thing separate from that process.

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
kyriathar, line 37: I think worth adding a note on why we decided to map to `Manifest`s

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -78,4 +82,26 @@ private boolean isCPK(@NotNull FileTreeElement element) {
             throw new InvalidUserDataException(e.getMessage(), e);
         }
     }
+
+    public void checkForDuplicates() {
+        Set<String> cpkNames = new HashSet<>();
```
kyriathar, line 87:
```suggestion
        Set<String> cpkCordappNames = new HashSet<>();
```

kyriathar (approved):
LGTM

PR #604, opened 2024-06-24
CORE-20557: Fix issue in CPB and CPK plugins preventing dependencies of transitive dependencies from being included

malachyb:
The packaging plugins had an issue which prevented certain transitive dependencies from being included in the outputs, which meant that packages had to declare transitive dependencies themself. This PR changes that by creating a temporary directory where jar files are extracted from the CPB file of declared dependencies and including those files as dependencies.

### DIFF
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbPlugin.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbPlugin.java
@@ -73,6 +73,7 @@ public final class CpbPlugin implements Plugin<Project> {
             cpbTask.getArchiveVersion().convention(cpkTask.flatMap(Jar::getArchiveVersion));
 
             cpbTask.doFirst(task -> cpbTask.checkForDuplicateCpkCordappNames());
+            cpbTask.doFirst(task -> cpbTask.extractTransitiveDeps());
 
             cpbTask.doLast(task -> {
                 if (cordappExtension.getSigning().getEnabled().get()) {
--- a/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
+++ b/cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
@@ -13,10 +13,16 @@ import org.jetbrains.annotations.NotNull;
 import java.io.BufferedInputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
 import java.util.jar.JarInputStream;
 
 import static java.util.Collections.singleton;
@@ -29,12 +35,15 @@ import static net.corda.plugins.cpk2.CordappUtils.CPK_FILE_EXTENSION;
 public class CpbTask extends Jar {
     private static final String CPB_ARTIFACT_CLASSIFIER = "package";
     public static final String CPB_FILE_EXTENSION = "cpb";
+    private static final String CPB_FILE_SUFFIX = "." + CPB_FILE_EXTENSION;
     private static final String CPK_FILE_SUFFIX = '.' + CPK_FILE_EXTENSION;
     private static final Set<String> EXCLUDED_CPK_TYPES = singleton("corda-api");
     public static final String CPB_NAME_ATTRIBUTE = "Corda-CPB-Name";
     public static final String CPB_VERSION_ATTRIBUTE = "Corda-CPB-Version";
     public static final String CPB_FORMAT_VERSION = "Corda-CPB-Format";
     public static final String CPB_CURRENT_FORMAT_VERSION = "2.0";
+    private static File jarDir;
+    Set<String> cordappFileNames = new HashSet<>();
 
     public CpbTask() {
         setGroup(CORDAPP_TASK_GROUP);
@@ -59,11 +68,21 @@ public class CpbTask extends Jar {
             m.getAttributes().put(CPB_NAME_ATTRIBUTE, getArchiveBaseName());
             m.getAttributes().put(CPB_VERSION_ATTRIBUTE, getArchiveVersion());
         });
+
+        try {
+            jarDir = Files.createTempDirectory("").toFile();
+            jarDir.deleteOnExit();
+        } catch (IOException e) {
+            getLogger().warn("Could not create jar directory: {}", e.getMessage());
+            jarDir = null;
+        }
     }
 
     @Override
     @NotNull
     public AbstractCopyTask from(@NotNull Object... args) {
+        args = Arrays.copyOf(args, args.length + 1);
+        args[args.length - 1] = jarDir;
         return super.from(args, copySpec ->
             copySpec.exclude(this::isCPK)
         );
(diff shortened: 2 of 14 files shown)


cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -104,4 +123,57 @@ public void checkForDuplicateCpkCordappNames() {
             }
         }
     }
+
+    public void extractTransitiveDeps() {
+        if (jarDir != null) {
+            FileCollection jars = getInputs().getFiles();
+            Set<String> jarNames = new HashSet<>();
+            for (File file : jars) {
+                jarNames.add(file.getName());
```
Sakpal: Can we move this inside the next for loop, which is iterating through the same FileCollection?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbPlugin.java
```diff
@@ -73,6 +73,7 @@ public void apply(@NotNull Project project) {
             cpbTask.getArchiveVersion().convention(cpkTask.flatMap(Jar::getArchiveVersion));
 
             cpbTask.doFirst(task -> cpbTask.checkForDuplicateCpkCordappNames());
+            cpbTask.doFirst(task -> cpbTask.extractTransitiveDeps());
```
Sakpal, line 76: How much time has this task added to the entire build process?

malachyb: Running a clean build on the DC repo (which creates 11 cpbs) 3 times on the previous version took me an average of 38.396 seconds. That same build with the new version 3 times took an average of 38.244 seconds. Since I doubt I've sped up the process I'm interpreting that as my changes not having much of an effect on the build time.

cordapp-cpk2/src/main/java/net/corda/plugins/cpk2/DependencyCalculator.java
```diff
@@ -236,6 +261,59 @@ public void calculate() {
...
+        Path path = Paths.get(pathToDir.toString(), jarEntry.getName());
+        if (!Files.exists(pathToDir)) {
+            Files.createDirectories(pathToDir);
+        }
+        if (!Files.exists(path)) {
+            InputStream inputStream = jarFile.getInputStream(jarEntry);
+            Files.copy(inputStream, path);
+            path.toFile().deleteOnExit();
+        }
+        file = path.toFile();
```
kyriathar: Let's put together definition of var and initialization

cordapp-cpk2/src/main/java/net/corda/plugins/cpk2/DependencyCalculator.java
```diff
@@ -236,6 +261,59 @@ public void calculate() {
...
+                        cpbs.add(siblingFile);
+                    }
+                }
+            }
+        }
+        return cpbs;
+    }
+
+    @NotNull
+    private static Set<File> extractJarsFromCpb(Set<File> cpbs, Set<String> cordappFileNames) throws IOException {
```
kyriathar: This function could be named `extractTransitiveDependeciesFromCpb`?

cordapp-cpk2/src/main/java/net/corda/plugins/cpk2/DependencyCalculator.java
```diff
@@ -236,6 +261,59 @@ public void calculate() {
...
+        Path pathToDir = Files.createTempDirectory("jars");
+        pathToDir.toFile().deleteOnExit();
```
kyriathar: Why `deleteOnExit` and not handle it ourselves synchronously?

cordapp-cpk2/src/main/java/net/corda/plugins/cpk2/DependencyCalculator.java
```diff
@@ -236,6 +261,59 @@ public void calculate() {
...
+                    }
+                }
+            }
+        }
+        return cpbs;
+    }
+
+    @NotNull
+    private static Set<File> extractJarsFromCpb(Set<File> cpbs, Set<String> cordappFileNames) throws IOException {
+        Path pathToDir = Files.createTempDirectory("jars");
```
kyriathar: Maybe rename to `tempDir`?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -59,11 +68,21 @@ public CpbTask() {
             m.getAttributes().put(CPB_NAME_ATTRIBUTE, getArchiveBaseName());
             m.getAttributes().put(CPB_VERSION_ATTRIBUTE, getArchiveVersion());
         });
+
+        try {
+            jarDir = Files.createTempDirectory("").toFile();
+            jarDir.deleteOnExit();
+        } catch (IOException e) {
+            getLogger().warn("Could not create jar directory: {}", e.getMessage());
+            jarDir = null;
```
kyriathar: This probably is still  null in this case

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -59,11 +68,21 @@ public CpbTask() {
...
+        }
     }
 
     @Override
     @NotNull
     public AbstractCopyTask from(@NotNull Object... args) {
+        args = Arrays.copyOf(args, args.length + 1);
+        args[args.length - 1] = jarDir;
```
kyriathar, line 81: Where are these arguments passed to?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -104,4 +123,55 @@ public void checkForDuplicateCpkCordappNames() {
             }
         }
     }
+
+    public void extractTransitiveDeps() {
+        if (jarDir != null) {
```
kyriathar: Would we rather want to `checkNotNull` here? I think we always expect this dir to exist?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -104,4 +123,55 @@ public void checkForDuplicateCpkCordappNames() {
...
+            FileCollection jars = getInputs().getFiles();
+            Set<String> jarNames = new HashSet<>();
+            Set<File> cpbs = new HashSet<>();
+            for (File file : jars) {
```
kyriathar: jar

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -104,4 +123,55 @@ public void checkForDuplicateCpkCordappNames() {
...
+                jarNames.add(file.getName());
+                File parent = file.getParentFile();
```
kyriathar: What kind of file is this that we expect it to have multiple CPBs?

kyriathar (commented):
The current review (comments) is just epidermic on minor things as a first pass.

A few questions on the description:

> The packaging plugins had an issue which prevented certain transitive dependencies from being included in the outputs, which meant that packages had to declare transitive dependencies themself. 

What do we mean with "_certain_"? Isn't this issue happening for all transitive dependencies?

> This PR changes that by creating a temporary directory where jar files are extracted from the CPB file of declared dependencies and including those files as dependencies.

I think we already talked about this so just confirming; eventually we are emitting the transitive dependencies in the flat list of dependencies we have in CPB right?

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
```diff
@@ -59,11 +68,21 @@ public CpbTask() {
...
             m.getAttributes().put(CPB_VERSION_ATTRIBUTE, getArchiveVersion());
         });
+
+        try {
+            jarDir = Files.createTempDirectory("").toFile();
+            jarDir.deleteOnExit();
+        } catch (IOException e) {
+            getLogger().warn("Could not create jar directory: {}", e.getMessage());
+            jarDir = null;
+        }
```
josephzunigadaly: I think we should re-throw the exception or remove the try/catch because this is needed for the bug fix to work. If we drop the exception and continue, we are allowing the incorrect behaviour to continue happening in this error path.

It is better to stop the build with an error. The end user will see the build fail and can fix the cause of the error. If we continue, we will build outputs with missing dependencies and gradle will print out "success" at the end of the build.

cordapp-cpk2/src/main/java/net/corda/plugins/cpk2/DependencyCalculator.java
malachyb: I thought using the built-in `deleteOnExit` made more sense than making our own implementation since the files will be used in a separate class from where we create them and passing the data between them would make things more complicated.

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
malachyb: The arguments are passed to `super.from` below which is used to source the files that will be included in the zip file

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
malachyb: Good point, I don't think that's needed.

cordapp-cpk2/src/main/java/net/corda/plugins/cpb2/CpbTask.java
malachyb: I've changed it so it fails with the error now

malachyb:
@kyriathar
> What do we mean with "certain"? Isn't this issue happening for all transitive dependencies?

I probably didn't word that as well as I could have. It's under specific circumstances that the issue occurs i.e. when we import a project dependency that has its own transitive dependencies.

> I think we already talked about this so just confirming; eventually we are emitting the transitive dependencies in the flat list of dependencies we have in CPB right?

Yes, we place the jars in the temporary directory into the flat list of dependencies.

PR #605, opened 2024-07-16
ES-2517: Update gradle.properties to Artifactory version which is oudated

ronanbrowne:
Bump `artifactory_version` to `4.31.0`


wzur-r3 (approved):
LGTM

PR #611, opened 2025-04-09
ENT-12854 - Bumped version for next release

chriscochrane:
Bumped the plugins version for the next release.
Just done release 5.0.19; someone forgot to bump it last time.
Additionally - ensure the repo-key for publishing is correct - it was previously always publishing to "corda-dev".
Apparently this was broken a while back when we switched to using Infra's shared build pipeline - since then all releases of the plugins were released to 'corda-dev'; previously the build infra 'did the right thing' (somehow) and published to corda-releases.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -228,7 +228,7 @@ artifactory {
     publish {
         contextUrl = artifactory_contextUrl
         repository {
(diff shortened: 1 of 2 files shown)


wzur-r3 (approved):
LGTM

